PR #138, opened 2015-01-07
Basic URL Extension functionality working.

kulmugdha:
1. Template parser now allows a parameter to be an optional parameter in a complex segment if
   it is the last and only optional parameter and it is followed by a period.
2. Template matcher modified to take into consideration the optional parameter in the complex
   segment. Also the period shouldn't be present if the optional parameter is not present

### DIFF
--- a/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
@@ -220,14 +220,21 @@ namespace Microsoft.AspNet.Routing.Template
                             // we won't necessarily add it to the URI we generate.
                             if (!context.Buffer(converted))
                             {
-                                return null;
+                                    return null;
                             }
                         }
                         else
                         {
                             if (!context.Accept(converted))
                             {
-                                return null;
+                                if (part.IsOptional && segment.Parts[j - 1].IsOptionalSeperator)
+                                {
+                                    context.Remove(segment.Parts[j - 1].Text);
+                                }
+                                else
+                                {
+                                    return null;
+                                }
                             }
                         }
                     }
@@ -470,6 +477,12 @@ namespace Microsoft.AspNet.Routing.Template
 
                 _uri.Append(value);
                 return true;
+            }
+
+            public void Remove(string literal)
+            {
+                int startIndex = _uri.Length - literal.Length;
+                _uri.Remove(startIndex, literal.Length);
             }
 
             public bool Buffer(string value)
--- a/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
@@ -8,6 +8,7 @@ using System.Globalization;
 using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.AspNet.WebUtilities;
+using Microsoft.AspNet.Http.Extensions;
 
 namespace Microsoft.AspNet.Routing.Template
 {
@@ -225,9 +226,17 @@ namespace Microsoft.AspNet.Routing.Template
                         }
                         else
                         {
+                            // If the value is not accepted, it is null or empty value in the 
+                            // middle of the segment. We accept this if the parameter is an
+                            // optional parameter and it is preceded by an optional seperator.
+                            // I this case, we need to remove the optional seperator that we
+                            // have added to the URI
+                            // Example: template = {id}.{format?}. parameters: id=5
+                            // In this case after we have generated "5.", we wont find any value 
+                            // for format, so we remove '.' and generate 5.
                             if (!context.Accept(converted))
                             {
-                                if (part.IsOptional && segment.Parts[j - 1].IsOptionalSeperator)
+                                if (j != 0 && part.IsOptional && segment.Parts[j - 1].IsOptionalSeperator)
                                 {
                                     context.Remove(segment.Parts[j - 1].Text);
                                 }
@@ -481,8 +490,7 @@ namespace Microsoft.AspNet.Routing.Template
 
             public void Remove(string literal)
             {
-                int startIndex = _uri.Length - literal.Length;
-                _uri.Remove(startIndex, literal.Length);
+                _uri.Length -= literal.Length;
             }
 
             public bool Buffer(string value)
--- a/src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
@@ -11,7 +11,6 @@ namespace Microsoft.AspNet.Routing.Template
     {
         private const string SeparatorString = "/";
         private const char SeparatorChar = '/';
-        private const string PeriodString = ".";
 
         private static readonly char[] Delimiters = new char[] { SeparatorChar };
 
@@ -170,16 +169,51 @@ namespace Microsoft.AspNet.Routing.Template
                                          IReadOnlyDictionary<string, object> defaults,
                                          RouteValueDictionary values)
         {
+            var indexOfLastSegment = routeSegment.Parts.Count - 1;
+            if(routeSegment.Parts[indexOfLastSegment].IsOptional && 
+                routeSegment.Parts[indexOfLastSegment - 1].IsOptionalSeperator)
+            {
+                if (MatchComplexSegmentCore(routeSegment, requestSegment, Defaults, values, indexOfLastSegment))
+                {
+                    return true;
+                }
+                else
+                {
+                    if(requestSegment.EndsWith(routeSegment.Parts[indexOfLastSegment-1].Text))
+                    {
+                        return false;
+                    }
+                    
+                    return MatchComplexSegmentCore(
+                        routeSegment, 
+                        requestSegment, 
+                        Defaults, 
+                        values, 
+                        indexOfLastSegment - 2);
+                }
+            }
+            else
+            {
+                return MatchComplexSegmentCore(routeSegment, requestSegment, Defaults, values, indexOfLastSegment);
+            }
+        }
+
+        private bool MatchComplexSegmentCore(TemplateSegment routeSegment,
+                                         string requestSegment,
+                                         IReadOnlyDictionary<string, object> defaults,
+                                         RouteValueDictionary values,
+                                         int indexOfLastSegmentUsed)
+        {
             Debug.Assert(routeSegment != null);
             Debug.Assert(routeSegment.Parts.Count > 1);
 
             // Find last literal segment and get its last index in the string
             var lastIndex = requestSegment.Length;
-            var indexOfLastSegmentUsed = routeSegment.Parts.Count - 1;
+            //var indexOfLastSegmentUsed = routeSegment.Parts.Count - 1;
 
             TemplatePart parameterNeedsValue = null; // Keeps track of a parameter segment that is pending a value
             TemplatePart lastLiteral = null; // Keeps track of the left-most literal we've encountered
-            TemplatePart firstParameterFound = null;
+
             var outValues = new RouteValueDictionary();
 
             while (indexOfLastSegmentUsed >= 0)
@@ -190,12 +224,7 @@ namespace Microsoft.AspNet.Routing.Template
                 if (part.IsParameter)
                 {
                     // Hold on to the parameter so that we can fill it in when we locate the next literal
-                    parameterNeedsValue = part;
-
-                    if (firstParameterFound == null)
-                    {
-                        firstParameterFound = part;
-                    }
+                    parameterNeedsValue = part;                    
                 }
                 else
                 {
@@ -204,7 +233,7 @@ namespace Microsoft.AspNet.Routing.Template
 
                     var startIndex = lastIndex - 1;
                     // If we have a pending parameter subsegment, we must leave at least one character for that
-                    if (parameterNeedsValue != null && !parameterNeedsValue.IsOptional)
+                    if (parameterNeedsValue != null)
                     {
                         startIndex--;
                     }
@@ -219,27 +248,8 @@ namespace Microsoft.AspNet.Routing.Template
                                                                     StringComparison.OrdinalIgnoreCase);
                     if (indexOfLiteral == -1) 
                     {
-                        // Here we are in a situation where we have one less value for the parameters we have and
-                        // the last parameter in the template (the first one we found was optional). Example:
-                        // template: {p1}/{p2}/{p3?}
-                        // request: foo.bar. 
-                        // So we "restart" the loop with secondlast parameter of the template, rather than the last                            
-                        if (firstParameterFound.IsOptional && firstParameterFound.IsFollowingAPeriod)
-                        {
-                            lastIndex = requestSegment.Length;
-                            indexOfLastSegmentUsed = routeSegment.Parts.Count - 3;
-
-                            parameterNeedsValue = null;
-                            lastLiteral = null;
-                            outValues = new RouteValueDictionary();
-                            firstParameterFound = null;
-                            continue;
-                        }
-                        else
-                        {
-                            // If we couldn't find this literal index, this segment cannot match
-                            return false;
-                        }                        
+                        // If we couldn't find this literal index, this segment cannot match
+                        return false;          
                     }
 
                     // If the first subsegment is a literal, it must match at the right-most extent of the request URI.
@@ -316,18 +326,23 @@ namespace Microsoft.AspNet.Routing.Template
                 indexOfLastSegmentUsed--;
             }
 
-            foreach (var item in outValues)
-            {
-                values.Add(item.Key, item.Value);
-            }
-
             // If the last subsegment is a parameter, it's OK that we didn't parse all the way to the left extent of
             // the string since the parameter will have consumed all the remaining text anyway. If the last subsegment
             // is a literal then we *must* have consumed the entire text in that literal. Otherwise we end up matching
             // the route "Foo" to the request URI "somethingFoo". Thus we have to check that we parsed the *entire*
             // request URI in order for it to be a match.
-            // This check is related to the check we do earlier in this function for LiteralSubsegments.            
-            return (lastIndex == 0) || routeSegment.Parts[0].IsParameter;
+            // This check is related to the check we do earlier in this function for LiteralSubsegments.
+            if (lastIndex == 0 || routeSegment.Parts[0].IsParameter)
+            {
+                foreach (var item in outValues)
+                {
+                    values.Add(item.Key, item.Value);
+                }
+
+                return true;
+            }
+
+            return false;
         }
     }
 }
--- a/src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
@@ -170,8 +170,18 @@ namespace Microsoft.AspNet.Routing.Template
                                          RouteValueDictionary values)
         {
             var indexOfLastSegment = routeSegment.Parts.Count - 1;
-            if(routeSegment.Parts[indexOfLastSegment].IsOptional && 
-                routeSegment.Parts[indexOfLastSegment - 1].IsOptionalSeperator)
+
+            // We match the request to the template starting at the rightmost parameter
+            // If the last segment of template is optional, then request can match the 
+            // template with or without the last parameter. So we start with regular matching,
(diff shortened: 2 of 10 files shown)


kulmugdha:
@rynowak -- please have a look at this first draft.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -322,10 +323,27 @@ private static bool IsSegmentValid(TemplateParserContext context, TemplateSegmen
...
+                    // This is the last part
+                    if (i == segment.Parts.Count - 1)
+                    {
+                        if(segment.Parts[i-1].Text == PeriodString)
+                        {
+                            part.IsFollowingAPeriod = true;
+                        }
+                        else
+                        {
+                            context.Error = Resources.TemplateRoute_CannotHaveOptionalParameterInMultiSegment;
```
rynowak: Are these error messages still accurate? or do they need to be polished?

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -322,10 +323,27 @@ private static bool IsSegmentValid(TemplateParserContext context, TemplateSegmen
...
                 if (part.IsParameter && part.IsOptional && segment.Parts.Count > 1)
                 {
-                    context.Error = Resources.TemplateRoute_CannotHaveOptionalParameterInMultiSegment;
-                    return false;
+                    // This is the last part
+                    if (i == segment.Parts.Count - 1)
+                    {
+                        if(segment.Parts[i-1].Text == PeriodString)
+                        {
+                            part.IsFollowingAPeriod = true;
```
rynowak: Add a property `IsOptionalSeparator` to `segment.Parts[i-1]` - rather than saying `{foo} is after a period` say instead `. is optional`

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -322,10 +323,27 @@ private static bool IsSegmentValid(TemplateParserContext context, TemplateSegmen
...
                 var part = segment.Parts[i];
+                
                 if (part.IsParameter && part.IsOptional && segment.Parts.Count > 1)
                 {
-                    context.Error = Resources.TemplateRoute_CannotHaveOptionalParameterInMultiSegment;
-                    return false;
+                    // This is the last part
+                    if (i == segment.Parts.Count - 1)
+                    {
+                        if(segment.Parts[i-1].Text == PeriodString)
```
rynowak: spacing

rynowak: `Debug.Assert(segment.Parts[i-1].IsLiteral)`

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -178,6 +179,8 @@ public class TemplateMatcher
 
             TemplatePart parameterNeedsValue = null; // Keeps track of a parameter segment that is pending a value
             TemplatePart lastLiteral = null; // Keeps track of the left-most literal we've encountered
+            TemplatePart firstParameterFound = null;
+            var outValues = new RouteValueDictionary();
```
rynowak: I have a simplifying suggestion here. You can basically leave this implementation untouched and call it twice.

1). Rename `MatchComplexSegment` -> `MatchComplexSegmentCore`
2). Add a new parameter 'int startIndex' to `MatchComplexSegmentCore`
3). Add a new method `MatchComplexSegment` which the same signature as the current code

Logic inside `MatchComplexSegment` (pseudocode):
```
var length = segment.Parts.Length;
if (segment.Parts[length -1].IsOptionalSeparator)
{
     return MatchComplexSegmentCore(..., startIndex: length - 1) || MatchComplextSegmentCore(..., startIndex: length - 3);
}
else
{
    return MatchComplexSegmentCore(..., startIndex: length - 1)
}
```

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -16,6 +16,7 @@ public static class TemplateParser
         private const char EqualsSign = '=';
         private const char QuestionMark = '?';
         private const char Asterisk = '*';
+        private const string PeriodString = ".";
```
rynowak: be cool like the others. Make it a 'char' and drop the `String` in the name.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -322,10 +323,27 @@ private static bool IsSegmentValid(TemplateParserContext context, TemplateSegmen
             for (var i = 0; i < segment.Parts.Count; i++)
             {
                 var part = segment.Parts[i];
+                
                 if (part.IsParameter && part.IsOptional && segment.Parts.Count > 1)
```
rynowak: Should we protect against cases like: `{foo}/.{bar}`?

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -228,6 +228,231 @@ public void Parse_ComplexSegment_LPL()
...
+
+            // Act
+            var actual = TemplateParser.Parse(template);
+
+            // Assert
+            Assert.Equal<RouteTemplate>(expected, actual, new TemplateEqualityComparer());
+        }
+
+        [Fact]
+        public void Parse_ComplexSegment_OptionalParameterFollowingPeriod_Threeparameters()
```
rynowak: Three**P**arameters

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -228,6 +228,231 @@ public void Parse_ComplexSegment_LPL()
...
+
+            // Act
+            var actual = TemplateParser.Parse(template);
+
+            // Assert
+            Assert.Equal<RouteTemplate>(expected, actual, new TemplateEqualityComparer());
+        }
+
+        [Fact]
+        public void Parse_ComplexSegment_ThreeparametersSeperatedByPeriod()
```
rynowak: Three**P**arameters

rynowak:
:watch: 

Test coverage of the matcher and parser is looking pretty good, want to also see the following added:
1). Some good coverage around the parser errors that are possible that are close to this feature
2). A few tests that use TemplateRoute and match/general URLs. Basically the unit tests are good, think of these as integration tests.

*Real* Functional tests can wait for the MVC part of this feature.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
kulmugdha: I need to compare it to a part in the segment which is a string. Then I need to use ToString at that point, which didnt feel right.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
kulmugdha: this particular case seems valid, we need to think about protecting against {foo}/.{bar?} as optional parameter followed by period has special meaning.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
kulmugdha: They definitely need some change.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -228,6 +228,231 @@ public void Parse_ComplexSegment_LPL()
...
+        [InlineData("{p1}.{p2?}.{p3}")]
+        [InlineData("{p1?}.{p2}/{p3}")]
+        [InlineData("{p3}.{p1?}.{p2?}")]
+        [InlineData("{p1}-{p2?}")]
+        [InlineData("{p1}..{p2?}")]
+        [InlineData("{p1}.abc.{p2?}")]
+        public void Parse_ComplexSegment_OptionalParametersSeperatedByPeriod_Invalid(string template)
+        {
+            // Act and Assert
+            Assert.Throws<ArgumentException>(() => TemplateParser.Parse(template));
```
rynowak: These should be validating the message as well.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -228,6 +228,231 @@ public void Parse_ComplexSegment_LPL()
...
+
+        [Theory]        
+        [InlineData("{p1?}.{p2?}")]
+        [InlineData("{p1}.{p2?}.{p3}")]
+        [InlineData("{p1?}.{p2}/{p3}")]
+        [InlineData("{p3}.{p1?}.{p2?}")]
+        [InlineData("{p1}-{p2?}")]
+        [InlineData("{p1}..{p2?}")]
+        [InlineData("{p1}.abc.{p2?}")]
+        public void Parse_ComplexSegment_OptionalParametersSeperatedByPeriod_Invalid(string template)
```
rynowak: would like to a few cases where `.` comes first:
- `.{p1?}` - is this valid?
- `..{p1?}` - is this valid?

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
kulmugdha: I decided against protecting it. It just seems unnecessary restriction to put.

kulmugdha:
@rynowak -- I have incorporated the comments. Please have a look.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -220,14 +220,21 @@ public string BindValues(IDictionary<string, object> acceptedValues)
...
                         }
                         else
                         {
                             if (!context.Accept(converted))
                             {
-                                return null;
+                                if (part.IsOptional && segment.Parts[j - 1].IsOptionalSeperator)
+                                {
+                                    context.Remove(segment.Parts[j - 1].Text);
+                                }
```
rynowak: This deserves a comment explaining it.

rynowak: can we ever get here when `j == 0`? (leading to a bound check failure)

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -472,6 +479,12 @@ public bool Accept(string value)
                 return true;
             }
 
+            public void Remove(string literal)
+            {
+                int startIndex = _uri.Length - literal.Length;
```
rynowak: var

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -472,6 +479,12 @@ public bool Accept(string value)
...
+                _uri.Remove(startIndex, literal.Length);
+            }
```
rynowak: `_uri.Length -= literal.Length` will do what you want. This just moves back the 'end' pointer of the buffer.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -169,16 +169,53 @@ public class TemplateMatcher
...
+            if(routeSegment.Parts[indexOfLastSegment].IsOptional && 
+                routeSegment.Parts[indexOfLastSegment - 1].IsOptionalSeperator)
+            {
+                if (MatchComplexSegmentCore(routeSegment, requestSegment, Defaults, values, indexOfLastSegment))
+                {
+                    return true;
+                }
+                else
+                {
+                    if(requestSegment.EndsWith(routeSegment.Parts[indexOfLastSegment-1].Text))
```
rynowak: space

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -169,16 +169,53 @@ public class TemplateMatcher
                                          IReadOnlyDictionary<string, object> defaults,
                                          RouteValueDictionary values)
         {
+            var indexOfLastSegment = routeSegment.Parts.Count - 1;
+            if(routeSegment.Parts[indexOfLastSegment].IsOptional &&
```
rynowak: space

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -169,16 +169,53 @@ public class TemplateMatcher
                                          IReadOnlyDictionary<string, object> defaults,
                                          RouteValueDictionary values)
         {
+            var indexOfLastSegment = routeSegment.Parts.Count - 1;
```
rynowak: comments please, explain 'why'

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -169,16 +169,53 @@ public class TemplateMatcher
...
+                        indexOfLastSegment - 2);
+                }
+            }
+            else
+            {
+                return MatchComplexSegmentCore(routeSegment, requestSegment, Defaults, values, indexOfLastSegment);
+            }
+        }
+
+        private bool MatchComplexSegmentCore(TemplateSegment routeSegment,
```
rynowak: Please undo all the changes in this method unless they have a reason to be here.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -223,83 +223,7 @@ public class TemplateRouteTest
...
-            // Act
-            await route.RouteAsync(context);
-
-            // Assert
-            Assert.NotNull(routeValues);
-            Assert.False(routeValues.ContainsKey("id"));
-
-            Assert.False(context.RouteData.Values.ContainsKey("id"));
-        }
-
```
rynowak: Why are these tests deleted?

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -1070,7 +994,8 @@ public void GetVirtualPath_Sends_ProvidedValues()
                 .Returns<string>(null);
 
             var route = CreateRoute(target.Object, "{controller}/{action}");
-            var context = CreateVirtualPathContext(new { action = "Store" }, new { Controller = "Home", action = "Blog" });
+            var context = CreateVirtualPathContext(
+                new { action = "Store" }, new { Controller = "Home", action = "Blog" });
```
rynowak: if you're putting a new line in a method call, put each parameter on its own line

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -1352,7 +1280,126 @@ public void GetVirtualPath_InlineConstraints_CompositeConstraint_FromConstructor
...
+
+            var context = CreateVirtualPathContext(
+                values: new { action = "Index", controller = "Home", name = "products" , format = "xml"});
+
+            // Act
+            var path = route.GetVirtualPath(context);
+
+            // Assert
+            Assert.Equal("Home/Index/products.xml", path);
+        }
```
rynowak: Would like to see another variation of this where `format` is not explicitly specified

rynowak:
:watch: product code is looking close to completion. Needs comments and examples.

For tests, still need to see integration tests for URL matching (TemplateRouteTest) and unit tests for URL generation (TemplateBinderTest)

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
kulmugdha: No, in case like this {action}/{name?} -- the context accepts it and we are never here. For other cases, the parser makes sure that in a composite segment the optional parameter is always preceded by optional separator. I'll still add the check.

rynowak: If it's something that can't happen, add an assert `Debug.Assert(...)` and a brief comment explaining that the parser does it. This lets us know if the code breaks, but without having to really *handle* a case that can never happen.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
kulmugdha: The changes I have here are needed. indexOfLastSegmentUsed now comes from parameters. Also twe need to save the original values so if we return false we don't change them. It restore the new values if we are returning true.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
kulmugdha: They are repeated below in the file again. I think they lasted thru a merge as they had wrong name "InlineConstrait".

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
kulmugdha: Do you mean this test GetVirtualPath_OptionalParameter_ParameterNotPresentInTemplate_PresentInValues where format is not in template. Or GetVirtualPath_OptionalParameter_ParameterPresentInValues this where format is not in values or something else?

kulmugdha:
@rynowak -- I have incorporated all the changes I could. Had questions about few that I have asked above. Please have a look.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
kulmugdha: Are you talking about open curly brace? It is in line with the if above...

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
rynowak: yeah you're right.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
rynowak: :+1:

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
rynowak: Sorry, I put this comment on the wrong test.

I'd like to see a variation of `GetVirtualPath_OptionalParameter_ParameterPresentInValuesAndDefaults` where the default value is actually used. (Hint, it should be omitted from the generated link)

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -862,6 +786,55 @@ public class TemplateRouteTest
...
+            // Arrange
+            var route = CreateRoute("{controller}/{action}.{format?}", new { action = "Index" });
+            var context = CreateRouteContext("/Home/Create.");
+
+            // Act
+            await route.RouteAsync(context);
+
+            // Assert
+            Assert.False(context.IsHandled);
+        }
```
rynowak: Would like to see one more here where format has a default value

rynowak:
:watch: I think I answered all your questions, this is getting very close to done!

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
kulmugdha: Yes, I wanted to talk to you about this test case, I had this one and wasn't sure why it was omitted from the link. Is it because it is ambient value and gets invalidated...

rynowak: Default values don't get put in the link unless necessary.

Ex:
`{controller=Home}/{action=Index}` with `new { controller = "Home", action = "About" }` will generate `/Home/About`

`{controller=Home}/{action=Index}` with `new { controller = "Home", action = "Index" }` will generate `/`

kulmugdha:
@rynowak added 2 test cases where format has the default value. I think that completes this.

rynowak:
Looks good! :shipit:

rynowak:
@kulmugdha is there any reason this is still sitting here? Please close if not

PR #142, opened 2015-01-23
Change the namespace for the MapRoute extension

pranavkm:


rynowak:
Should we change `IRouteBuilder` also?

Eilon:
I think we would want only the extension methods to go in the `Builder` namespace. Otherwise Intellisense gets to cluttered.

pranavkm:
Yup, that's what we decided on with this change.

PR #143, opened 2015-01-30
AddRouting patch

jeroenslor:
A patch that can avoid the following exception when using the MapRoute extension and also adds an extra configuration option parameter.

The exception that occurs (also when running the sample):
An exception of type 'System.InvalidOperationException' occurred in Microsoft.Framework.DependencyInjection.dll but was not handled in user code

Additional information: Unable to resolve service for type 'Microsoft.Framework.OptionsModel.IOptions`1[Microsoft.AspNet.Routing.RouteOptions]' while attempting to activate 'Microsoft.AspNet.Routing.DefaultInlineConstraintResolver'.

### DIFF
--- a/src/Microsoft.AspNet.Routing/RoutingServices.cs
+++ b/src/Microsoft.AspNet.Routing/RoutingServices.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
+using System;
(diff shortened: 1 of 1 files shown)


Eilon:
@rynowak please take a look and review/merge/close. Not sure how this relates to the recent huge changes you made for options/configure/setup.

rynowak:
Yeah, I think this is goodness. We just haven't looked at `AddRouting`

rynowak:
87360d861f16fa43400d356748f03669270507d1

PR #149, opened 2015-02-05
Fixing Routing Issue: 136

kulmugdha:
The RegEx inline constraint doesn't take care of Escape characters.

### DIFF
--- a/src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.AspNet.Routing.Template
 {
@@ -17,6 +18,7 @@ namespace Microsoft.AspNet.Routing.Template
         private const char QuestionMark = '?';
         private const char Asterisk = '*';
         private const string PeriodString = ".";
+        private const char CloseBracket = ')';
 
         public static RouteTemplate Parse(string routeTemplate)
         {
@@ -134,22 +136,29 @@ namespace Microsoft.AspNet.Routing.Template
 
             while (true)
             {
-                if (context.Current == Separator)
+                if (context.Current == OpenBrace)
                 {
-                    // This is a dangling open-brace, which is not allowed
-                    context.Error = Resources.TemplateRoute_MismatchedParameter;
-                    return false;
-                }
-                else if (context.Current == OpenBrace)
-                {
-                    // If we see a '{' while parsing a parameter name it's invalid. We'll just accept it for now
-                    // and let the validation code for the name find it.
+                    // This is an open brace inside of a parameter, it has to be escaped
+                    if (context.Next())
+                    {
+                        if (context.Current != OpenBrace)
+                        {
+                            context.Error = Resources.TemplateRoute_UnescapedBrace;
+                            return false;
+                        }
+                    }
+                    else
+                    {
+                        // This is a dangling open-brace, which is not allowed
+                        context.Error = Resources.TemplateRoute_MismatchedParameter;
+                        return false;
+                    }
                 }
                 else if (context.Current == CloseBrace)
                 {
+                    // This is a closed brace inside of a parameter. It needs to be escaped, to be valid inside the parameter
                     if (!context.Next())
                     {
-                        // This is the end of the string - and we have a valid parameter
                         context.Back();
                         break;
                     }
@@ -176,10 +185,11 @@ namespace Microsoft.AspNet.Routing.Template
             }
 
             var rawParameter = context.Capture();
+            var decoded = rawParameter.Replace("}}", "}").Replace("{{", "{");
 
             // At this point, we need to parse the raw name for inline constraint,
             // default values and optional parameters.
-            var templatePart = InlineRouteParameterParser.ParseRouteParameter(rawParameter);
+            var templatePart = InlineRouteParameterParser.ParseRouteParameter(decoded);
 
             if (templatePart.IsCatchAll && templatePart.IsOptional)
             {
@@ -272,7 +282,7 @@ namespace Microsoft.AspNet.Routing.Template
                 }
             }
 
-            var decoded = encoded.Replace("}}", "}").Replace("{{", "}");
+            var decoded = encoded.Replace("}}", "}").Replace("{{", "{");
             if (IsValidLiteral(context, decoded))
             {
                 segment.Parts.Add(TemplatePart.CreateLiteral(decoded));
--- a/src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
@@ -18,7 +18,6 @@ namespace Microsoft.AspNet.Routing.Template
         private const char QuestionMark = '?';
         private const char Asterisk = '*';
         private const string PeriodString = ".";
-        private const char CloseBracket = ')';
 
         public static RouteTemplate Parse(string routeTemplate)
         {
@@ -143,6 +142,7 @@ namespace Microsoft.AspNet.Routing.Template
                     {
                         if (context.Current != OpenBrace)
                         {
+                            // If we see something like "{p1:regex(^\d{3", we will come here.
                             context.Error = Resources.TemplateRoute_UnescapedBrace;
                             return false;
                         }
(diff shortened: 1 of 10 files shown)


kulmugdha:
@rynowak @yishaigalatzer -- Please have a look.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -17,6 +18,7 @@ public static class TemplateParser
         private const char QuestionMark = '?';
         private const char Asterisk = '*';
         private const string PeriodString = ".";
+        private const char CloseBracket = ')';
```
rynowak: That's definitely a Paren, not a bracket

rynowak: Also I don't see any uses of this so you can remove it.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -134,22 +136,29 @@ private static bool ParseParameter(TemplateParserContext context, TemplateSegmen
...
+                        if (context.Current != OpenBrace)
+                        {
+                            context.Error = Resources.TemplateRoute_UnescapedBrace;
+                            return false;
+                        }
+                    }
+                    else
+                    {
+                        // This is a dangling open-brace, which is not allowed
+                        context.Error = Resources.TemplateRoute_MismatchedParameter;
```
rynowak: Consider with all of these things that you can include an example of what it catches. This would be something like: `{param{`. If you collect all the examples of the cases you're catching you can make sure they have something similar in tests. Obvi, I didn't follow my own advice when writing this code, but I wish I had.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -134,22 +136,29 @@ private static bool ParseParameter(TemplateParserContext context, TemplateSegmen
...
+                        return false;
+                    }
                 }
                 else if (context.Current == CloseBrace)
                 {
+                    // This is a closed brace inside of a parameter. It needs to be escaped, to be valid inside the parameter
```
rynowak: check line-length here. This comment isn't 100% accurate. An unescaped brace is **the end** of the parameter, the comment has the implication that it would be invalid.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
kulmugdha: yes you are right, it was added as I was thinking )] as stopping condition.

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -134,22 +136,29 @@ private static bool ParseParameter(TemplateParserContext context, TemplateSegmen
...
                     if (!context.Next())
                     {
-                        // This is the end of the string - and we have a valid parameter
```
rynowak: why remove this? this comment is still true

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -272,7 +282,7 @@ private static bool ParseLiteral(TemplateParserContext context, TemplateSegment
                 }
             }
 
-            var decoded = encoded.Replace("}}", "}").Replace("{{", "}");
+            var decoded = encoded.Replace("}}", "}").Replace("{{", "{");
```
rynowak: OOh looks like we had a bad bug here

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -462,7 +462,33 @@ public void Parse_ComplexSegment_OptionalParameterFollowingPeriod_PeriodAfterSla
...
+        [InlineData(@"{p1:regex(^\d{{1,2}}\/\d{{1,2}}\/\d{{4}}$)}")] // date
+        [InlineData(@"{p1:regex(^\w+\@\w+\.\w+)}")] // email
+        [InlineData(@"{p1:regex(([}}])\w+}")] // Not balanced }
+        [InlineData(@"{p1:regex(([{{(])\w+}")] // Not balanced }
+        public void Parse_RegularExpressions(string template)
+        {
+            // Arrange
+            var expected = new RouteTemplate(new List<TemplateSegment>());
+            expected.Segments.Add(new TemplateSegment());
+            InlineConstraint c = new InlineConstraint(template);
```
rynowak: var

rynowak: This doesn't seem right. The inline constraint shouldn't have ALL the text. It should just have `:regex(([{{(])\w+` (or similar)

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -471,7 +497,34 @@ public void Parse_ComplexSegment_OptionalParameterFollowingPeriod_PeriodAfterSla
             Assert.Equal<RouteTemplate>(expected, actual, new TemplateEqualityComparer());
         }
 
-        [Theory]        
+        [Theory]
+        [InlineData(@"{p1:regex(^\d{{3}}-\d{{3}}-\d{{4}}}$)}")] // extra }
+        [InlineData(@"{p1:regex(^\d{{3}}-\d{{3}}-\d{{4}}$)}}")] // extra } at the end
+        [InlineData(@"{{p1:regex(^\d{{3}}-\d{{3}}-\d{{4}}$)}")] // extra { at the begining
+        [InlineData(@"{p1:regex(([}])\w+}")] // Not escaped }
+        [InlineData(@"{p1:regex(^\d{{3}}-\d{{3}}-\d{{4}$)}")] // Not escaped }
```
rynowak: What happens in a case like this: `{p1:regex(abc}`? (unclosed parenthanses) Eventually an error will be reported but when and what?

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -634,10 +688,7 @@ public void InvalidTemplate_InvalidParameterNameWithOpenBracketThrows()
         {
             ExceptionAssert.Throws<ArgumentException>(
                 () => TemplateParser.Parse("{a}/{a{aa}/{z}"),
```
rynowak: Please include a **valid** case like this `{p{{}` (parameter name = `p{`). Is this a new thing that would not have been valid before.

Also `{p/}` or similar

rynowak: Do the same for default values, as it was not possible to put a reserved character in a default before.

rynowak:
:watch:

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
kulmugdha: I was just thinking of this test as I was adding the comments above. Adding it right now.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
kulmugdha: the template equality code wasnt checking for constrainsts at all. I have added that code.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
kulmugdha: I am guessing it wont fail in parsing and creating inline constraint, I will fail when creating constraint out of inline constraint, I will check.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
kulmugdha: for the cases like {p{{} (parameter name = p{)., the parameter parsing will parse this but parameter name validation will fail as these are not valid parameter names. the only place where parse will return success where these appear in constraint part of the parameter as it is not validated right away.

kulmugdha:
@rynowak I have incorporated the comments.

test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
```diff
@@ -256,6 +256,36 @@ public void ParseRouteParameter_ConstraintWithQuestionMarkInPattern_PatternIsPar
...
+
+            Assert.Single(templatePart.InlineConstraints);
+            Assert.Single(templatePart.InlineConstraints, c => c.Constraint == @"regex(^\d{{3}}-\d{{3}}-\d{{4}}$)");
+        }
+
+        [Fact]
+        public void ParseRouteParameter_ConstraintWithBraces_WithDefaultValue()
+        {
+            // Arrange & Act
+            var templatePart = ParseParameter(@"p1:regex(^\d{{3}}-\d{{3}}-\d{{4}}$)=123-456-7890"); // ssn
```
rynowak: I want to see a test where `{{` appears in the default value. Remember, this was not possible before.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateParserTests.cs
rynowak: :+1:

test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
rynowak: NVM I see you added it elsewhere.

test/Microsoft.AspNet.Routing.Tests/DefaultInlineConstraintResolverTest.cs
```diff
@@ -287,6 +298,21 @@ public void ResolveConstraint_AmbiguousConstructors_Throws()
...
+        [InlineData("int/")]
+        [InlineData("in{t")]
+        public void ResolveConstraint_Invalid_Throws(string constraint)
+        {
+            // Arrange
+            var routeOptions = new RouteOptions();
+            var resolver = GetInlineConstraintResolver(routeOptions);
+
+            // Act & Assert
+            Assert.Null(resolver.ResolveConstraint(constraint));
```
rynowak: I want to see an integration test like this, including the exception that's thrown. It should look like this

`new TemplateRoute("{c}/{a}/{ssn:regex(foo}");`

rynowak:
:watch: almost done here, still looking for one more test (see comment)

kulmugdha:
@rynowak -- I have added the test case. Look for CreateTemplate_InlineConstraint_Regex_Malformed().

rynowak:
:shipit:

kulmugdha:
checked in with 4e5fc2e2dd4f7dd7a8c2fd8a936f6b2a002a1078

PR #150, opened 2015-02-10
Add support for best-effort link-generation

rynowak:
This change adds a feature needed for aspnet/Mvc#302

There's a new option in routing that allows link-generation to proceed
when the route values cannot be validated. The key scenario for this is
during development of an MVC site. Routing will refuse to generate a link
to actions which don't exists, this is a breaking change from the MVC5
behavior. Setting UseBestEffortLinkGeneration will allow routing to return
a value even when we can't match the action.

This option will remain off by default - setting this to on will impact
link-generation in a bunch of scenarios involving areas where we've
improved the logic for MVC6. If you're considering leaving this on outside
of development scenarios, then make sure to be as explicit with route values
as possible (don't rely on ambient values).

Functional tests to follow up in the MVC repo.

### DIFF
--- a/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
@@ -210,12 +210,12 @@ namespace Microsoft.AspNet.Routing.Template
                 return path;
             }
 
-			// If we can produce a value go ahead and do it, the caller can check context.IsBound
-			// to see if the values were validated.
+            // If we can produce a value go ahead and do it, the caller can check context.IsBound
+            // to see if the values were validated.
             path = _binder.BindValues(values.AcceptedValues);
-			context.IsBound = childContext.IsBound;
+            context.IsBound = childContext.IsBound;
 
-			return path;
+            return path;
         }
 
         private VirtualPathContext CreateChildVirtualPathContext(
@@ -276,7 +276,7 @@ namespace Microsoft.AspNet.Routing.Template
                 foreach (var inlineConstraint in parameter.InlineConstraints)
                 {
                     constraintBuilder.AddResolvedConstraint(parameter.Name, inlineConstraint.Constraint);
-                }                
+                }
             }
 
             return constraintBuilder.Build();
--- a/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
@@ -202,26 +202,20 @@ namespace Microsoft.AspNet.Routing.Template
 
             // Validate that the target can accept these values.
             var childContext = CreateChildVirtualPathContext(context, values.AcceptedValues);
+
             var path = _target.GetVirtualPath(childContext);
             if (path != null)
             {
                 // If the target generates a value then that can short circuit.
-                context.IsBound = true;
                 return path;
             }
-            else if (!childContext.IsBound)
-            {
-                // The target has rejected these values.
-                return null;
-            }
 
+			// If we can produce a value go ahead and do it, the caller can check context.IsBound
+			// to see if the values were validated.
             path = _binder.BindValues(values.AcceptedValues);
-            if (path != null)
-            {
-                context.IsBound = true;
-            }
+			context.IsBound = childContext.IsBound;
 
-            return path;
+			return path;
         }
 
         private VirtualPathContext CreateChildVirtualPathContext(
--- a/test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
+++ b/test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Routing.Logging;
 using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
 using Moq;
 using Xunit;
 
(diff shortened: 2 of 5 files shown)


rynowak:
@harshgMSFT ping

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -202,26 +202,20 @@ public virtual string GetVirtualPath(VirtualPathContext context)
...
+			// to see if the values were validated.
             path = _binder.BindValues(values.AcceptedValues);
-            if (path != null)
-            {
-                context.IsBound = true;
-            }
+			context.IsBound = childContext.IsBound;
 
-            return path;
+			return path;
```
harshgMSFT: formatting.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -202,26 +202,20 @@ public virtual string GetVirtualPath(VirtualPathContext context)
 
             // Validate that the target can accept these values.
             var childContext = CreateChildVirtualPathContext(context, values.AcceptedValues);
+
             var path = _target.GetVirtualPath(childContext);
             if (path != null)
             {
                 // If the target generates a value then that can short circuit.
-                context.IsBound = true;
```
harshgMSFT: not sure if I understand this change, if a target generates a value why do we not mark context.IsBound?

test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
```diff
@@ -231,6 +232,300 @@ public void NamedRouteTests_GetNamedRoute_AmbiguousRoutesInCollection_ThrowsForA
...
+            var path = routeCollection.GetVirtualPath(virtualPathContext);
+
+            Assert.Equal("best", path);
+        }
+
+        [Fact]
+        public void GetVirtualPath_NamedRoute_BestEffort_BestMatchInNestedCollection()
+        {
+            // Arrange
+            var bestMatch = CreateNamedRoute("NoMatch", accept: true, matchValue: "bar");
```
harshgMSFT: nit: did you mean bad

test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
```diff
@@ -231,6 +232,300 @@ public void NamedRouteTests_GetNamedRoute_AmbiguousRoutesInCollection_ThrowsForA
...
+
+        [Fact]
+        public void GetVirtualPath_NamedRoute_BestEffort_BestInTopCollection()
+        {
+            // Arrange
+            var bestMatch = CreateNamedRoute("Match", accept: true, matchValue: "best");
+            var noMatch = CreateNamedRoute("NoMatch", accept: true, matchValue: "bad");
+
+            var routeCollection = new RouteCollection();
+            routeCollection.Add(noMatch);
```
harshgMSFT: is this intended? I thought we would add best route in top collection. there seems to be no difference between GetVirtualPath_NamedRoute_BestEffort_BestInTopCollection and GetVirtualPath_NamedRoute_BestEffort_BestMatchInNestedCollection

test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
```diff
@@ -315,36 +615,80 @@ private static INamedRouter CreateNamedRoute(string name, bool accept = false)
...
 
             return new RouteContext(context.Object);
         }
 
-        private static Mock<IRouter> CreateRoute(bool accept = true)
+        private static Mock<IRouter> CreateRoute(bool accept = true, bool match = false, string matchValue = null)
         {
+            if (matchValue == null)
+            {
+                matchValue = "value";
```
harshgMSFT: this seems odd, why couldn't this be a default parameter value?

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -917,7 +917,7 @@ public void GetVirtualPath_RejectedByHandler()
 
             // Assert
             Assert.False(context.IsBound);
-            Assert.Null(path);
+            Assert.Equal("Home", path);
```
harshgMSFT: should we change the name of the test as well?

harshgMSFT:
:watch: for minor comments.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
rynowak: `context.IsBound` means that the values are 'validated' since we're inspecting the values we never set them to validated. 

This is the difference between a 'match' and a 'fallback match' in the new system.

harshgMSFT: :+1: ok.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
rynowak: My VS settings got messed up and inserted some tabs, fixed now.

test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
rynowak: Good catch, that's my mistake.

test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
rynowak: fixed

test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
rynowak: sure

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
rynowak: done

rynowak:
@harshgMSFT updated

harshgMSFT:
:shipit:

PR #152, opened 2015-02-15
Fixing 148

kulmugdha:
Fixing the error message. The error message for malformed template was too complex listing all the errors that can happen in the message. I have separated the message in 2 different messages. 
1. When there is a parameter or a literal following the optional parameter.
2. when optional parameter is preceded by a literal other than period which is not allowed. 
 
 @rynowak @yishaigalatzer please have a look.

### DIFF
--- a/src/Microsoft.AspNet.Routing/Resources.resx
+++ b/src/Microsoft.AspNet.Routing/Resources.resx
@@ -187,9 +187,9 @@
     <value>In a route parameter, '{' and '}' must be escaped with '{{' and '}}'</value>
   </data>
   <data name="TemplateRoute_OptionalParameterCanbBePrecededByPeriod" xml:space="preserve">
-    <value>In the segment '{0}', the optional parameter '{1}' is preceded by an invalid segment '{2}'. Only valid literal to precede an optional parameter is a period (.).</value>
+    <value>In the segment '{0}', the optional parameter '{1}' is preceded by an invalid segment '{2}'.  Only a period (.) can precede an optional parameter.</value>
   </data>
   <data name="TemplateRoute_OptionalParameterHasTobeTheLast" xml:space="preserve">
-    <value>In the segment '{0}', optional parameter '{1}' should be the last parameter. An optional parameter must be at the end of the segment, it was followed by '{2}'</value>
+    <value>An optional parameter must be at the end of the segment. In the segment '{0}', optional parameter '{1}' is followed by '{2}'.</value>
   </data>
 </root>
--- a/src/Microsoft.AspNet.Routing/Resources.resx
+++ b/src/Microsoft.AspNet.Routing/Resources.resx
@@ -153,9 +153,6 @@
   <data name="TemplateRoute_CannotHaveConsecutiveSeparators" xml:space="preserve">
     <value>The route template separator character '/' cannot appear consecutively. It must be separated by either a parameter or a literal value.</value>
   </data>
-  <data name="TemplateRoute_CanHaveOnlyLastParameterOptional_IfFollowingOptionalSeperator" xml:space="preserve">
-    <value>In a path segment that contains more than one section, such as a literal section or a parameter, there can only be one optional parameter. The optional parameter must be the last parameter in the segment and must be preceded by one single period (.).</value>
-  </data>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Routing/Resources.resx
```diff
@@ -189,4 +189,10 @@
   <data name="TemplateRoute_UnescapedBrace" xml:space="preserve">
     <value>In a route parameter, '{' and '}' must be escaped with '{{' and '}}'</value>
   </data>
+  <data name="TemplateRoute_OptionalParameterCanbBePrecededByPeriod" xml:space="preserve">
+    <value>In the complex segment '{0}',  the optional parameter '{1}' is preceded by an invalid segment '{2}'. Only valid literal to precede an optional parameter is a period (.).</value>
```
rynowak: A few suggestions here:
1) Get rid of 'complex' I don't think it's especially meaningful to a user
2) Remove extra space between `{0},` and `the`
3) Change the last sentence here to `Only a period (.) can precede an optional parameter.`

src/Microsoft.AspNet.Routing/Resources.resx
```diff
@@ -189,4 +189,10 @@
...
+  </data>
+  <data name="TemplateRoute_OptionalParameterHasTobeTheLast" xml:space="preserve">
+    <value>In the complex segment '{0}', optional parameter '{1}' should be the last parameter. No literal or parameter is not allowed after optional parameter.</value>
```
rynowak: Suggestions:
1). Get rid of `complex` again
2). Change last sentence to `An optional parameter must be at the end of the segment, it was followed by '{2}'` (include the string)

src/Microsoft.AspNet.Routing/Template/TemplateParser.cs
```diff
@@ -340,26 +340,43 @@ private static bool IsSegmentValid(TemplateParserContext context, TemplateSegmen
...
 
+                        // the optional parameter is preceded by a period
                         if (segment.Parts[i - 1].Text == PeriodString)
                         {
                             segment.Parts[i - 1].IsOptionalSeperator = true;
                         }
                         else
                         {
-                            context.Error = 
-                                Resources.TemplateRoute_CanHaveOnlyLastParameterOptional_IfFollowingOptionalSeperator;
```
rynowak: Can this error message be removed from resources or is it used elsewhere?

rynowak:
:watch:

src/Microsoft.AspNet.Routing/Resources.resx
kulmugdha: With this, the complete message will look like:

`In the segment '{0}', optional parameter '{1}' should be the last parameter. An optional parameter must be at the end of the segment, it was followed by '{2}'`

which doesn't look right with twice mention of optional parameter should be the last. How does this sound?

`An optional parameter must be at the end of the segment. In the segment '{0}', optional parameter '{1}' is followed by '{2}'`

kulmugdha:
@rynowak : Incorporated the comments, and have a question. Please have a look.

src/Microsoft.AspNet.Routing/Resources.resx
```diff
@@ -189,4 +186,10 @@
   <data name="TemplateRoute_UnescapedBrace" xml:space="preserve">
     <value>In a route parameter, '{' and '}' must be escaped with '{{' and '}}'</value>
   </data>
+  <data name="TemplateRoute_OptionalParameterCanbBePrecededByPeriod" xml:space="preserve">
+    <value>In the segment '{0}', the optional parameter '{1}' is preceded by an invalid segment '{2}'. Only valid literal to precede an optional parameter is a period (.).</value>
+  </data>
```
rynowak: This is not fixed, `Only valid literal to precede an optional parameter is a period (.)` is not a complete sentence

kulmugdha:
@rynowak -- Changed the error messages.

rynowak:
:shipit:

kulmugdha:
checked in f9a9b8068161ec6bb131a180cbb8406c693c1b03

PR #155, opened 2015-03-05
Logging API changes

BrennanConroy:


rynowak:
:shipit:

test/Microsoft.AspNet.Routing.Tests/Logging/NullLogger.cs
```diff
@@ -15,7 +15,7 @@ public IDisposable BeginScope(object state)
             return NullDisposable.Instance;
         }
 
-        public void Write(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
+        public void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
```
rynowak: Were these ever added to common? Including that here instead may prevent future pain

BrennanConroy: They were, I'll fix that

Issue #154, opened 2015-03-04
Add data tokens for link generation


danroth27:

ianhong:
3693002b0234a7df2c08be694b1514426e057c16
and 
https://github.com/aspnet/Mvc/commit/27bdec40a6a26c83e8219b94f99c692f15dd126f

Issue #31, opened 2014-04-03
Should GetVirtualPath return PathString


rynowak:
Inside URL generation in MVC we need to do some stuff that PathString already handles. Would be easier if GetVirtualPath just returned PathString.

We have to harden this code against a missing leading slash because 3rd party route code could do it. Using PathString would enforce this.

```
            var context = new VirtualPathContext(_httpContext, _ambientValues, values);

            var path = _router.GetVirtualPath(context);

            PathString pathString;
            if (path.Length > 0 && !path.StartsWith("/", StringComparison.Ordinal))
            {
                pathString = new PathString("/" + path);
            }
            else
            {
                pathString = new PathString(path);
            }

            return _httpContext.Request.PathBase.Add(pathString).Value;

```

ianhong:
@rynowak 
Just to make sure I'm working on #158 to add VirtualPathData and make GetVirtualPath to return an object of it.

public class VirtualDataPath
{
..
        public string VirtualPath        {           get;set        }
}
So, we want to change the type of VirtualPath from string to PathString??

rynowak:
Yes, that's the change :+1:. Hopefully there's not too much test impact here since it has a conversion to string.

ianhong:
ff209f04bb2daded5ae49fae015b6631d8a861d0

PR #158, opened 2015-03-11
Add data tokens for link generation 

ianhong:
@harshgMSFT, @rynowak, and @Eilon 

This is a PR for #154. Please take a look at this.

Thanks

### DIFF
--- a/src/Microsoft.AspNet.Routing/IRouter.cs
+++ b/src/Microsoft.AspNet.Routing/IRouter.cs
@@ -9,6 +9,6 @@ namespace Microsoft.AspNet.Routing
     {
         Task RouteAsync(RouteContext context);
 
-        string GetVirtualPath(VirtualPathContext context);
+        VirtualPathData GetVirtualPath(VirtualPathContext context);
     }
 }
--- a/src/Microsoft.AspNet.Routing/RouteCollection.cs
+++ b/src/Microsoft.AspNet.Routing/RouteCollection.cs
@@ -97,11 +97,11 @@ namespace Microsoft.AspNet.Routing
             }
         }
 
-        public virtual string GetVirtualPath(VirtualPathContext context)
+        public virtual VirtualPathData GetVirtualPath(VirtualPathContext context)
         {
             EnsureOptions(context.Context);
 
-            // If we're using Best-Effort link generation then it means that we'll first look for a route where 
+            // If we're using Best-Effort link generation then it means that we'll first look for a route where
             // the route values are validated (context.IsBound == true). If we can't find a match like that, then
             // we'll return the path from the first route to return one.
             var useBestEffort = _options.UseBestEffortLinkGeneration;
@@ -109,11 +109,11 @@ namespace Microsoft.AspNet.Routing
             if (!string.IsNullOrEmpty(context.RouteName))
             {
                 var isValidated = false;
-                string bestPath = null;
+                VirtualPathData bestPathData = null;
                 INamedRouter matchedNamedRoute;
                 if (_namedRoutes.TryGetValue(context.RouteName, out matchedNamedRoute))
                 {
-                    bestPath = matchedNamedRoute.GetVirtualPath(context);
+                    bestPathData = matchedNamedRoute.GetVirtualPath(context);
                     isValidated = context.IsBound;
                 }
 
@@ -124,13 +124,13 @@ namespace Microsoft.AspNet.Routing
                     // reset because we're sharing the context
                     context.IsBound = false;
 
-                    var path = unnamedRoute.GetVirtualPath(context);
-                    if (path == null)
+                    var pathData = unnamedRoute.GetVirtualPath(context);
+                    if (pathData == null)
                     {
                         continue;
                     }
 
-                    if (bestPath != null)
+                    if (bestPathData != null)
                     {
                         // There was already a previous route which matched the name.
                         throw new InvalidOperationException(
@@ -139,15 +139,15 @@ namespace Microsoft.AspNet.Routing
                     else if (context.IsBound)
                     {
                         // This is the first 'validated' match that we've found.
-                        bestPath = path;
+                        bestPathData = pathData;
                         isValidated = true;
                     }
                     else
                     {
-                        Debug.Assert(bestPath == null);
+                        Debug.Assert(bestPathData == null);
 
                         // This is the first 'unvalidated' match that we've found.
-                        bestPath = path;
+                        bestPathData = pathData;
                         isValidated = false;
                     }
                 }
@@ -155,7 +155,16 @@ namespace Microsoft.AspNet.Routing
                 if (isValidated || useBestEffort)
                 {
                     context.IsBound = isValidated;
-                    return NormalizeVirtualPath(bestPath);
+
+                    if (bestPathData != null)
+                    {
+                        bestPathData = new VirtualPathData(
+                            bestPathData.Router,
+                            NormalizeVirtualPath(bestPathData.VirtualPath),
+                            bestPathData.DataTokens);
+                    }
+
+                    return bestPathData;
                 }
                 else
                 {
@@ -164,13 +173,13 @@ namespace Microsoft.AspNet.Routing
             }
             else
             {
-                string bestPath = null;
+                VirtualPathData bestPathData = null;
                 for (var i = 0; i < Count; i++)
                 {
                     var route = this[i];
 
-                    var path = route.GetVirtualPath(context);
-                    if (path == null)
+                    var pathData = route.GetVirtualPath(context);
+                    if (pathData == null)
                     {
                         continue;
                     }
@@ -178,18 +187,24 @@ namespace Microsoft.AspNet.Routing
                     if (context.IsBound)
                     {
                         // This route has validated route values, short circuit.
-                        return NormalizeVirtualPath(path);
+                        return new VirtualPathData(
+                            pathData.Router,
+                            NormalizeVirtualPath(pathData.VirtualPath),
+                            pathData.DataTokens);
                     }
-                    else if (bestPath == null)
+                    else if (bestPathData == null)
                     {
                         // The values aren't validated, but this is the best we've seen so far
-                        bestPath = path;
+                        bestPathData = pathData;
                     }
                 }
 
                 if (useBestEffort)
                 {
-                    return NormalizeVirtualPath(bestPath);
+                    return new VirtualPathData(
+                        bestPathData.Router,
+                        NormalizeVirtualPath(bestPathData.VirtualPath),
+                        bestPathData.DataTokens);
                 }
                 else
                 {
@@ -222,9 +237,9 @@ namespace Microsoft.AspNet.Routing
             }
 
             return url;
-    }
+        }
 
-    private void EnsureLogger(HttpContext context)
+        private void EnsureLogger(HttpContext context)
         {
             if (_logger == null)
             {
--- a/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
@@ -212,18 +212,22 @@ namespace Microsoft.AspNet.Routing.Template
 
             // If we can produce a value go ahead and do it, the caller can check context.IsBound
             // to see if the values were validated.
-            var tempDataTokens = new RouteValueDictionary();
+
+            // When we still cannot produce a value, this should return null.
+            var tempPath = _binder.BindValues(values.AcceptedValues);
+            if (tempPath == null)
+            {
+                return null;
+            }
+
+            pathData = new VirtualPathData(this, tempPath);
             if (DataTokens != null)
             {
                 foreach (var dataToken in DataTokens)
                 {
-                    tempDataTokens.Add(dataToken.Key, dataToken.Value);
+                    pathData.DataTokens.Add(dataToken.Key, dataToken.Value);
                 }
             }
-            pathData = new VirtualPathData(
-                router: this,
-                virtualPath: _binder.BindValues(values.AcceptedValues),
-                dataTokens: tempDataTokens);
 
             context.IsBound = childContext.IsBound;
 
--- a/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
@@ -180,7 +180,7 @@ namespace Microsoft.AspNet.Routing.Template
             }
         }
 
-        public virtual string GetVirtualPath(VirtualPathContext context)
+        public virtual VirtualPathData GetVirtualPath(VirtualPathContext context)
         {
             var values = _binder.GetValues(context.AmbientValues, context.Values);
             if (values == null)
@@ -203,19 +203,31 @@ namespace Microsoft.AspNet.Routing.Template
             // Validate that the target can accept these values.
             var childContext = CreateChildVirtualPathContext(context, values.AcceptedValues);
 
-            var path = _target.GetVirtualPath(childContext);
-            if (path != null)
+            var pathData = _target.GetVirtualPath(childContext);
+            if (pathData != null)
             {
                 // If the target generates a value then that can short circuit.
-                return path;
+                return pathData;
             }
 
             // If we can produce a value go ahead and do it, the caller can check context.IsBound
             // to see if the values were validated.
-            path = _binder.BindValues(values.AcceptedValues);
+            var tempDataTokens = new RouteValueDictionary();
+            if (DataTokens != null)
+            {
+                foreach (var dataToken in DataTokens)
+                {
+                    tempDataTokens.Add(dataToken.Key, dataToken.Value);
+                }
+            }
+            pathData = new VirtualPathData(
+                router: this,
+                virtualPath: _binder.BindValues(values.AcceptedValues),
+                dataTokens: tempDataTokens);
+
             context.IsBound = childContext.IsBound;
 
-            return path;
+            return pathData;
         }
 
         private VirtualPathContext CreateChildVirtualPathContext(
--- a/src/Microsoft.AspNet.Routing/VirtualPathData.cs
+++ b/src/Microsoft.AspNet.Routing/VirtualPathData.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNet.Routing
     public class VirtualPathData
     {
         private string _virtualPath;
-        private readonly IReadOnlyDictionary<string, object> _dataToken;
+        private readonly IDictionary<string, object> _dataToken;
 
         /// <summary>
         ///  Initializes a new instance of the <see cref="VirtualPathData"/> class.
@@ -20,7 +20,7 @@ namespace Microsoft.AspNet.Routing
         /// <param name="router">The object that is used to generate the URL.</param>
         /// <param name="virtualPath">The generated URL.</param>
         public VirtualPathData([NotNull] IRouter router, string virtualPath)
-            : this(router, virtualPath, dataTokens: RouteValueDictionary.Empty)
+            : this(router, virtualPath, dataTokens: new RouteValueDictionary())
         {
         }
 
@@ -33,9 +33,9 @@ namespace Microsoft.AspNet.Routing
         public VirtualPathData(
             [NotNull] IRouter router,
             string virtualPath,
-            IReadOnlyDictionary<string, object> dataTokens)
+            IDictionary<string, object> dataTokens)
         {
-            _dataToken = dataTokens ?? RouteValueDictionary.Empty;
+            _dataToken = dataTokens ?? new RouteValueDictionary();
             Router = router;
             VirtualPath = virtualPath;
         }
@@ -43,7 +43,7 @@ namespace Microsoft.AspNet.Routing
         /// <summary>
         /// Gets the collection of custom values for the <see cref="Router"/>.
         /// </summary>
-        public IReadOnlyDictionary<string, object> DataTokens
+        public IDictionary<string, object> DataTokens
         {
             get { return _dataToken; }
         }
--- a/src/Microsoft.AspNet.Routing/VirtualPathData.cs
+++ b/src/Microsoft.AspNet.Routing/VirtualPathData.cs
@@ -35,9 +35,17 @@ namespace Microsoft.AspNet.Routing
             string virtualPath,
             IDictionary<string, object> dataTokens)
         {
-            _dataToken = dataTokens ?? new RouteValueDictionary();
             Router = router;
             VirtualPath = virtualPath;
+
+            _dataToken = new RouteValueDictionary();
+            if(dataTokens != null)
+            {
+                foreach(var dataToken in dataTokens)
+                {
+                    _dataToken.Add(dataToken.Key, dataToken.Value);
+                }
+            }
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.Routing/VirtualPathData.cs
+++ b/src/Microsoft.AspNet.Routing/VirtualPathData.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Routing
+{
+    /// <summary>
+    /// Represents information about the route and virtual path that are the result of
+    /// generating a URL with the ASP.NET routing middleware.
+    /// </summary>
+    public class VirtualPathData
+    {
+        private string _virtualPath;
+        private readonly IReadOnlyDictionary<string, object> _dataToken;
+
+        /// <summary>
+        ///  Initializes a new instance of the <see cref="VirtualPathData"/> class.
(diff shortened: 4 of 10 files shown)


msotclas:
Hi __@ianhong__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Youngjune Hong). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/Microsoft.AspNet.Routing/IRouter.cs
```diff
@@ -9,6 +9,6 @@ public interface IRouter
     {
         Task RouteAsync(RouteContext context);
 
-        string GetVirtualPath(VirtualPathContext context);
+        VirtualPathData GetVirtualPath(VirtualPathContext context);
```
ianhong: This interface misses the xml doc like in the other sources in this repo. I think we can add it later when we clean up this repo later. It will make them more consistent.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            Router = router;
+            VirtualPath = virtualPath;
+            _dataToken = dataToken;
+        }
+
+        /// <summary>
+        /// Gets the collection of custom values for the <see cref="Router"/>.
+        /// </summary>
+        public RouteValueDictionary DataTokens
```
ianhong: @harshgMSFT 
In MVC5, the datatoken is used for namespace, and area, but in MVC6, there is no support of namespace, and it uses the attribute routing for area. So, there is no actual usage in framework itself. 
I talked to @Eilon and this is mainly for end users when they want to use datatoken to generate custom URLs.

src/Microsoft.AspNet.Routing/RouteCollection.cs
```diff
@@ -155,7 +155,16 @@ public virtual string GetVirtualPath(VirtualPathContext context)
                 if (isValidated || useBestEffort)
                 {
                     context.IsBound = isValidated;
-                    return NormalizeVirtualPath(bestPath);
+
+                    if(bestPathData != null)
+                    {
+                        bestPathData = new VirtualPathData(
```
ianhong: Created a new VirtualPathData not to change one passed from another router.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -203,19 +203,27 @@ public virtual string GetVirtualPath(VirtualPathContext context)
...
             {
                 // If the target generates a value then that can short circuit.
-                return path;
+                return pathData;
             }
 
             // If we can produce a value go ahead and do it, the caller can check context.IsBound
             // to see if the values were validated.
-            path = _binder.BindValues(values.AcceptedValues);
+            pathData = new VirtualPathData(this, _binder.BindValues(values.AcceptedValues));
```
ianhong: I think we should set use 'this' object for the Router of VirtualPathData, because the _target was not the router that created the virtual path.

src/Microsoft.AspNet.Routing/RouteCollection.cs
```diff
@@ -155,7 +155,16 @@ public virtual string GetVirtualPath(VirtualPathContext context)
                 if (isValidated || useBestEffort)
                 {
                     context.IsBound = isValidated;
-                    return NormalizeVirtualPath(bestPath);
+
+                    if(bestPathData != null)
```
Eilon: Formatting

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
Eilon: Yeah I think that's right. Ultimately it needs to be the object that has the DataTokens on it.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -203,19 +203,27 @@ public virtual string GetVirtualPath(VirtualPathContext context)
...
+            if(DataTokens != null)
```
Eilon: Formatting

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Microsoft.AspNet.Routing
+{
+    /// <summary>
+    /// Represents information about the route and virtual path that are the result of
+    /// generating a URL with the ASP.NET routing middleware.
+    /// </summary>
+    public class VirtualPathData
+    {
+        private string _virtualPath;
+        private readonly RouteValueDictionary _dataToken;
```
Eilon: It looks like Routing now uses either `IDictionary<string, object>` or `IReadOnlyDictionary<string, object>` as its interface contracts and not `RouteValueDictionary`.

test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
```diff
@@ -96,7 +100,7 @@ public class RouteCollectionTest
             Assert.NotNull(values.Routes);
             Assert.Equal(true, values.Handled);
         }
-
+
```
Eilon: Clean whitespace

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -1066,24 +1074,25 @@ public void RouteWithCatchAllRejectsConstraints()
...
             Assert.True(context.IsBound);
-            Assert.NotNull(virtualPath);
-            Assert.Equal("hello/1234", virtualPath);
+            Assert.NotNull(pathData);
+            Assert.Equal("hello/1234", pathData.VirtualPath);
+            Assert.Same(route, pathData.Router);
+            Assert.Empty(pathData.DataTokens);
         }
-
+
```
Eilon: (Remove more whitespace in various places)

Eilon:
I don't think I saw any tests that check what happens if there *are* data tokens... can you add some?

src/Microsoft.AspNet.Routing/VirtualPathData.cs
ianhong: Since its actual type is RouteValueDctionary I used that one. I think it should be IReadOnlyDctionary<string, object> here since there is no need to change the dictionary from user codes.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+        }
+
+        /// <summary>
+        ///  Initializes a new instance of the <see cref="VirtualPathData"/> class.
+        /// </summary>
+        /// <param name="router">The object that is used to generate the URL.</param>
+        /// <param name="virtualPath">The generated URL.</param>
+        /// <param name="dataToken">The collection of custom values.</param>
+        public VirtualPathData(IRouter router, string virtualPath, IReadOnlyDictionary<string, object> dataToken)
```
Eilon: dataToken --> dataTokens

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -203,19 +203,23 @@ public virtual string GetVirtualPath(VirtualPathContext context)
...
+                return pathData;
             }
 
             // If we can produce a value go ahead and do it, the caller can check context.IsBound
             // to see if the values were validated.
-            path = _binder.BindValues(values.AcceptedValues);
+            pathData = new VirtualPathData(
+                router: this,
+                virtualPath:_binder.BindValues(values.AcceptedValues),
+                dataToken: _dataTokens ?? RouteValueDictionary.Empty);
```
Eilon: System.Web.Routing always copied the data tokens, so I think we should do the same. The VirtualPathData is a snapshot in time so we don't want to point to the original reference.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -203,19 +203,23 @@ public virtual string GetVirtualPath(VirtualPathContext context)
...
-                return path;
+                return pathData;
             }
 
             // If we can produce a value go ahead and do it, the caller can check context.IsBound
             // to see if the values were validated.
-            path = _binder.BindValues(values.AcceptedValues);
+            pathData = new VirtualPathData(
+                router: this,
+                virtualPath:_binder.BindValues(values.AcceptedValues),
```
Eilon: Formatting

Eilon:
I still don't think I saw any tests that check what happens if there are data tokens... can you add some?

ianhong:
@Eilon 
Ah..sorry for this. The last push was not for PR, but for backup. I'm stil working on those extra tests. I moved from my old machine to a new one for the name changes.  I should have mentioned it. 
But, thanks for reviewing, it still helps.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
ianhong: Got it. I removed it since now it has IReadOnlyDictionary but if it is a snapshot, I'll change this.

ianhong:
@Eilon 
I updated this PR for another review. Please review this one.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -203,19 +203,31 @@ public virtual string GetVirtualPath(VirtualPathContext context)
...
             // to see if the values were validated.
-            path = _binder.BindValues(values.AcceptedValues);
+            var tempDataTokens = new RouteValueDictionary();
+            if (DataTokens != null)
+            {
+                foreach (var dataToken in DataTokens)
+                {
+                    tempDataTokens.Add(dataToken.Key, dataToken.Value);
+                }
+            }
```
rynowak: There should be a one-liner to do this copy.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -0,0 +1,71 @@
...
+        public IRouter Router { get; set; }
+
+        /// <summary>
+        /// Gets or sets the URL that was generated from the <see cref="Router"/>.
+        /// </summary>
+        public string VirtualPath
+        {
+            get
+            {
+                return _virtualPath ?? string.Empty;
```
rynowak: let this be null if it's null.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -971,6 +973,86 @@ public void GetVirtualPath_Fail()
             Assert.Null(path);
         }
 
+        [Theory]
+        [MemberData("DataTokensTestData")]
+        public void GetVirtualPath_ReturnsDataTokensWhenTargetReturnsVirtualPathData
+            (RouteValueDictionary dataTokens)
```
rynowak: linebreak after open-paren

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -971,6 +973,86 @@ public void GetVirtualPath_Fail()
...
+            {
+                Assert.True(pathData.DataTokens.ContainsKey(dataToken.Key));
+                Assert.Equal(dataToken.Value, pathData.DataTokens[dataToken.Key]);
+            }
+        }
+
+        [Theory]
+        [MemberData("DataTokensTestData")]
+        public void GetVirtualPath_ReturnsDataTokens_WhenTargetReturnsNullVirtualPathData
+            (RouteValueDictionary dataTokens)
```
rynowak: linebreak after paren

test/Microsoft.AspNet.Routing.Tests/VirtualPathDataTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            Assert.Same(path, pathData.VirtualPath);
+            Assert.NotNull(pathData.DataTokens);
+            Assert.Empty(pathData.DataTokens);
+
+            // Arrange
+            var dataTokens = new RouteValueDictionary();
+            dataTokens["TestKey"] = "TestValue";
+
+            // Act
+            pathData = new VirtualPathData(router, path, dataTokens);
```
rynowak: This is a separate test or a data driven test. 99.99% of the time there should be a single 'Act'

rynowak:
:watch: 

Aside from small comments I left here:

1). There should also be a PR for this in MVC since this is going to create breaking API changes. Let's sign off both and push them together

2). We should consider whether or not 'stack of routers' has a role in link generation (see `RouteContext.Routers`). I'm personally on the fence.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
ianhong: @rynowak 
It seems not. I found that from one of the constructors of RouteVauleDictionary, but it takes IDictionary<string, object> instead of IReadOnlyDictionary. Is there any thing you have in mind??

src/Microsoft.AspNet.Routing/VirtualPathData.cs
ianhong: @rynowak 
Returning string.Empty is the original behavior from MVC5.
@Eilon 
I know you wanted to keep the old behavior, but let me know if you are fine with returning null??

ianhong:
@rynowak 
Sure, I sent this PR first to make sure before updating MVC. I'll start updating MVC tomorrow morning. 
By the way, I'm trying to find out what your second comment means. 
I can't see the Routers property from RouteContext. 
https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNet.Routing/RouteContext.cs 

Is there anything I mssied?

src/Microsoft.AspNet.Routing/VirtualPathData.cs
Eilon: Yeah this is the old behavior from System.Web.Routing so I'd like to keep that.

Eilon: Plus it's generally recommended to avoid returning null from a string property.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            _dataToken = dataTokens ?? RouteValueDictionary.Empty;
+            Router = router;
+            VirtualPath = virtualPath;
+        }
+
+        /// <summary>
+        /// Gets the collection of custom values for the <see cref="Router"/>.
+        /// </summary>
+        public IReadOnlyDictionary<string, object> DataTokens
```
Eilon: Looking at System.Web Routing I think this property is meant to be mutable, so IDictionary and not IReadOnlyDictionary.

ianhong: Yes, it was mutable...because the old type was RouteValueDictionary. But, it  extends both of IReadOnly and IDictionary. So, I think we can use either of them. 
Since it should be a snapshot, I thought it should be IReadOnly, plus the DataTokens of TemplateRoute is IReadOnly too. Do you still want to change this type to IDictionary??

Eilon: I think in all the new code the contracts are either IDictionary or IReadOnlyDictionary, never RouteValueDictionary. I didn't see even one place in the code where there's a contract specifying RouteValueDictionary - did you find any?

Regarding snapshots, in this case the snapshots don't need to be read-only. Snapshot here just means that the new object doesn't track changes in the origin object, not that the new object remains immutable.

ianhong: What I meant was in System.Web.Routing it was RouteValueDictionary, not in the new code. I got it.. and I'll change its type to IDictoinary, and now I think I can use the one liner that @rynowak mentioned to copy. :-).

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
rynowak: Add it :+1:

src/Microsoft.AspNet.Routing/VirtualPathData.cs
rynowak: wierd. `string.empty` is a legit result from calling getvirtualpath, so something obviously bogus (`null`) is indistinguishable from something real.

You'll find an instance of this if you look at the updated templateroute code.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -203,19 +203,31 @@ public virtual string GetVirtualPath(VirtualPathContext context)
...
+            pathData = new VirtualPathData(
+                router: this,
+                virtualPath: _binder.BindValues(values.AcceptedValues),
+                dataTokens: tempDataTokens);
```
rynowak: This isn't right. `_binder.BindValues(values.AcceptedValues)` will return `null` when there's no match, that should result in a `pathData = null`.

ianhong: I think you should be correct. It currently returns an empty string instead of null?? I'll change this to return null when _binder.BindValues(values.AccpetedValues) is null.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
Eilon: I think in this case where there's a "result" object that itself is either null (no match) or not-null (match) the situation is different. It will never make sense to have a null VirtualPath string inside a VirtualPathData.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
ianhong: After I made a change as you suggested. I looked into GetVirtualPath method again to add an unit test on it, but ..it seems like _binder.BindValues won't return null in this case because that chance was already filtered out in the following codes:
https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs#L185-L190
Let me know if I missed anything here.

ianhong: But, I'll keep it makes return null for a kind of another protection, but I'm not going to add another unit test if it is the case.

ianhong:
@rynowak 
I updated per your comments. I'll send the reacting PR for MVC soon.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        /// <param name="router">The object that is used to generate the URL.</param>
+        /// <param name="virtualPath">The generated URL.</param>
+        /// <param name="dataTokens">The collection of custom values.</param>
+        public VirtualPathData(
+            [NotNull] IRouter router,
+            string virtualPath,
+            IDictionary<string, object> dataTokens)
+        {
+            _dataToken = dataTokens ?? new RouteValueDictionary();
```
Eilon: This ctor is suppose to copy the contents of the dictionary, not the reference.

ianhong: @Eilon 
Actually, in the codes I showed you on the other day did copy this DataToken in ctor. And In the same codes, I  removed 
https://github.com/aspnet/Routing/blob/fix154/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs#L224-L230

But, I thought you didn't want that. so I reverted them.

I can do copy in this constructor, but I'm wondering if you still keep the following for statement :
https://github.com/aspnet/Routing/blob/fix154/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs#L224-L230

I think this is redundant.

Eilon: The VirtualPathData is a snapshot so that it can be mutated without modifying the original data. Remember, the goal here is to preserve behavior from System.Web's Routing code.

ianhong: Got it. Actually I like moving the responsbility of taking the snapshot to the constructor. It looks simpler than before. I'll also update the following code now can be pathData = new VirtualPathData(this, tempPath, DataToken), to take advantgae of this change. Let me know if you have any other thoughts. 

// In ttps://github.com/aspnet/Routing/blob/fix154/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs#L224-L230
            pathData = new VirtualPathData(this, tempPath); 
224             if (DataTokens != null) 
225             { 
226                 foreach (var dataToken in DataTokens) 
227                 { 
228                     pathData.DataTokens.Add(dataToken.Key, dataToken.Value); 
229                 } 
230             }

ianhong:
@Eilon and @rynowak 
I updated this PR for Eilon's comment.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -0,0 +1,79 @@
...
+        public VirtualPathData(
+            [NotNull] IRouter router,
+            string virtualPath,
+            IDictionary<string, object> dataTokens)
+        {
+            Router = router;
+            VirtualPath = virtualPath;
+
+            _dataToken = new RouteValueDictionary();
+            if(dataTokens != null)
```
Eilon: Formatting

test/Microsoft.AspNet.Routing.Tests/VirtualPathDataTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            dataTokens["TestKey"] = "TestValue";
+
+            // Act
+            var pathData = new VirtualPathData(router, path, dataTokens);
+
+            // Assert
+            Assert.Same(router, pathData.Router);
+            Assert.Same(path, pathData.VirtualPath);
+            Assert.NotNull(pathData.DataTokens);
+            Assert.Equal("TestValue", pathData.DataTokens["TestKey"]);
```
Eilon: Could also assert that `pathData.DataTokens.Count == 1`.

test/Microsoft.AspNet.Routing.Tests/VirtualPathDataTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            Assert.Empty(pathData1.DataTokens);
+
+            Assert.Same(router, pathData2.Router);
+            Assert.Same(path, pathData2.VirtualPath);
+            Assert.NotNull(pathData2.DataTokens);
+            Assert.Empty(pathData2.DataTokens);
+        }
+
+        [Fact]
+        public void Constructor_SetDataTokens()
```
Eilon: Could rename test to `Constructor_CopiesDataTokens`.

ianhong:
Thanks @Eilon 
I forgot to format if..and I updated this PR once more.

test/Microsoft.AspNet.Routing.Tests/VirtualPathDataTests.cs
ianhong: @Eilon 
I think your sign off is important for this PR, since you are the expert on this. And, my another work item #31 depends on this. Please finish this PR whenever possible; I think this gets almost there. :-)

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -971,6 +973,86 @@ public void GetVirtualPath_Fail()
...
+            target
+                .Setup(r => r.GetVirtualPath(It.IsAny<VirtualPathContext>()))
+                .Callback<VirtualPathContext>(c => c.IsBound = true)
+                .Returns(() => new VirtualPathData(target.Object, path, dataTokens));
+
+            var route = CreateRoute(
+                target.Object,
+                "{controller}",
+                defaults: null,
+                dataTokens: dataTokens);
```
rynowak: In this test you have two routes that both share the data tokens.
- If you intended to test the 'merge' scenario then you need to have different data so you can really see that it works. 
- Otherwise please clean this up so we can see what the test intends to cover

rynowak:
:watch: for the one remaining test issue

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
ianhong: @rynowak 
Actually the datatoken passed to Router should be ignored. I'll update this test to make it clear.

ianhong:
@rynowak 
I updated the test to make it clear that the datatoken passed to the router should be ignored when a target returns something.

rynowak:
:shipit:

ianhong:
https://github.com/aspnet/Mvc/commit/27bdec40a6a26c83e8219b94f99c692f15dd126f

and 
3693002b0234a7df2c08be694b1514426e057c16

Issue #159, opened 2015-03-12
Consider ConfigureRouting instead of ConfigureRouteOptions


HaoK:

danroth27:
37dc8581de593aab1e8a535392671f89b40308e3

PR #163, opened 2015-03-16
Removing 'Options' from the 'ConfigureRouteOptions' extension method name.

sornaks:
https://github.com/aspnet/Mvc/issues/2169

### DIFF
--- a/src/Microsoft.AspNet.Routing/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Routing/ServiceCollectionExtensions.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Framework.DependencyInjection
         /// </summary>
         /// <param name="services">The services available in the application.</param>
         /// <param name="setupAction">An action to configure the <see cref="RouteOptions"/>.</param>
-        public static void ConfigureRouteOptions(
+        public static void ConfigureRouting(
             this IServiceCollection services,
             [NotNull] Action<RouteOptions> setupAction)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Routing/ServiceCollectionExtensions.cs
```diff
@@ -17,7 +17,7 @@ public static class ServiceCollectionExtensions
         /// </summary>
         /// <param name="services">The services available in the application.</param>
         /// <param name="setupAction">An action to configure the <see cref="RouteOptions"/>.</param>
-        public static void ConfigureRouteOptions(
+        public static void ConfigureRoute(
```
kichalla: looks like there is another issue which says `ConfigureRouting` (https://github.com/aspnet/Routing/issues/159)

dougbu:
:shipit: but check to see if MusicStore and other repos need to react to this and the MVC update

src/Microsoft.AspNet.Routing/ServiceCollectionExtensions.cs
dougbu: good point @kichalla, `ConfigureRoute` isn't correct because the `setupAction` can affect multiple routes

sornaks: I changed it to `ConfigureRouting` and merged the commit right after the comment.

sornaks:
Hmmm, MusicStore doesn't seem to use it. Let me check other repos..

kichalla:
:shipit:

PR #168, opened 2015-03-20
Fix routing sample

kichalla:
@rynowak @HaoK

Issue:
The routing sample currently throws an exception when run as its unable to resolve the `IOptions<Routing>` service required on `DefaultInlineConstraintResolver`.

Some Info:
I initially suspected that the change in this PR is going to effect MVC as the `AddMvc` extension also adds `AddOptions`.
https://github.com/aspnet/Mvc/blob/015edefa96c2eec7a474489a93a533e10ac74c7e/src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs#L90

But after making a local Routing repo change, I build and ran MVC repo against this change and all tests passed.

BTW, the `AddDataProtection` extension also internally does `AddOptions`.

### DIFF
--- a/src/Microsoft.AspNet.Routing/RoutingServices.cs
+++ b/src/Microsoft.AspNet.Routing/RoutingServices.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Framework.DependencyInjection
     {
         public static IServiceCollection AddRouting(this IServiceCollection services)
         {
+            services.AddOptions();
             services.TryAdd(ServiceDescriptor.Transient<IInlineConstraintResolver, DefaultInlineConstraintResolver>());
             return services;
         }


msotclas:
Hi __@kichalla__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Kiran Challa). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

HaoK:
:shipit:  Yes Routing needs to call AddOptions if its using Options!

rynowak:
:shipit:

HaoK:
So, maybe there should be a test that verifies things are working properly when just calling AddRouting()?

HaoK:
i.e. samples shouldn't be the canary for something like this regressing...

kichalla:
Thanks @HaoK ...sure, makes sense...

PR #172, opened 2015-03-24
- Removed existing logger scopes as we want to minimize the number of sc...

kichalla:
...opes being created.

- Cleaned up tests related to removal of scopes.

@rynowak


rynowak:
Removing the scopes is a good start. I think we could even live without the logging in router-middleware and route-collection as well -- @loudej thoughts?

We want logging when there's an important event that happens - in our code that happens inside `TemplateRoute` pretty much exclusively. The middleware and the collection really just provide aggregation, and that's the kind of thing we want to cut.

rynowak:
I'd also really like to see diversification in the kinds of messages logged by `TemplateRoute` `CreateRouteAsyncValues` right now produces a one-size-fits-all message. Can you come up with a list of possible scenarios for `TemplateRoute` and we can think about which data goes with each one?

kichalla:
Some examples:
1. 	A request resulted in desired response
2. 	A request resulted in 404 response
	- Possibilities:
		- Routes did not match due to template mismatch
		- Routes did not match due to constraints failure.
		- Route matched but application layer explicitly returned a 404 (ex: product not found)
		- Route matched but no controller/action present
3. 	A request resulted in undesired response.
	- Possiblities:
		- Route matched but resulted in an unexpected controller/action getting executed.

src/Microsoft.AspNet.Routing/RouteConstraintMatcher.cs
```diff
@@ -29,30 +30,21 @@ public static class RouteConstraintMatcher
...
-                        logger.WriteValues(new RouteConstraintMatcherMatchValues()
+                        string currentRouteValues = string.Empty;
+                        if (routeValues != null)
                         {
-                            ConstraintKey = kvp.Key,
-                            Constraint = kvp.Value,
-                            Matched = false
-                        });
+                           currentRouteValues  = string.Join(", ", 
+                               routeValues.Select(rv => $"Key={rv.Key};Value={rv.Value}"));
```
rynowak: We don't want to 'encode' values for logging. Just log the value that's relevant to the constraint.

src/Microsoft.AspNet.Routing/RouteConstraintMatcher.cs
```diff
@@ -29,30 +30,21 @@ public static class RouteConstraintMatcher
                 var constraint = kvp.Value;
                 if (!constraint.Match(httpContext, route, kvp.Key, routeValues, routeDirection))
                 {
-                    if (routeDirection.Equals(RouteDirection.IncomingRequest)
-                        && logger.IsEnabled(LogLevel.Verbose))
+                    if (routeDirection.Equals(RouteDirection.IncomingRequest))
                     {
-                        logger.WriteValues(new RouteConstraintMatcherMatchValues()
+                        string currentRouteValues = string.Empty;
```
rynowak: var

src/Microsoft.AspNet.Routing/RouterMiddleware.cs
```diff
@@ -30,22 +30,16 @@ public class RouterMiddleware
...
-                await _router.RouteAsync(context);
+            await _router.RouteAsync(context);
 
-                if (_logger.IsEnabled(LogLevel.Verbose))
-                {
-                    _logger.WriteValues(new RouterMiddlewareInvokeValues() { Handled = context.IsHandled });
-                }
+            if (!context.IsHandled)
+            {
+                _logger.LogWarning("Request did not match any route in MVC.");
```
rynowak: This should be verbose not warning, and routing isn't an MVC feature.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -99,84 +99,59 @@ public string RouteTemplate
...
-                    // If we got back a null value set, that means the URI did not match
-                    return;
-                }
+            if (values == null)
+            {
+                // If we got back a null value set, that means the URI did not match
 
-                var oldRouteData = context.RouteData;
+                _logger.LogVerbose(
+                        $"Request did not match the route with name '{Name}' and template '{RouteTemplate}'.");
```
rynowak: this seems good.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -99,84 +99,59 @@ public string RouteTemplate
...
 
-                var newRouteData = new RouteData(oldRouteData);
-                MergeValues(newRouteData.DataTokens, _dataTokens);
-                newRouteData.Routers.Add(_target);
-                MergeValues(newRouteData.Values, values);
+                return;
```
rynowak: Should we log a successful match?

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -99,84 +99,59 @@ public string RouteTemplate
...
+                this,
+                RouteDirection.IncomingRequest,
+                _constraintLogger))
+            {
+                return;
+            }
 
-                    await _target.RouteAsync(context);
+            _logger.LogInformation(
+                $"Request successfully matched the route with name '{Name}' and template '{RouteTemplate}'.");
```
rynowak: This seems good.

rynowak: Should we log the route values? Probably safe to leave it out for now and see if it's necessary later.

rynowak:
Overall seems on the right track :watch:

src/Microsoft.AspNet.Routing/RouterMiddleware.cs
kichalla: Good point :+1:

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
kichalla: :+1: I think we should, for example in the case where route matching succeeds but action selection fails, route values would be useful in diagnosis.

src/Microsoft.AspNet.Routing/RouteConstraintMatcher.cs
```diff
@@ -24,35 +22,30 @@ public static class RouteConstraintMatcher
                 return true;
             }
 
+            logger.LogVerbose("Verifying route constraints");
+            if(constraints != null && constraints.Count == 0)
+            {
+                logger.LogVerbose("No route constraints found.");
+            }
```
rynowak: None of this should be here

src/Microsoft.AspNet.Routing/RouteConstraintMatcher.cs
```diff
@@ -24,35 +22,30 @@ public static class RouteConstraintMatcher
...
+
             foreach (var kvp in constraints)
             {
                 var constraint = kvp.Value;
+
+                logger.LogVerbose("Verifying constraint '{RouteConstraint}' for '{RouteKey}'", constraint, kvp.Key);
```
rynowak: This is tracing, we're not doing tracing.

src/Microsoft.AspNet.Routing/RouteConstraintMatcher.cs
```diff
@@ -24,35 +22,30 @@ public static class RouteConstraintMatcher
...
-                        logger.WriteValues(new RouteConstraintMatcherMatchValues()
-                        {
-                            ConstraintKey = kvp.Key,
-                            Constraint = kvp.Value,
-                            Matched = false
-                        });
+                        logger.LogWarning(
+                            "Route value for '{RouteKey}' did not match the constraint '{RouteConstraint}'.",
+                            kvp.Key,
+                            kvp.Value);
```
rynowak: Can we include the value as well?

rynowak: This is not a warning, it's not really a mistake or a programming error, it should be verbose.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -1,15 +1,14 @@
...
-using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Routing.Logging;
-using Microsoft.AspNet.Routing.Logging.Internal;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.Logging;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
```
rynowak: Sort

src/Microsoft.AspNet.Routing/RouteConstraintMatcher.cs
```diff
@@ -1,12 +1,10 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
 using Microsoft.AspNet.Http;
-using Microsoft.AspNet.Routing.Logging;
-using Microsoft.AspNet.Routing.Logging.Internal;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.Logging;
+using System.Collections.Generic;
```
rynowak: Sort

src/Microsoft.AspNet.Routing/Logging/RouteValuesLogValues.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.Logging;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNet.Routing.Logging
+{
+    public class RouteValuesLogValues : ILogValues
```
rynowak: I get why you're doing this, but it's a deviation from the pattern. Let's get buy-in to this being necessary and good.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -99,84 +98,71 @@ public string RouteTemplate
...
+                return;
+            }
 
-                    return;
-                }
+            _logger.LogVerbose(
+                    "Request successfully matched the template of the route with name '{RouteName}' and template" +
+                    " '{RouteTemplate}'.",
+                    Name,
+                    RouteTemplate);
```
rynowak: I think we should make this be 'information'

rynowak: I think we can skip this.

rynowak:
:watch:

davidfowl:
Well no the builds are failing

rynowak:
@davidfowl stop commenting on PRs and come to the meeting you set up :laughing:

src/Microsoft.AspNet.Routing/RouteConstraintMatcher.cs
```diff
@@ -29,30 +27,21 @@ public static class RouteConstraintMatcher
...
-                        {
-                            ConstraintKey = kvp.Key,
-                            Constraint = kvp.Value,
-                            Matched = false
-                        });
+                        object routeValue;
+                        routeValues.TryGetValue(kvp.Key, out routeValue);
+
+                        logger.LogVerbose(
+                            "Route value '{RouteValue}' of key '{RouteKey}' did not match " +
```
rynowak: of key -> with key

test/Microsoft.AspNet.Routing.Tests/Template/TemplateRouteTest.cs
```diff
@@ -65,132 +82,62 @@ public void CreateTemplate_InlineConstraint_Regex_Malformed()
...
-            Assert.Equal(template, values.Template);
-            Assert.NotNull(values.DefaultValues);
-            Assert.NotNull(values.ProducedValues);
-            Assert.Equal(true, values.MatchedTemplate);
-            Assert.Equal(false, values.MatchedConstraints);
-            Assert.Equal(false, values.Matched);
-            Assert.Equal(context.IsHandled, values.Handled);
+            Assert.Empty(sink.Scopes);
+            Assert.Single(sink.Writes);
+            Assert.Equal(expectedMessage, sink.Writes[0].State?.ToString());
```
rynowak: Let's also verify the key-values here.

rynowak:
:watch: for test improvements

rynowak:
:shipit: 

Think about commonizing the code that you're using to compare dictionaries, this is something that's not trivial in xunit, and you'll be doing a fair bit of it.

kichalla:
Thanks...sure will work on creating an assert for dictionaries...

PR #182, opened 2015-05-01
Fix RouteValueDictionary to handle read-only dictionaries

kichalla:
@rynowak @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
+++ b/src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNet.Routing
         private readonly Dictionary<string, object> _dictionary;
 
         /// <summary>
-        /// Creates an empty RouteValueDictionary.
+        /// Creates an empty <see cref="RouteValueDictionary"/>.
         /// </summary>
         public RouteValueDictionary()
         {
@@ -31,28 +31,32 @@ namespace Microsoft.AspNet.Routing
         }
 
         /// <summary>
-        /// Creates a RouteValueDictionary initialized with the provided input values.
+        /// Creates a <see cref="RouteValueDictionary"/> initialized with the provided input value.
         /// </summary>
-        /// <param name="values">Input values to copy into the dictionary.</param>
-        public RouteValueDictionary([NotNull] IDictionary<string, object> values)
-        {
-            _dictionary = new Dictionary<string, object>(values, StringComparer.OrdinalIgnoreCase);
-        }
-
-        /// <summary>
-        /// Creates a RouteValueDictionary initialized with the provided input values.
-        /// </summary>
-        /// <param name="values">Input values to copy into the dictionary.</param>
+        /// <param name="obj">An object to initialize the dictionary. The value can be of type
+        /// <see cref="IDictionary{TKey, TValue}"/> or <see cref="IReadOnlyDictionary{TKey, TValue}"/> or
+        /// any other object.
+        /// </param>
         /// <remarks>
-        /// The input parameter is interpreted as a set of key-value-pairs where the property names
-        /// are keys, and property values are the values, and copied into the dictionary. Only public
-        /// instance non-index properties are considered.
+        /// If the value is a dictionary, then its entries are copied. Otherwise the object is interpreted as a set
+        /// of key-value-pairs where the property names are keys, and property values are the values, and copied
+        /// into the dictionary. Only public instance non-index properties are considered.
         /// </remarks>
         public RouteValueDictionary(object obj)
             : this()
         {
             if (obj != null)
             {
+                var keyValuePairCollection = obj as IEnumerable<KeyValuePair<string, object>>;
+                if (keyValuePairCollection != null)
+                {
+                    foreach (var kvp in keyValuePairCollection)
+                    {
+                        Add(kvp.Key, kvp.Value);
+                    }
+                    return;
+                }
+
                 var type = obj.GetType();
                 var allProperties = type.GetRuntimeProperties();
 
--- a/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
@@ -370,11 +370,7 @@ namespace Microsoft.AspNet.Routing.Template
 
                 if (_defaults != null)
                 {
-                    _filters = new RouteValueDictionary(defaults);
-                    foreach (var kvp in _defaults)
-                    {
-                        _filters.Add(kvp.Key, kvp.Value);
-                    }
+                    _filters = new RouteValueDictionary(_defaults);
                 }
             }
 
--- a/test/Microsoft.AspNet.Routing.Tests/RouteValueDictionaryTests.cs
+++ b/test/Microsoft.AspNet.Routing.Tests/RouteValueDictionaryTests.cs
@@ -168,6 +168,48 @@ namespace Microsoft.AspNet.Routing.Tests
                 expected);
         }
 
+        [Fact]
+        public void CreateFromReadOnlyDictionary_CopiesValues()
+        {
+            // Arrange
+            var dictionary = new Dictionary<string, object>();
+            dictionary.Add("Name", "James");
+            dictionary.Add("Age", 30);
+            var readonlyDictionary = (IReadOnlyDictionary<string, object>)dictionary;
+
+            // Act
+            var dict = new RouteValueDictionary(readonlyDictionary);
+
+            // Assert
+            Assert.Equal(2, dict.Count);
+            Assert.Equal("James", dict["Name"]);
+            Assert.Equal(30, dict["Age"]);
+        }
+
+        [Fact]
+        public void CreateFromReadOnlyDictionary_ModifyRouteValueDictionary()
+        {
+            // Arrange
+            var dictionary = new Dictionary<string, object>();
+            dictionary.Add("Name", "James");
+            dictionary.Add("Age", 30);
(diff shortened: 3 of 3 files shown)


src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -371,10 +371,6 @@ private class TemplateBindingContext
                 if (_defaults != null)
                 {
                     _filters = new RouteValueDictionary(defaults);
```
rynowak: use `_defaults`

kichalla:
Updated. /cc: @rynowak @Eilon (I was able to build MVC successfully with a local reference to this routing change)

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -31,28 +31,32 @@ public RouteValueDictionary()
...
-        /// <param name="values">Input values to copy into the dictionary.</param>
+        /// <param name="obj">An object to initialize the dictionary. The value can be of type
+        /// <see cref="IDictionary{TKey, TValue}"/> or <see cref="IReadOnlyDictionary{TKey, TValue}"/> or
+        /// any other object.
+        /// </param>
         /// <remarks>
-        /// The input parameter is interpreted as a set of key-value-pairs where the property names
-        /// are keys, and property values are the values, and copied into the dictionary. Only public
-        /// instance non-index properties are considered.
+        /// If the value is not a dictionary, then the object is interpreted as a set of key-value-pairs
```
rynowak: What if the object is a dictionary?

kichalla: Thought the `param` comment should be sufficient?

rynowak: I think it would be nice to explain in the remarks that we copy in the entries. The `<param>` just says that dictionaries are supported.

kichalla: Sure, I update the remarks section now.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -31,28 +31,32 @@ public RouteValueDictionary()
...
-        /// <param name="values">Input values to copy into the dictionary.</param>
+        /// <param name="obj">An object to initialize the dictionary. The value can be of type
+        /// <see cref="IDictionary{TKey, TValue}"/> or <see cref="IReadOnlyDictionary{TKey, TValue}"/> or
+        /// any other object.
+        /// </param>
         /// <remarks>
-        /// The input parameter is interpreted as a set of key-value-pairs where the property names
-        /// are keys, and property values are the values, and copied into the dictionary. Only public
-        /// instance non-index properties are considered.
+        /// If the value is a dictionary, then its entries are copied else the object is interpreted as a set
```
rynowak: Grammar fix:

`its entries are copied else the object is interpreted`
to
`its entries are copied. Otherwise the object is interpreted`

rynowak:
:shipit: after fixing run-on-sentence

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -370,11 +370,7 @@ private class TemplateBindingContext
 
                 if (_defaults != null)
                 {
-                    _filters = new RouteValueDictionary(defaults);
-                    foreach (var kvp in _defaults)
-                    {
-                        _filters.Add(kvp.Key, kvp.Value);
-                    }
+                    _filters = new RouteValueDictionary(_defaults);
```
dougbu: - good you fixed the duplicate entries in the `RouteValueDictionary` instance (though that seems unrelated to the point of this PR :smiley_cat: ).  there's still a bug in that `TemplateBindingContext.DebuggerToString()` dereferences `Filters` unconditionally and that property's left `null` if `_defaults == null`. switch to `?.` and all will be well.
- minor but perhaps worth removing the `_filters` and `_acceptedValues` backing fields in `TemplateBindingContext`.  they are only set in that class's constructor; so, a get-only auto-property works.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -31,28 +31,32 @@ public RouteValueDictionary()
...
-            _dictionary = new Dictionary<string, object>(values, StringComparer.OrdinalIgnoreCase);
-        }
-
-        /// <summary>
-        /// Creates a RouteValueDictionary initialized with the provided input values.
-        /// </summary>
-        /// <param name="values">Input values to copy into the dictionary.</param>
+        /// <param name="obj">An object to initialize the dictionary. The value can be of type
+        /// <see cref="IDictionary{TKey, TValue}"/> or <see cref="IReadOnlyDictionary{TKey, TValue}"/> or
+        /// any other object.
```
dougbu: why mention `IDictionary` or `IReadOnlyDictionary` when the code cares about neither?  suggest removing duplication of the `<remarks/>` and reduce this to `An <see cref="object"/> to initialize the <see cref="ViewDataDictionary"/>.`

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -31,28 +31,32 @@ public RouteValueDictionary()
...
+        /// </param>
         /// <remarks>
-        /// The input parameter is interpreted as a set of key-value-pairs where the property names
-        /// are keys, and property values are the values, and copied into the dictionary. Only public
-        /// instance non-index properties are considered.
+        /// If the value is a dictionary, then its entries are copied. Otherwise the object is interpreted as a set
```
dougbu: dictionary -> `dictionary or other <see cref="IEnumerable{KeyValuePair{string, object}}"/>`

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -31,28 +31,32 @@ public RouteValueDictionary()
         }
 
         /// <summary>
-        /// Creates a RouteValueDictionary initialized with the provided input values.
+        /// Creates a <see cref="RouteValueDictionary"/> initialized with the provided input value.
```
dougbu: "input value" -> `<paramref name="obj"/>` (and I think we use "specified" more than "provided")

test/Microsoft.AspNet.Routing.Tests/RouteValueDictionaryTests.cs
```diff
@@ -168,6 +168,48 @@ public void CreateFromObject_MixedCaseThrows()
...
         }
 
+        [Fact]
+        public void CreateFromReadOnlyDictionary_CopiesValues()
+        {
+            // Arrange
+            var dictionary = new Dictionary<string, object>();
+            dictionary.Add("Name", "James");
+            dictionary.Add("Age", 30);
+            var readonlyDictionary = (IReadOnlyDictionary<string, object>)dictionary;
```
dougbu: since the parameter passed to the constructor is always of type `object`, this line makes no difference.  should instead pass a `List<KeyValuePair<string, object>>` or use a state machine method that returns `IEnumerable<KeyValuePair<string, object>>` and uses `yield return`.
same for next test method.

test/Microsoft.AspNet.Routing.Tests/RouteValueDictionaryTests.cs
```diff
@@ -168,6 +168,48 @@ public void CreateFromObject_MixedCaseThrows()
...
+            var dict = new RouteValueDictionary(readonlyDictionary);
+
+            // Assert
+            Assert.Equal(2, dict.Count);
+            Assert.Equal("James", dict["Name"]);
+            Assert.Equal(30, dict["Age"]);
+        }
+
+        [Fact]
+        public void CreateFromReadOnlyDictionary_ModifyRouteValueDictionary()
```
dougbu: what are the expectations of this test?  is it more like `_Add_ModifiesRouteValueDictionary`?

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -31,28 +31,32 @@ public RouteValueDictionary()
...
+        /// </param>
         /// <remarks>
-        /// The input parameter is interpreted as a set of key-value-pairs where the property names
-        /// are keys, and property values are the values, and copied into the dictionary. Only public
-        /// instance non-index properties are considered.
+        /// If the value is a dictionary, then its entries are copied. Otherwise the object is interpreted as a set
+        /// of key-value-pairs where the property names are keys, and property values are the values, and copied
+        /// into the dictionary. Only public instance non-index properties are considered.
         /// </remarks>
         public RouteValueDictionary(object obj)
```
Eilon: Just noticed this param name is `obj`, which looks weird. Can you change it to `values` or `routeValues` or something like that? (In System.Web it was `values`.)

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
kichalla: - That change is necessary as we would end up trying to add duplicate entries(due to the fix of constructor for read-only dictionary that this PR handles)...good catch regarding the `TemplateBindingContext.DebuggerToString()`..will fix it..
- I do not see much value in making this change...will leave it as it is..

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
kichalla: We wanted to add it in the param description as remarks do not show up in intellisense.

PR #184, opened 2015-05-05
Using 'nameof' operator instead of magic strings

hishamco:
Using the "nameof" operator instead of magic string will help us if the variable change in the future for whatever the reason

### DIFF
--- a/src/Microsoft.AspNet.Routing/Constraints/LengthRouteConstraint.cs
+++ b/src/Microsoft.AspNet.Routing/Constraints/LengthRouteConstraint.cs
@@ -24,7 +24,7 @@ namespace Microsoft.AspNet.Routing.Constraints
             if (length < 0)
             {
                 var errorMessage = Resources.FormatArgumentMustBeGreaterThanOrEqualTo(0);
-                throw new ArgumentOutOfRangeException("length", length, errorMessage);
+                throw new ArgumentOutOfRangeException(nameof(length), errorMessage);
             }
 
             MinLength = MaxLength = length;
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNet.Routing/Constraints/LengthRouteConstraint.cs
```diff
@@ -24,7 +24,7 @@ public LengthRouteConstraint(int length)
             if (length < 0)
             {
                 var errorMessage = Resources.FormatArgumentMustBeGreaterThanOrEqualTo(0);
-                throw new ArgumentOutOfRangeException("length", length, errorMessage);
+                throw new ArgumentOutOfRangeException(nameof(length), errorMessage);
```
dougbu: why remove the value from the `Exception`?

src/Microsoft.AspNet.Routing/RouteContext.cs
```diff
@@ -31,7 +31,7 @@ public RouteData RouteData
             {
                 if (value == null)
                 {
-                    throw new ArgumentNullException("value");
+                    throw new ArgumentNullException(nameof(value));
```
dougbu: FYI the changed `ArgumentNullException`s in this PR could all use `[NotNull]` or (for setters) `[param: NotNull]`. But there are few enough of these that there's no reason to hold up this PR.

dougbu:
:watch: for switch back to current `ArgumentOutOfRangeException` constructors

hishamco:
@dougbu I will look to all the PR request that I made, if [NotNull] it will fit I will close them

hishamco:
@dougbu I made a new PR #185 that have all the fixes

Issue #188, opened 2015-06-12
Please add support for AppendTrailingSlash to RouteOptions


danroth27:
Originally reported by @rehansaeed in https://github.com/aspnet/Mvc/issues/2691

MVC 5 had the RouteCollection.AppendTrailingSlash setting. Please add the same option under the new RouteOptions and perhaps another option to RemoveTrailingSlash for those who prefer not to have one. This is a requirement for Search Engine Optimization.

```
services.ConfigureRouting(routeOptions =>
    {
        routeOptions.AppendTrailingSlash = true;
        routeOptions.LowercaseUrls = true;
    });
```

PR #190, opened 2015-06-17
Added support for AppendTrailingSlash in RouteOptions

ajaybhargavb:
Issue - #188 
@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Routing/RouteOptions.cs
+++ b/src/Microsoft.AspNet.Routing/RouteOptions.cs
@@ -15,6 +15,11 @@ namespace Microsoft.AspNet.Routing
         /// </summary>
         public bool LowercaseUrls { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether a trailing slash should be appended to the generated URLs.
+        /// </summary>
+        public bool AppendTrailingSlash { get; set; }
+
         private IDictionary<string, Type> _constraintTypeMap = GetDefaultConstraintMap();
 
         public IDictionary<string, Type> ConstraintMap
--- a/test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
+++ b/test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
@@ -20,16 +20,22 @@ namespace Microsoft.AspNet.Routing
     public class RouteCollectionTest
     {
         [Theory]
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Routing/RouteOptions.cs
```diff
@@ -15,6 +15,11 @@ public class RouteOptions
         /// </summary>
         public bool LowercaseUrls { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether a trailing slash should be appended to the generated URL.
+        /// </summary>
+        public bool AppendTrailingSlash { get; set; }
```
rynowak: `to the generated URL.` -> `to generated URLs.`

This is in the 'options' context, so we should talk about operations broadly.

test/Microsoft.AspNet.Routing.Tests/RouteCollectionTest.cs
```diff
@@ -49,7 +49,38 @@ public class RouteCollectionTest
...
-        
+
+        [Theory]
+        [InlineData(@"Home/Index/23", "/home/index/23/", true)]
+        [InlineData(@"Home/Index/23", "/Home/Index/23", false)]
+        [InlineData(@"Home/Index/23/?Param1=ABC&Param2=Xyz", "/Home/Index/23/?Param1=ABC&Param2=Xyz", true)]
+        [InlineData(@"Home/Index/23?Param1=ABC&Param2=Xyz", "/Home/Index/23/?Param1=ABC&Param2=Xyz", true)]
+        [InlineData(@"Home/Index/23#Param1=ABC&Param2=Xyz", "/Home/Index/23#Param1=ABC&Param2=Xyz", false)]
+        [InlineData(@"Home/Index/23#Param1=ABC&Param2=Xyz", "/Home/Index/23/#Param1=ABC&Param2=Xyz", true)]
+        public void GetVirtualPath_CanAppendTrailingSlash_BasedOnOptions(
```
rynowak: I'd personally write all of these tests as `appendTrailingSlash = true`. We have lots of tests already that **don't** add a URL.

I'd be more interested in some tests that mix lowercase with this feature since it's all in one big method.

rynowak:
:watch:

ajaybhargavb:
Updated.

rynowak:
:shipit:

Issue #191, opened 2015-06-19
Decrease amount of logging for non-matching routes


rynowak:
Add logging to a project and turn on `Verbose`

You'll see output like this for a 404:

```
verbose : [Microsoft.AspNet.Mvc.Routing.InnerAttributeRoute] Request did not mat
ch any attribute route.
verbose : [Microsoft.AspNet.Routing.Template.TemplateRoute] Request did not matc
h the route with name 'default' and template '{controller=Home}/{action=Index}'.

verbose : [Microsoft.AspNet.Builder.RouterMiddleware] Request did not match any
routes.
```

You'll see the messages about `not matching a route` from attribute routing and template route very frequently. In the case that nothing matches, the last message is good enough, we should get rid of the  first two.

ajaybhargavb:
0882a393a4744afc7f6327834f62e2e100fabb22 and aspnet/Mvc@5632a330986b4e915ba351e57e477062d81a58fd

PR #193, opened 2015-06-22
Removed unwanted logging for non-matching routes

ajaybhargavb:
Issue - #191 
@rynowak


rynowak:
:shipit: for the part in the routing repo. MVC also needs to change (attribute routing)

PR #194, opened 2015-07-14
Adding more test cases before removing RegEx for parameter matching

kulmugdha:
@rynowak -- These are just the new test cases that I have added. Will work on the dev code soon.

### DIFF
--- a/test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
+++ b/test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
@@ -57,6 +57,36 @@ namespace Microsoft.AspNet.Routing.Tests
         }
 
         [Fact]
+        public void ParseRouteParameter_ConstraintAndOptional_WithDefaultValue_ParsedCorrectly()
+        {
+            // Arrange & Act
+            var templatePart = ParseParameter(@"param:int=12?");
+
+            // Assert
+            Assert.Equal("param", templatePart.Name);
+            Assert.Equal("12", templatePart.DefaultValue);
+            Assert.True(templatePart.IsOptional);
+
+            Assert.Single(templatePart.InlineConstraints);
+            Assert.Single(templatePart.InlineConstraints, c => c.Constraint == @"int");
+        }
+
+        [Fact]
+        public void ParseRouteParameter_ConstraintAndOptional_WithDefaultValueWithQuestionMark_ParsedCorrectly()
+        {
+            // Arrange & Act
+            var templatePart = ParseParameter(@"param:int=12??");
+
+            // Assert
+            Assert.Equal("param", templatePart.Name);
+            Assert.Equal("12?", templatePart.DefaultValue);
+            Assert.True(templatePart.IsOptional);
+
+            Assert.Single(templatePart.InlineConstraints);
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
```diff
@@ -68,6 +83,33 @@ public void ParseRouteParameter_ConstraintWithArgumentsAndOptional_ParsedCorrect
 
             Assert.Single(templatePart.InlineConstraints);
             Assert.Single(templatePart.InlineConstraints, c => c.Constraint == @"test(\d+)");
+
+            templatePart = ParseParameter(@"param:test(\d+)=abc?");
```
rynowak: **never** do this kind of thing. Convert it to data-driven test (`[Theory]`).

Also the second kind of case you added here is a different scenario right? One of these has a default value and one doesn't.

test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
```diff
@@ -57,6 +57,21 @@ public void ParseRouteParameter_ConstraintAndOptional_ParsedCorrectly()
         }
 
         [Fact]
+        public void ParseRouteParameter_ConstraintAndOptional_WithDefaultValue_ParsedCorrectly()
+        {
+            // Arrange & Act
+            var templatePart = ParseParameter(@"param:int=12?");
```
rynowak: What happens if we double up the question mark? What about one of the other delimiters.

test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
rynowak: Ah, I see this is the same code as the test below, yeah just remove this.

test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
```diff
@@ -86,6 +128,38 @@ public void ParseRouteParameter_ChainedConstraints_ParsedCorrectly()
         }
 
         [Fact]
+        public void ParseRouteParameter_ChainedConstraints_ColonInPattern_ParsedCorrectly()
+        {
+            // Arrange & Act
+            var templatePart = ParseParameter(@"param:test(\d+):test(\w:+)");
```
rynowak: Also interesting to look at doubling up the delimiters here

rynowak:
Would like to see a few tests for doubling up of delimiters. Other than that looks like a good set.

rynowak:
:watch:

test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
kulmugdha: yeah sorry forgot to remove this.

kulmugdha:
@rynowak added a few more tests. Please have a look.

rynowak:
:shipit:

Eilon:
Already merged: https://github.com/aspnet/Routing/commit/ddfe5e9736aed167aa785563c7fc7b203372eb50

PR #196, opened 2015-07-30
Updating InlineRouteParameterParserTests to perform Assert.Equal \

pranavkm:
Assert.Collection instead of Assert.Single.

This gives more accurate failures in case the assertions fail.

### DIFF
--- a/test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
+++ b/test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Routing.Template;
 using Microsoft.Framework.DependencyInjection;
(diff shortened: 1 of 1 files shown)


pranavkm:
cc @rynowak

test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
```diff
@@ -125,10 +124,9 @@ public void ParseRouteParameter_ChainedConstraints_ParsedCorrectly()
             // Assert
             Assert.Equal("param", templatePart.Name);
 
-            Assert.Equal(2, templatePart.InlineConstraints.Count());
-
-            Assert.Single(templatePart.InlineConstraints, c => c.Constraint == @"test(d+)");
-            Assert.Single(templatePart.InlineConstraints, c => c.Constraint == @"test(w+)");
+            Assert.Collection(templatePart.InlineConstraints,
```
rynowak: Have never used this before - pretty rad :heart_eyes:

rynowak:
:shipit:

PR #197, opened 2015-07-30
Remove the use of Regex in the TemplateRouteParser

### LINKED ISSUE #164: Remove the use of Regex in the TemplateRouteParser
image

pranavkm:
Fixes #164

### DIFF
--- a/src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
+++ b/src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
@@ -1,10 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Text.RegularExpressions;
 using Microsoft.AspNet.Routing.Template;
 using Microsoft.Framework.Internal;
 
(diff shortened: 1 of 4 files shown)


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
```diff
@@ -12,82 +9,230 @@ namespace Microsoft.AspNet.Routing
...
+            var currentIndex = startIndex;
 
-            routeParameter = isCatchAll ? routeParameter.Substring(1) : routeParameter;
-            routeParameter = isOptional ? routeParameter.Substring(0, routeParameter.Length - 1) : routeParameter;
+            // Parse parameter name
+            var parameterName = string.Empty;
 
-            var parameterMatch = _parameterRegex.Match(routeParameter);
-            if (!parameterMatch.Success)
+            while (currentIndex <= endIndex)
```
pranavkm: I could theoretically roll this into this into a single state machine along with the constraint parsing - but this made it slightly easier to refactor the lot into a separate method.

pranavkm:
What's the best way to profile startup in isolation? Couldn't quite figure that out.

Eilon:
Remove the NuGet reference too: https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNet.Routing/project.json#L22

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
```diff
@@ -12,82 +9,230 @@ namespace Microsoft.AspNet.Routing
...
+            };
         }
 
-        private static IEnumerable<InlineConstraint> GetInlineConstraints(Group constraintGroup)
+        private enum ParseState
         {
-            var constraints = new List<InlineConstraint>();
+            Start,
+            ParsingName,
+            InsideParantheses,
```
Eilon: Tpyo

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
```diff
@@ -12,82 +9,230 @@ namespace Microsoft.AspNet.Routing
...
             {
-                var defaultValueMatch = defaultValueGroup.Value;
-
-                // Strip out the equal sign at the beginning
-                Debug.Assert(defaultValueMatch.StartsWith("=", StringComparison.Ordinal));
-                return defaultValueMatch.Substring(1);
-            }
-
-            return null;
+                var currentChar = currentIndex > endIndex ? char.MinValue : routeParameter[currentIndex];
```
Eilon: How safe is it to use `\0` as a token? Is that char completely disallowed in routes? I would certainly agree at the very least it's obscure and crazy, but I'm wondering if it's truly safe.

test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
```diff
@@ -192,6 +192,23 @@ public void ParseRouteParameter_ChainedConstraints_WithDefaultValue_DoubleDelimi
                 constraint => Assert.Equal(@"test(\w+)", constraint.Constraint));
         }
 
+        [Theory]
+        [InlineData("=")]
+        [InlineData("+=")]
+        [InlineData(">= || <= || ==")]
+        public void ParseRouteParamter_WithDefaultValue_ContainingDelimiter(string defaultValue)
```
Eilon: Tpyo

Eilon:
I assume that given the plethora of tests we now have for this, that the parsing logic is perfectly sound.

As far as isolating startup perf, hmm... It was @davidfowl who did the initial find. I suppose you could just run the route template parsing code in a loop 1,000,000 times before and after and get a good idea in isolation.

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
pranavkm: An alternative might be to use `-1` in line with [`StringReader.Read`](https://msdn.microsoft.com/en-us/library/ath1fht8\(v=vs.110\).aspx)

davidfowl:
Anyway we can get a before and after profile?

rynowak:
> What's the best way to profile startup in isolation? Couldn't quite figure that out.

The only really meaningful metric here is time until first request. You'll need to set up a site with your changes as compiled binaries, or else all you'll see is compilation.

So basically start the app (let the profiler launch it)
Start a browser
Stop profiling after the first request is serviced

Eilon:
There's startup time but we should still take a quick look at runtime as well to make sure the new code is on par with the old regex.

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
```diff
@@ -12,82 +9,230 @@ namespace Microsoft.AspNet.Routing
...
-
-            // Register inline constraints if present
-            var constraintGroup = parameterMatch.Groups["constraint"];
-            var inlineConstraints = GetInlineConstraints(constraintGroup);
+            string defaultValue = null;
+            if (currentIndex <= endIndex &&
+                routeParameter[currentIndex] == '=')
+            {
+                defaultValue = routeParameter.Substring(currentIndex + 1, endIndex - currentIndex);
+            }
```
rynowak: What about a case like `foo=` (`currentIndex == endIndex`). I suppose you get what you asked for. Is there a test for this case?

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
rynowak: Or to just use nullable

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
```diff
@@ -12,82 +9,230 @@ namespace Microsoft.AspNet.Routing
...
+                            case char.MinValue:
+                                state = ParseState.End;
+                                var constraintText = routeParameter.Substring(startIndex, currentIndex - startIndex);
+                                inlineConstraints.Add(new InlineConstraint(constraintText));
+                                break;
+                            case ')':
+                                // Only consume a ')' token if
+                                // (a) it is the last token
+                                // (b) the next character is the start of the new constraint ':'
+                                // (c) the next character is the start of the default parameter.
```
rynowak: default parameter -> default value

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
```diff
@@ -12,82 +9,230 @@ namespace Microsoft.AspNet.Routing
...
+                                        constraintText = routeParameter.Substring(startIndex, currentIndex - startIndex + 1);
+                                        inlineConstraints.Add(new InlineConstraint(constraintText));
+                                        break;
+                                }
+                                break;
+                            case ':':
+                            case '=':
+                                // In the original implementation, the Regex would've backtracked if it encountered an
+                                // unbalanced opening bracket followed by (not necessarily immediatiely) a delimiter.
+                                // Simply verifying that the parantheses will eventually be balanced should suffice to
```
rynowak: It's not required to be balanced, just closed.

This is totally valid: `{foo:regex(\\(\\(\\(\\()}`

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
Eilon: If that's better, do that.

rynowak:
> There's startup time but we should still take a quick look at runtime as well to make sure the new code is on par with the old regex.

This code runs once on startup unless you're doing something whacky.

Eilon:
@rynowak oh duh :smile:

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
pranavkm: `Substring` handles the case (primarily because we have a 0 length).

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
pranavkm: :+1:

src/Microsoft.AspNet.Routing/InlineRouteParameterParser.cs
pranavkm: Although this did uncover a bug in the current implementation. Adding tests for it.

pranavkm:
> Remove the NuGet reference too: https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNet.Routing/project.json#L22

@Eilon we reference `Regex` in other places in code (RegexRouteConstraint for instance).

rynowak:
:shipit:

Eilon:
@pranavkm oh yeah duh.

PR #200, opened 2015-08-04
React to options rename

HaoK:


HaoK:
Options usage is a bit weird in routing compared to the other repos (namely RouteCollection is doing service locator on HttpContext to get the options out), @divega do we want to proactively address these kinds of issues?

Eilon:
:shipit:

rynowak:
:shipit: 

@HaoK - Routing doesn't live in DI world - everything here is constructable. We could change that, but it makes it harder for users to extend - we haven't seen much (or any) demand for users to DI services into custom routing code.

src/Microsoft.AspNet.Routing/RouteCollection.cs
```diff
@@ -6,11 +6,8 @@
 using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
-using Microsoft.AspNet.Routing.Logging;
-using Microsoft.AspNet.Routing.Logging.Internal;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
-using Microsoft.Framework.Logging;
```
rynowak: :heart_eyes_cat:

HaoK:
3cab10cfd74e83c4d2cee693c63efb0184864151

Issue #181, opened 2015-04-30
Add timeout values to all regex expressions


blowdart:
To avoid regex attacks add a time-out value to every regular expression in the code base.

Eilon:
See this example for a recommended pattern: https://github.com/aspnet/FileSystem/pull/119/files

ajaybhargavb:
ee9945f06dac7176dfa88eda5af07b1ade35ec08

PR #204, opened 2015-09-10
Added timeout to regex

ajaybhargavb:
#181 
@blowdart @Eilon 

I haven't updated the regexes in the test code and sample. Not sure if we want to do that.


pranavkm:
:shipit:

blowdart:
Test code is fine to leave as is.

Samples however should be secured, because we want them to illustrate the right way to do things and because people cut and paste and edit as they know the samples work.

PR #208, opened 2015-09-23
Add access to Optional state of parameters to TemplateRoute

avanderhoorn:
This adds a property to `TemplateRoute` which allows access to the Optional state of parameters of the target route. Will be used in Glimpse to indicate to the user what parameters are optional vs not.

### DIFF
--- a/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.Framework.DependencyInjection;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -91,6 +94,11 @@ public class TemplateRoute : INamedRouter
             get { return _dataTokens; }
         }
 
+        public IReadOnlyDictionary<string, bool> Optionals
```
rynowak: This is much too specific to your needs. Just make `_parsedTemplate` into a public property get-only

rynowak: Bonus points for turning List -> IReadOnly list on that class

avanderhoorn: Ok... so you are happy for that object to be exposed?

avanderhoorn: And what would you like the property to be called `ParsedTemplate`?

rynowak: Yes to both :+1:

avanderhoorn:
Closing this PR in favour of #209

PR #209, opened 2015-09-23
Expose parsed route so more detailed route is available (i.e. optional state)

avanderhoorn:
This adds a property to `TemplateRoute` which allows access to the `ParsedTemplate` state of the target route. Will be used in Glimpse to indicate to the user what parameters are optional vs not, etc

### DIFF
--- a/test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
+++ b/test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNet.Routing.Template;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.OptionsModel;
 using Xunit;
+using System.Linq;
 
 namespace Microsoft.AspNet.Routing.Tests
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNet.Routing.Tests/InlineRouteParameterParserTests.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.OptionsModel;
 using Xunit;
+using System.Linq;
```
avanderhoorn: Because we are now an `IList`, `ToArray` is only available via Linq.

rynowak: undo

rynowak: Ah I see

rynowak:
:shipit:

rynowak:
59b698c

avanderhoorn:
Thanks!

PR #212, opened 2015-09-29
Make dictionary allocations lazy on RouteData

rynowak:
This change makes the allocation of DataTokens and Values on RouteData
lazy, and elides copies when copying an 'empty' RouteData.

In our current architecture this change will eliminate 2 * (N + 1)
dictionary allocations/copies per request, where N is the number of routes
processed. In a large system with lots of attribute routes, this number
could be very significant.

For a small MVC site (ModelBinding, Validation, Views) with one route, it
still shows a modest reduction of dictionary allocations without adding
much complexity.

### DIFF
--- a/src/Microsoft.AspNet.Routing/RouteData.cs
+++ b/src/Microsoft.AspNet.Routing/RouteData.cs
@@ -11,14 +11,16 @@ namespace Microsoft.AspNet.Routing
     /// </summary>
     public class RouteData
     {
+        private Dictionary<string, object> _dataTokens;
+        private RouteValueDictionary _values;
+
         /// <summary>
         /// Creates a new <see cref="RouteData"/> instance.
         /// </summary>
         public RouteData()
         {
-            DataTokens = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+            // Perf: Avoid allocating DataTokens and RouteValues unless needed.
             Routers = new List<IRouter>();
-            Values = new RouteValueDictionary();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


rynowak:
**Before**
image

**After**
image


Note the reduction in dictionaries and the boxed enumerators from copies.

src/Microsoft.AspNet.Routing/RouteData.cs
```diff
@@ -32,24 +34,55 @@ public RouteData(RouteData other)
...
             }
 
-            DataTokens = new Dictionary<string, object>(other.DataTokens, StringComparer.OrdinalIgnoreCase);
             Routers = new List<IRouter>(other.Routers);
-            Values = new RouteValueDictionary(other.Values);
+            
+            // Perf: Avoid allocating DataTokens and RouteValues unless we need to make a copy.
+            if (other._dataTokens != null)
+            {
+                _dataTokens = new Dictionary<string, object>(other.DataTokens, StringComparer.OrdinalIgnoreCase);
```
rynowak: Another options I considered as well is to add a `_parent` field and further defer that copy. I guess I'll try that also and see if it's better.

rynowak: I went ahead and tried this, and was only barely measurably better, but a whole lot more complex to test and code. Leaving this change as-is.

dougbu: Any better if you use `CopyOnWriteDictionary`s?

dougbu:
:shipit:

src/Microsoft.AspNet.Routing/RouteData.cs
rynowak: I'm going to add another change here and measure. The one place the framework interacts with data tokens is in template route. I'm going to make it so that if you don't set data tokens then none of these dictionaries are ever created.

As for the route-values, I don't think copy on write will help too much, because they always get written by template route.

rynowak:
Updated this change to make DataTokens even lazier.

image

Overall reduction looks to be around 4mb / 3001 requests based on overall allocated bytes.

We're down to 125mb total for this app **without** buffer pooling. We were at 175mb about 2 months ago.

Issue #210, opened 2015-09-28
Template matching culture comparison


bbowyersmyth:
Is the template matching intended to be a CurrentCulture based comparison? https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs#L207 
Might be more readable to be explicit about the StringComparison there.

pranavkm:
It probably needs to be Culture insensitive and I guess case insensitive (@rynowak?) . Could you send a PR for this?

rynowak:
Yes, everything that I can think of in routing should be `OrdinalIgnoreCase`.

Eilon:
Agreed, this definitely looks like a bug.

ajaybhargavb:
d5d11eb30e008c2fc22af30dd57aeebdee4deeb1

DaveVdE:
If URLs are case-sensitive, why is literal matching done with `OrdinalIgnoreCase`?

Eilon:
URLs are technically case-sensitive as far as the person *constructing* the URL is concerned, but it's up to a web server to interpret a URL however it wants. So, in the case of ASP.NET, the routing system is entirely case-insensitive.

PR #213, opened 2015-09-29
Fix case sensitive comparison in TemplateMatcher

ajaybhargavb:
https://github.com/aspnet/Routing/issues/210

@pranavkm


Eilon:
This change looks :shipit: to me, though can we add a unit test? (Make sure the test *fails* without your change.)

Also wondering if we did a scan through the rest of routing to check for various EndsWith/Contains/StartsWith and other string APIs to make sure we pass in the right comparer?

ajaybhargavb:
Already did. Couldn't find any. Will add a unit test.

Eilon:
Ping me when the test is in :smile:

ajaybhargavb:
Spoke in person. No need to add a test because this case is only hit for a period (.)

PR #215, opened 2015-10-05
Remove string.Split from routing

rynowak:
This change removes the call to string.Split and a few substrings, and
replaces it with a tokenizer API. The tokenizer isn't really optimized
right now for compute - it should probably be an iterator- but it's a
significant improvement on what we're doing.

### DIFF
--- a/src/Microsoft.AspNet.Routing/Internal/PathSegment.cs
+++ b/src/Microsoft.AspNet.Routing/Internal/PathSegment.cs
@@ -1,0 +1,136 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Routing.Internal
+{
+    public struct PathSegment : IEquatable<PathSegment>, IEquatable<string>
+    {
+        private readonly string _path;
+        private readonly int _start;
+        private readonly int _length;
+
+        private string _segment;
+
+        public PathSegment(string path, int start, int length)
+        {
+            if (path == null)
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+
+            if (start < 1 || start >= path.Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(start));
+            }
+
+            if (length < 0 || start + length > path.Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(length));
+            }
+
+            _path = path;
+            _start = start;
+            _length = length;
+
+            _segment = null;
+        }
+
+        public int Length => _length;
+
+        public string GetRemainingPath()
+        {
+            if (_path == null)
+            {
+                return string.Empty;
+            }
+
+            return _path.Substring(_start);
+        }
+
+        public override bool Equals(object obj)
+        {
+            var other = obj as PathSegment?;
+            if (other == null)
+            {
+                return false;
+            }
+            else
+            {
+                return Equals(other.Value);
+            }
+        }
+
+        public override int GetHashCode()
+        {
+            if (_path == null)
+            {
+                return 0;
+            }
+
+            return StringComparer.OrdinalIgnoreCase.GetHashCode(ToString());
+        }
+
+        public override string ToString()
+        {
+            if (_path == null)
+            {
+                return string.Empty;
+            }
+
+            if (_segment == null)
+            {
+                _segment = _path.Substring(_start, _length);
+            }
+
+            return _segment;
+        }
+
+        public bool Equals(PathSegment other)
+        {
+            if (_path == null)
+            {
+                return other._path == null;
+            }
+
+            if (other._length != _length)
+            {
+                return false;
+            }
+
+            return string.Compare(
+                _path,
+                _start,
+                other._path,
+                other._start,
+                _length,
+                StringComparison.OrdinalIgnoreCase) == 0;
+        }
+
+        public bool Equals(string other)
+        {
+            if (_path == null)
+            {
+                return other == null;
+            }
+
+            if (other.Length != _length)
+            {
+                return false;
+            }
+
+            return string.Compare(_path, _start, other, 0, _length, StringComparison.OrdinalIgnoreCase) == 0;
+        }
+
+        public static bool operator ==(PathSegment x, PathSegment y)
+        {
+            return x.Equals(y);
+        }
+
+        public static bool operator !=(PathSegment x, PathSegment y)
+        {
+            return !x.Equals(y);
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
+++ b/src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
@@ -1,0 +1,203 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.Routing.Internal
+{
+    public struct PathTokenizer : IReadOnlyList<PathSegment>
+    {
+        private readonly string _path;
+        private int _count;
+
+        public PathTokenizer(PathString path)
+        {
+            _path = path.Value;
+            _count = -1;
+        }
+
+        public int Count
+        {
+            get
+            {
+                if (_count == -1)
+                {
+                    // We haven't computed the real count of segments yet.
+                    if (_path.Length == 0)
+                    {
+                        // The empty string has length of 0.
+                        _count = 0;
+                        return _count;
+                    }
+
+                    // A string of length 1 must be "/" - all PathStrings start with '/'
+                    if (_path.Length == 1)
+                    {
+                        // We treat this as empty - there's nothing to parse here for routing, because routing ignores
+                        // a trailing slash.
+                        _count = 0;
+                        return _count;
+                    }
+
+                    // This is a non-trival PathString
+                    _count = 1;
+
+                    // Since a non-empty PathString must begin with a `/`, we can just count the number of occurrences
+                    // of `/` to find the number of segments. However, we don't look at the last character, because
+                    // routing ignores a trailing slash.
+                    for (var i = 1; i < _path.Length - 1; i++)
+                    {
+                        if (_path[i] == '/')
+                        {
+                            _count++;
+                        }
+                    }
+                }
+
+                return _count;
+            }
+        }
+
+        public PathSegment this[int index]
+        {
+            get
+            {
+                if (index >= Count)
+                {
+                    throw new IndexOutOfRangeException();
+                }
+
+
+                var currentSegmentIndex = 0;
+                var currentSegmentStart = 1;
+
+                // Skip the first `/`.
+                var i = 1;
+                while ((i = _path.IndexOf('/', i)) != -1)
+                {
+                    if (currentSegmentIndex++ == index)
+                    {
+                        return new PathSegment(_path, currentSegmentStart, i - currentSegmentStart);
+                    }
+                    else
+                    {
+                        currentSegmentStart = i + 1;
+                        i++;
+                    }
+                }
+
+                // If we get here we're at the end of the string. The implementation of .Count should protect us
+                // from these cases. 
+                Debug.Assert(_path[_path.Length - 1] != '/');
+                Debug.Assert(currentSegmentIndex == index);
+
+                return new PathSegment(_path, currentSegmentStart, _path.Length - currentSegmentStart);
+            }
+        }
+
+        public Enumerator GetEnumerator()
+        {
+            return new Enumerator(this);
+        }
+
+        IEnumerator<PathSegment> IEnumerable<PathSegment>.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -36,18 +38,12 @@ public class TemplateMatcher
...
 
-        public IDictionary<string, object> Match(string requestPath)
+        public IDictionary<string, object> Match(PathString path)
         {
-            if (requestPath == null)
-            {
-                throw new ArgumentNullException(nameof(requestPath));
-            }
-
-            var requestSegments = requestPath.Split(Delimiters);
```
rynowak: The goal of this change is to avoid this call to split, and a few other places where we have to substring.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -67,7 +63,7 @@ public class TemplateMatcher
                     var part = routeSegment.Parts[0];
                     if (part.IsLiteral)
                     {
-                        if (!string.Equals(part.Text, requestSegment, StringComparison.OrdinalIgnoreCase))
+                        if (!requestSegment.Equals(part.Text))
```
rynowak: This now does a comparison against a range on the original string and is allocation free. Since this is a literal segment, we never need to substring this.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -78,7 +74,7 @@ public class TemplateMatcher
 
                         if (part.IsCatchAll)
                         {
-                            var captured = string.Join(SeparatorString, requestSegments, i, requestSegments.Length - i);
+                            var captured = requestSegment.GetRemaingingPath();
```
rynowak: This is much better allocation-wise than joining and putting back together.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -99,7 +95,7 @@ public class TemplateMatcher
                         {
                             if (requestSegment.Length > 0)
                             {
-                                values.Add(part.Name, requestSegment);
+                                values.Add(part.Name, requestSegment.ToString());
```
rynowak: Here we need to substring because this is a 'capture' of the parameter value.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -124,14 +120,14 @@ public class TemplateMatcher
                 }
                 else
                 {
-                    if (!MatchComplexSegment(routeSegment, requestSegment, Defaults, values))
+                    if (!MatchComplexSegment(routeSegment, requestSegment.ToString(), Defaults, values))
```
rynowak: I went ahead and did the substring here. We can optimize further if needed, but complex segments are fairly rare.

src/Microsoft.AspNet.Routing/Template/TemplateRoute.cs
```diff
@@ -115,13 +115,7 @@ public string RouteTemplate
...
             EnsureLoggers(context.HttpContext);
 
-            var requestPath = context.HttpContext.Request.Path.Value;
-
-            if (!string.IsNullOrEmpty(requestPath) && requestPath[0] == '/')
-            {
-                requestPath = requestPath.Substring(1);
-            }
-
+            var requestPath = context.HttpContext.Request.Path;
```
rynowak: Another substring that's not needed.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateMatcherTests.cs
```diff
@@ -19,7 +20,7 @@ public void MatchSingleRoute()
             var matcher = CreateMatcher("{controller}/{action}/{id}");
 
             // Act
-            var match = matcher.Match("Bank/DoAction/123");
+            var match = matcher.Match("/Bank/DoAction/123");
```
rynowak: All these tests updated to have the leading `/` required by `PathString`

test/Microsoft.AspNet.Routing.Tests/Template/TemplateMatcherTests.cs
```diff
@@ -923,7 +924,7 @@ private TemplateMatcher CreateMatcher(string template, object defaults = null)
                 defaults ?? new Dictionary<string, object>());
 
             // Act
-            var match = matcher.Match(path);
+            var match = matcher.Match(new PathString(path));
```
rynowak: This isn't really needed, but it makes things more explicit for someone else that has to debug this in the future.

rynowak:
Working on MVC react PR + perf numbers

src/Microsoft.AspNet.Routing/Infrastructure/PathSegment.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                return string.Empty;
+            }
+
+            return _path.Substring(_start);
+        }
+
+        public override bool Equals(object obj)
+        {
+            var other = obj as PathSegment?;
```
pranavkm:
```C#
if (other is PathSegment)
{
   return Equals((PathSegment)obj);
}

return false;
```

rynowak: gross

rynowak:
Some perf numbers for this change: 
image

This is allocations based on 3000 requests to the MVC plaintext benchmark at aspnet/benchmark.

This reduction in allocations is per-route. So really what we're showing with these numbers would be multiplied many times based on the number of routes.

src/Microsoft.AspNet.Routing/Infrastructure/PathTokenizer.cs
```diff
@@ -0,0 +1,101 @@
...
+                        return _count;
+                    }
+
+                    // This is a non-trival PathString
+                    _count = 1;
+
+                    // Since a non-empty PathString must begin with a `/`, we can just count the number of occurrences
+                    // of `/` to find the number of segments. However, we don't look at the last character, because
+                    // routing ignores a trailing slash.
+                    for (var i = 1; i < _path.Length - 1; i++)
```
pranavkm: Maybe use `IndexOf` in this block?

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
pranavkm: Very cool

src/Microsoft.AspNet.Routing/Infrastructure/PathTokenizer.cs
Eilon: Use it for what? The count is needed here... `_path.Where(c => c == '/').Count()` ? :trollface:

src/Microsoft.AspNet.Routing/Infrastructure/PathSegment.cs
```diff
@@ -0,0 +1,136 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Routing.Infrastructure
```
Eilon: This sounds like something that should really go in a `.Internal` namespace because we don't intend anyone ever to use this outside of Routing's own needs. Might need to do a Routing API review...

src/Microsoft.AspNet.Routing/Infrastructure/PathTokenizer.cs
```diff
@@ -0,0 +1,101 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.Routing.Infrastructure
```
Eilon: Move to `.Internal`

src/Microsoft.AspNet.Routing/Infrastructure/PathTokenizer.cs
rynowak: he's suggesting the `while ((i = _path.IndexOf('/', i) != -1)` pattern

src/Microsoft.AspNet.Routing/Infrastructure/PathSegment.cs
rynowak: `Internal` it is for `PathTokenizer`. Per discussion with @davidfowl - moving `PathSegment` + `StringSegment` to primitives.

src/Microsoft.AspNet.Routing/Internal/PathSegment.cs
```diff
@@ -0,0 +1,136 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Routing.Internal
+{
+    public struct PathSegment : IEquatable<PathSegment>, IEquatable<string>
```
rynowak: Will replace this with `StringSegment` when possible.

rynowak:
@Eilon @pranavkm updated

src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
```diff
@@ -0,0 +1,203 @@
...
+                {
+                    throw new IndexOutOfRangeException();
+                }
+
+
+                var currentSegmentIndex = 0;
+                var currentSegmentStart = 1;
+
+                // Skip the first `/`.
+                var i = 1;
```
pranavkm: `slashIndex` (or something along those lines)

src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
```diff
@@ -0,0 +1,203 @@
...
+        {
+            return GetEnumerator();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        public struct Enumerator : IEnumerator<PathSegment>
```
pranavkm: Where does this get used?

src/Microsoft.AspNet.Routing/Infrastructure/PathTokenizer.cs
Eilon: Eh, that seems rather dubious in terms of both clarity and perf.

src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
rynowak: The next PR :grinning:

src/Microsoft.AspNet.Routing/Infrastructure/PathTokenizer.cs
```diff
@@ -0,0 +1,101 @@
...
+                {
+                    // We haven't computed the real count of segments yet.
+                    if (_path.Length == 0)
+                    {
+                        // The empty string has length of 0.
+                        _count = 0;
+                        return _count;
+                    }
+
+                    // A string of length 1 must be "/" - all PathStrings start with '/'
```
dougbu: Worth `Debug.Assert(_path[0] == '/');` in this block?

src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
```diff
@@ -0,0 +1,203 @@
...
+                    {
+                        // The empty string has length of 0.
+                        _count = 0;
+                        return _count;
+                    }
+
+                    // A string of length 1 must be "/" - all PathStrings start with '/'
+                    if (_path.Length == 1)
+                    {
+                        // We treat this as empty - there's nothing to parse here for routing, because routing ignores
```
dougbu: Worth Debug.Assert(_path[0] == '/'); in this block?

src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
```diff
@@ -0,0 +1,203 @@
...
+
+                    // This is a non-trival PathString
+                    _count = 1;
+
+                    // Since a non-empty PathString must begin with a `/`, we can just count the number of occurrences
+                    // of `/` to find the number of segments. However, we don't look at the last character, because
+                    // routing ignores a trailing slash.
+                    for (var i = 1; i < _path.Length - 1; i++)
+                    {
+                        if (_path[i] == '/')
```
dougbu: Should this blow up if given an empty segment (`"//"`)?

src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
rynowak: sure

src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
rynowak: No, an empty segment is valid.

pranavkm:
:shipit:

dougbu:
:shipit:

src/Microsoft.AspNet.Routing/Internal/PathTokenizer.cs
rynowak: :+1:

rynowak:
Some more numbers for posterity. On 3000 requests of a fairly small webapi style site, these allocations are totally removed by this change:

image

rynowak:
371d4e62da8f4bfd80a35ea2f6d1400b32ed947a

PR #216, opened 2015-10-06
Avoid allocating in TemplateMatcher on failure

rynowak:
This change rejiggers the URL matching algorithm into using a two-pass
system to avoid allocating anything when a URL fails to match a route.

Here's some allocation data based a very simple API site using attribute routing. Allocations of Dictionary and friends for 3000 requests:

**Before**
image

**After**
image

Notice the sharp decrease in the number of `Dictionary<string, object>` and `RouteValueDictionary` allocations.

This change is worth about 2mb out of 67mb total.

### DIFF
--- a/src/Microsoft.AspNet.Routing/Template/RouteTemplate.cs
+++ b/src/Microsoft.AspNet.Routing/Template/RouteTemplate.cs
@@ -41,6 +41,16 @@ namespace Microsoft.AspNet.Routing.Template
 
         public IList<TemplateSegment> Segments { get; private set; }
 
+        public TemplateSegment GetSegment(int index)
+        {
+            if (index < 0)
+            {
+                throw new IndexOutOfRangeException();
+            }
+
+            return index >= Segments.Count ? null : Segments[index];
+        }
+
         private string DebuggerToString()
         {
             return string.Join(SeparatorString, Segments.Select(s => s.DebuggerToString()));
--- a/src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
@@ -40,85 +40,151 @@ namespace Microsoft.AspNet.Routing.Template
 
         public IDictionary<string, object> Match(PathString path)
         {
-            var values = new RouteValueDictionary();
+            var i = 0;
+            var pathTokenizer = new PathTokenizer(path);
 
-            var requestSegments = new PathTokenizer(path);
-            for (var i = 0; i < requestSegments.Count; i++)
+            // Perf: We do a traversal of the request-segments + route-segments twice.
+            //
+            // For most segment-types, we only really need to any work on one of the two passes.
+            //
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Routing/Template/RouteTemplate.cs
```diff
@@ -41,6 +41,16 @@ public RouteTemplate(List<TemplateSegment> segments)
 
         public IList<TemplateSegment> Segments { get; private set; }
 
+        public TemplateSegment GetSegment(int index)
+        {
+            if (index < 0)
+            {
+                throw new IndexOutOfRangeException();
```
pranavkm: `ArgumentOutOfRange`?

rynowak: Sure. Practically speaking though this will never happen. This is mostly here as documentation for the fact that I don't expect it to happen.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -40,126 +40,175 @@ public class TemplateMatcher
...
+                    // provided.
+                    var part = routeSegment.Parts[0];
+                    if (requestSegment.Length > 0)
+                    {
+                        values.Add(part.Name, requestSegment.ToString());
+                    }
+                    else
+                    {
+                        object defaultValue;
+                        if (Defaults.TryGetValue(part.Name, out defaultValue))
```
pranavkm: Shouldn't this always add the defaultValue (if looks like it maps back to https://github.com/aspnet/Routing/pull/216/files?diff=split#diff-9de01c65f25f1a764f72e1641b01410aL151)

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
```diff
@@ -40,126 +40,175 @@ public class TemplateMatcher
...
                 {
-                    values.Add(part.Name, defaultValue);
+                    // There's no default for this (non-optional) parameter so it can't match.
+                    return null;
                 }
-                else if (part.IsOptional)
+            }
+
+            // At this point we've very likely got a match, so start capturing values for real.
+            var values = new RouteValueDictionary();
```
pranavkm: Would this be easier to read if this was refactored in a separate method?

pranavkm:
:shipit:

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
rynowak: I actually like the ugliness of this. I think it's good. If there were two methods, they have to be maintained as one unit, and there's no strong contract (types) making them be the same.

src/Microsoft.AspNet.Routing/Template/TemplateMatcher.cs
rynowak: The case you're talking about is for a catch-all, which we handle separately in this code.

PR #220, opened 2015-10-23
Allow defaults to be null when constructiing TemplateMatcher

mcliment:
The code manages the null value to a default one.
Before c6941e797f1585730b69d088a3b1b0c17ce6afa7 there was a [NotNull] attribute that was enforced but the attribute was not correct, null should be allowed.


rynowak:
:shipit: - looks good to me. I'll merge

rynowak:
Merged 53fd5b5f546a468f49e7d54760cd58d0bc65c839. Thanks again for spotting this!

PR #234, opened 2015-11-21
Add support IEnumerable parameters for query string

grehovm:
I want to suggest a change to support the generation of a url with an array of parameters. It is necessary for pagination with filters (for example, a list of categories).

### DIFF
--- a/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
@@ -2,40 +2,47 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
-using System.IO;
-using System.Text;
 using System.Text.Encodings.Web;
-using Microsoft.AspNet.Http.Extensions;
+using Microsoft.AspNet.Routing.Internal;
+using Microsoft.Extensions.ObjectPool;
 
 namespace Microsoft.AspNet.Routing.Template
 {
     public class TemplateBinder
     {
+        private readonly UrlEncoder _urlEncoder;
+        private readonly ObjectPool<UriBuildingContext> _pool;
+
         private readonly RouteValueDictionary _defaults;
         private readonly RouteValueDictionary _filters;
         private readonly RouteTemplate _template;
-        private readonly UrlEncoder _urlEncoder;
 
         public TemplateBinder(
-            RouteTemplate template,
             UrlEncoder urlEncoder,
+            ObjectPool<UriBuildingContext> pool,
+            RouteTemplate template,
             RouteValueDictionary defaults)
         {
-            if (template == null)
-            {
-                throw new ArgumentNullException(nameof(template));
-            }
-
             if (urlEncoder == null)
             {
                 throw new ArgumentNullException(nameof(urlEncoder));
             }
 
-            _template = template;
+            if (pool == null)
+            {
+                throw new ArgumentNullException(nameof(pool));
+            }
+
+            if (template == null)
+            {
+                throw new ArgumentNullException(nameof(template));
+            }
+
             _urlEncoder = urlEncoder;
+            _pool = pool;
+            _template = template;
             _defaults = defaults;
 
             // Any default that doesn't have a corresponding parameter is a 'filter' and if a value
@@ -189,8 +196,14 @@ namespace Microsoft.AspNet.Routing.Template
         // Step 2: If the route is a match generate the appropriate URI
         public string BindValues(RouteValueDictionary acceptedValues)
         {
-            var context = new UriBuildingContext(_urlEncoder);
+            var context = _pool.Get();
+            var result = BindValues(context, acceptedValues);
+            _pool.Return(context);
+            return result;
+        }
 
+        private string BindValues(UriBuildingContext context, RouteValueDictionary acceptedValues)
+        {
             for (var i = 0; i < _template.Segments.Count; i++)
             {
                 Debug.Assert(context.BufferState == SegmentState.Beginning);
@@ -268,7 +281,7 @@ namespace Microsoft.AspNet.Routing.Template
             }
 
             // Generate the query string from the remaining values
-            var queryBuilder = new QueryBuilder();
+            var wroteFirst = false;
             foreach (var kvp in acceptedValues)
             {
                 if (_defaults != null && _defaults.ContainsKey(kvp.Key))
@@ -283,12 +296,22 @@ namespace Microsoft.AspNet.Routing.Template
                     continue;
                 }
 
-                queryBuilder.Add(kvp.Key, converted);
+                if (!wroteFirst)
+                {
+                    context.Writer.Write('?');
+                    wroteFirst = true;
+                }
+                else
+                {
+                    context.Writer.Write('&');
+                }
+
+                _urlEncoder.Encode(context.Writer, kvp.Key);
+                context.Writer.Write('=');
+                _urlEncoder.Encode(context.Writer, converted);
             }
 
-            var uri = context.GetUri();
-            uri.Append(queryBuilder);
-            return uri.ToString();
+            return context.ToString();
         }
 
         private TemplatePart GetParameter(string name)
@@ -350,7 +373,7 @@ namespace Microsoft.AspNet.Routing.Template
         }
 
         [DebuggerDisplay("{DebuggerToString(),nq}")]
-        private class TemplateBindingContext
+        private struct TemplateBindingContext
         {
             private readonly RouteValueDictionary _defaults;
             private readonly RouteValueDictionary _acceptedValues;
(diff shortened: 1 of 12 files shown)


grehovm:
Test results show that the String is processed incorrectly because it is IEnumerable <char>. It may be worth adding that a special type / interface for generating an array of parameters.

Eilon:
@grehovm thanks for the PR but this is not a feature we are interested in supporting at this time. One of the things we look at in routing is how well it interacts with other systems, such as MVC model binding, which doesn't support this comma-separated syntax. We discussed this within the team and feel that leaving this to individual apps to build would be better, e.g. by passing in `string.Join(",", list)` to the URL generation API. This leaves no ambiguity as to what will happen.

This also avoids issues with custom collection types that both implement `IEnumerable` and override `ToString()`. In this PR (much like the special case of `string`) it would produce the wrong behavior.

grehovm:
As i know MVC doesn't support something like this `url?category=1,2,3` so this PR fix URL generation API to generate `url?category=1&category=2&category=3` and MVC support this. Now, as i know, there is no way to make API to issue such a result. 
To avoid ambiguity need add special type/interface. (i can do it)
Now I have to use a custom function to generate links. But not all third-party solutions admit it.

Eilon:
Ah sorry I misunderstood that part. I think introducing an interface would make this even more complicated - you might as well just format the data yourself.

grehovm:
Now api encode all parameters and make something like this `url?category=1&amp;category=2&amp;category=3`. Third-party solutions(for pagination) often use standart api and don't work in this situation.
I don't understand that complicated in `IUrlEnumerable:IEnumerable{}`.
And replace in my PR `IEnumerable` to `IUrlEnumerable` (remove string workaround).
And it's done.

Eilon:
@grehovm ah, I see what you mean and I see the value in the overall functionality you're describing.

@rynowak @lodejard any thoughts on having some kind of new `IUrlEnumerable` marker interface that allows an `IEnumerable` to be expanded into multiple route values with the same key and multiple values? E.g. an `IUrlEnumerable` called `colors` with values `[red, green, blue]` would generate a query string like `colors=red&colors=green&colors=blue`.

In general I don't like marker interfaces, but I can't think of a better option for this at this time.

rynowak:
@Eilon  - heh

What about just not treating strings as `IEnumerable<>`?

Eilon:
@rynowak there are other types that are both `IEnumerable` and implement `ToString()`. I've been going back and forth on how common that might be, but it doesn't really seem that crazy to me that someone would have done that for the purpose of being "serializable" for routing. I dunno...

rynowak:
The problem with requiring a new interface is that you can't do simple elegant stuff with it anymore.

`@Url.Action("Compare", "Products", new { productId = products.Select(p => p.Id) })`. 

:heart_eyes:

Eilon:
Of course we'd have an extension method :smile: 

```c#
@Url.Action("Compare", "Products", new { productId = products.Select(p => p.Id).AsUrlEnumerable() });
```

:trollface:

grehovm:
I do not know what name to use for the interface and collection. if you tell me then i can fix it. I think this is the most common solution.

grehovm:
and what next?

Eilon:
@grehovm sorry for the delay, my apologies.

I talked to @rynowak a bit more about this and he's convinced me (for now :smile: ) that we should do this feature, but without `IUrlEnumerable`. We should just look for plain old `IEnumerable` (and special case `string` to exclude it).

src/Microsoft.AspNet.Routing/IUrlEnumerable.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Collections;
+
+namespace Microsoft.AspNet.Routing
+{
+    public interface IUrlEnumerable : IEnumerable
```
Eilon: Per my most recent comment, this interface is not needed.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -265,13 +266,28 @@ public string BindValues(IDictionary<string, object> acceptedValues)
...
-                var converted = Convert.ToString(kvp.Value, CultureInfo.InvariantCulture);
-                if (String.IsNullOrEmpty(converted))
+                var values = kvp.Value as IUrlEnumerable;
+                if (values != null)
                 {
-                    continue;
+                    foreach (var value in values)
+                    {
+                        var converted = Convert.ToString(value, CultureInfo.InvariantCulture);
+                        if (String.IsNullOrEmpty(converted))
```
Eilon: Use lowercase `string` instead of `String`.

Eilon: (Fix in other places too.)

src/Microsoft.AspNet.Routing/UrlEnumerableCollection.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Routing
+{
+    public class UrlEnumerableCollection : IUrlEnumerable
```
Eilon: No longer needed.

src/Microsoft.AspNet.Routing/UrlEnumerableExtension.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections;
+
+namespace Microsoft.AspNet.Routing
+{
+    public static class UrlEnumerableExtension
+    {
+        public static IUrlEnumerable AsUrlEnumerable(this IEnumerable self)
```
Eilon: No longer needed.

Eilon:
Would be good to add one or two unit tests to validate the behavior.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -265,13 +266,28 @@ public string BindValues(IDictionary<string, object> acceptedValues)
...
-                var converted = Convert.ToString(kvp.Value, CultureInfo.InvariantCulture);
-                if (String.IsNullOrEmpty(converted))
+                var values = kvp.Value as IEnumerable;
+                if (values != null && !(values is string))
                 {
-                    continue;
+                    foreach (var value in values)
+                    {
+                        var converted = Convert.ToString(value, CultureInfo.InvariantCulture);
+                        if (string.IsNullOrEmpty(converted))
```
pranavkm: Maybe use the inverse the check:
```
if (!string.IsNullOrEmpty(converted))
{
    queryBuilder.Add(kvp.Key, converted);
}
```
Likewise for the other case.

pranavkm:
:watch: for unit tests.

grehovm:
It was up to me, and `String` with a capital letter, too. I simply applied the most popular design pattern "copy-past".
I'm not good at writing tests.

grehovm:
Resolve conflict and use inverse check.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -591,11 +605,11 @@ public BufferValue(string value, bool requiresEncoding)
             {
                 Value = value;
                 RequiresEncoding = requiresEncoding;
-            }
```
Eilon: Fix formatting (indentation)

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -591,11 +605,11 @@ public BufferValue(string value, bool requiresEncoding)
...
+    }
 
             public bool RequiresEncoding { get; }
 
             public string Value { get; }
-        }
+}
```
Eilon: Fix formatting (indentation)

Eilon:
@grehovm sorry for the delays again - I was traveling and was then sick :cold_sweat: 

The code overall looks good (I just have two small formatting comments).

The only things remaining are to:
1. Add at least a few basic unit tests. It should be easy to copy an existing routing test that generates URLs, and pass in some IEnumerable data and verify the result is as expected. E.g. I think [this test](https://github.com/aspnet/Routing/blob/dev/test/Microsoft.AspNet.Routing.Tests/RouteTest.cs#L604-L618) can easily be modified for the new scenario.
2. Squash all the commits into one commit. This will make the commit history be a lot cleaner and easier to follow.

Thanks,
Eilon

grehovm:
1. The Travis broke down not on tests.
2. I don't know how to squash all commits into one. I twice merge changes and resolve conflicts. Otherwise tests and build isn't even start(exceptions).

Eilon:
You can ignore Travis for now, it's having some issues.

To squash commits it depends on which tool you're using. In TortoiseGit it's super easy: view the log, select multiple commits and select "Combine into one commit" and then push back to your branch/fork.

From the command line you can follow the guidance here: http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git/5201642#5201642

It's a very quick command to run and would help keep the history of this repo clean (it avoids seeing all the intermediate changes that have no meaning of their own).

grehovm:
I use TortoiseGit and he did something awful. I will try to re-create a fork and apply my changes.

pranavkm:
Merged in https://github.com/aspnet/Routing/commit/a61a68defcce061191f2bbe5d4a01807908ebc26

PR #238, opened 2015-11-24
Optimize RouteValueDictionary, expose concrete type

rynowak:
This change optimizes allocations by RouteValueDictionary based on usage.

First, implement a struct Enumerator, and expose the concrete RVD type
from all extensibility points. We wanted to try and decouple this code
from RVD originally and use IDictionary everywhere. After doing that we've
found that it allocates an unacceptable number of enumerators.

Secondly, optimize copies of RVD for the case where you're copying an RVC
to another (common case). When doing this we can copy the count to get the
right capacity, and copy the entries without allocating an enumerator.

Lastly, optimize RVD for the case where it's a wrapper around a poco
object. We 'upgrade' to a writable full dictionary if you try to write to
it, or call one of a number of APIs that are uncommonly used. We could
produce optimized versions of things like `Keys` and `CopyTo`.

### DIFF
--- a/src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
+++ b/src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
@@ -18,10 +18,6 @@ namespace Microsoft.AspNet.Routing
         /// </summary>
         internal static readonly IReadOnlyDictionary<string, object> Empty = new RouteValueDictionary();
 
-        private Dictionary<string, object> _dictionary;
-        private readonly PropertyHelper[] _properties;
-        private readonly object _value;
-
         /// <summary>
         /// Creates an empty <see cref="RouteValueDictionary"/>.
         /// </summary>
@@ -47,23 +43,23 @@ namespace Microsoft.AspNet.Routing
             var otherDictionary = values as RouteValueDictionary;
             if (otherDictionary != null)
             {
-                if (otherDictionary._dictionary != null)
+                if (otherDictionary.InnerDictionary != null)
                 {
-                    _dictionary = new Dictionary<string, object>(
-                        otherDictionary._dictionary.Count,
+                    InnerDictionary = new Dictionary<string, object>(
+                        otherDictionary.InnerDictionary.Count,
                         StringComparer.OrdinalIgnoreCase);
 
-                    foreach (var kvp in otherDictionary._dictionary)
+                    foreach (var kvp in otherDictionary.InnerDictionary)
                     {
-                        _dictionary[kvp.Key] = kvp.Value;
+                        InnerDictionary[kvp.Key] = kvp.Value;
                     }
 
                     return;
                 }
-                else if (otherDictionary._properties != null)
+                else if (otherDictionary.Properties != null)
                 {
-                    _properties = otherDictionary._properties;
-                    _value = otherDictionary._value;
+                    Properties = otherDictionary.Properties;
+                    Value = otherDictionary.Value;
                     return;
                 }
                 else
@@ -75,11 +71,11 @@ namespace Microsoft.AspNet.Routing
             var keyValuePairCollection = values as IEnumerable<KeyValuePair<string, object>>;
             if (keyValuePairCollection != null)
             {
-                _dictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+                InnerDictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
 
                 foreach (var kvp in keyValuePairCollection)
                 {
-                    _dictionary[kvp.Key] = kvp.Value;
+                    InnerDictionary[kvp.Key] = kvp.Value;
                 }
 
                 return;
@@ -87,12 +83,18 @@ namespace Microsoft.AspNet.Routing
 
             if (values != null)
             {
-                _properties = PropertyHelper.GetVisibleProperties(values);
-                _value = values;
+                Properties = PropertyHelper.GetVisibleProperties(values);
+                Value = values;
 
                 return;
             }
         }
+
+        private Dictionary<string, object> InnerDictionary { get; set; }
+
+        private PropertyHelper[] Properties { get; }
+
+        private object Value { get; }
 
         /// <inheritdoc />
         public object this[string key]
@@ -117,7 +119,7 @@ namespace Microsoft.AspNet.Routing
                 }
 
                 EnsureWritable();
-                _dictionary[key] = value;
+                InnerDictionary[key] = value;
             }
         }
 
@@ -130,13 +132,7 @@ namespace Microsoft.AspNet.Routing
         public IEqualityComparer<string> Comparer => StringComparer.OrdinalIgnoreCase;
 
         /// <inheritdoc />
-        public int Count
-        {
-            get
-            {
-                return _dictionary?.Count ?? _properties?.Length ?? 0;
-            }
-        }
+        public int Count => InnerDictionary?.Count ?? Properties?.Length ?? 0;
 
         /// <inheritdoc />
         bool ICollection<KeyValuePair<string, object>>.IsReadOnly => false;
@@ -147,7 +143,7 @@ namespace Microsoft.AspNet.Routing
             get
             {
                 EnsureWritable();
-                return _dictionary.Keys;
+                return InnerDictionary.Keys;
             }
         }
 
@@ -156,7 +152,7 @@ namespace Microsoft.AspNet.Routing
             get
             {
                 EnsureWritable();
-                return _dictionary.Keys;
+                return InnerDictionary.Keys;
             }
         }
 
@@ -166,7 +162,7 @@ namespace Microsoft.AspNet.Routing
             get
             {
                 EnsureWritable();
-                return _dictionary.Values;
+                return InnerDictionary.Values;
             }
         }
 
@@ -175,7 +171,7 @@ namespace Microsoft.AspNet.Routing
             get
             {
                 EnsureWritable();
-                return _dictionary.Values;
+                return InnerDictionary.Values;
             }
         }
 
@@ -183,7 +179,7 @@ namespace Microsoft.AspNet.Routing
         void ICollection<KeyValuePair<string, object>>.Add(KeyValuePair<string, object> item)
         {
             EnsureWritable();
-            ((ICollection<KeyValuePair<string, object>>)_dictionary).Add(item);
+            ((ICollection<KeyValuePair<string, object>>)InnerDictionary).Add(item);
         }
 
         /// <inheritdoc />
@@ -195,21 +191,21 @@ namespace Microsoft.AspNet.Routing
             }
 
             EnsureWritable();
-            _dictionary.Add(key, value);
+            InnerDictionary.Add(key, value);
         }
 
         /// <inheritdoc />
         public void Clear()
         {
             EnsureWritable();
-            _dictionary.Clear();
+            InnerDictionary.Clear();
         }
 
         /// <inheritdoc />
         bool ICollection<KeyValuePair<string, object>>.Contains(KeyValuePair<string, object> item)
         {
             EnsureWritable();
-            return ((ICollection<KeyValuePair<string, object>>)_dictionary).Contains(item);
+            return ((ICollection<KeyValuePair<string, object>>)InnerDictionary).Contains(item);
         }
 
         /// <inheritdoc />
@@ -220,21 +216,21 @@ namespace Microsoft.AspNet.Routing
                 throw new ArgumentNullException(nameof(key));
             }
 
-            if (_dictionary != null)
+            if (InnerDictionary != null)
             {
-                return _dictionary.ContainsKey(key);
+                return InnerDictionary.ContainsKey(key);
             }
-            else if (_properties != null)
+            else if (Properties != null)
             {
-                for (var i = 0; i < _properties.Length; i++)
+                for (var i = 0; i < Properties.Length; i++)
                 {
-                    var property = _properties[i];
+                    var property = Properties[i];
                     if (Comparer.Equals(property.Name, key))
                     {
                         return true;
                     }
                 }
-                
+
                 return false;
             }
             else
@@ -254,7 +250,7 @@ namespace Microsoft.AspNet.Routing
             }
 
             EnsureWritable();
-            ((ICollection<KeyValuePair<string, object>>)_dictionary).CopyTo(array, arrayIndex);
+            ((ICollection<KeyValuePair<string, object>>)InnerDictionary).CopyTo(array, arrayIndex);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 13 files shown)


rynowak:
The impact of this change is about a 2.1MB reduction in allocations from 3000 requests our benchmark  that runs a simple JSON api and generates a link.

**Before**
image

**After**
image

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -43,43 +43,54 @@ public RouteValueDictionary()
...
-            if (values != null)
+            var otherDictionary = values as RouteValueDictionary;
+            if (otherDictionary != null)
             {
-                var keyValuePairCollection = values as IEnumerable<KeyValuePair<string, object>>;
-                if (keyValuePairCollection != null)
+                if (otherDictionary._dictionary != null)
                 {
-                    foreach (var kvp in keyValuePairCollection)
+                    _dictionary = new Dictionary<string, object>(
```
pranavkm: `_dictionary = new Dictionary<string, object>(other._dictionary, StringComparer.OrdinalIgnoreCase)`.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -43,43 +43,54 @@ public RouteValueDictionary()
...
+
+                    foreach (var kvp in otherDictionary._dictionary)
                     {
-                        Add(kvp.Key, kvp.Value);
+                        _dictionary[kvp.Key] = kvp.Value;
                     }
+
+                    return;
+                }
+                else if (otherDictionary._properties != null)
```
pranavkm: Consider making these properties instead. It'd make it slightly easier to read.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -115,46 +127,26 @@ public RouteValueDictionary(object values)
...
-            }
-        }
-
-        /// <inheritdoc />
-        bool ICollection<KeyValuePair<string, object>>.IsReadOnly
-        {
-            get
-            {
-                return ((ICollection<KeyValuePair<string, object>>)_dictionary).IsReadOnly;
+                return _dictionary?.Count ?? _properties?.Length ?? 0;
```
pranavkm: :+1:

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -289,7 +302,126 @@ public bool TryGetValue(string key, out object value)
...
+        }
+
+        public struct Enumerator : IEnumerator<KeyValuePair<string, object>>
+        {
+            private readonly RouteValueDictionary _dictionary;
+
+            private int _index;
+            private Dictionary<string, object>.Enumerator _enumerator;
+
+            public Enumerator(RouteValueDictionary dictionary)
```
pranavkm: `routeValueDictionary`

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -289,7 +302,126 @@ public bool TryGetValue(string key, out object value)
...
+            {
+                if (dictionary == null)
+                {
+                    throw new InvalidOperationException();
```
pranavkm: ?

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -243,30 +253,32 @@ public bool ContainsKey(string key)
...
+            EnsureWritable();
             ((ICollection<KeyValuePair<string, object>>)_dictionary).CopyTo(array, arrayIndex);
         }
 
         /// <inheritdoc />
-        public Dictionary<string, object>.Enumerator GetEnumerator()
+        public Enumerator GetEnumerator()
         {
-            return _dictionary.GetEnumerator();
+            return new Enumerator(this);
```
pranavkm: Would it be simpler if you returned the dictionary enumerator here and used the custom enumerator if `properties` was non null?

pranavkm: Ahh, I guess it would box it. Never mind.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -289,7 +302,126 @@ public bool TryGetValue(string key, out object value)
...
+            public bool MoveNext()
+            {
+                if (_dictionary?._dictionary != null)
+                {
+                    if (_enumerator.MoveNext())
+                    {
+                        Current = _enumerator.Current;
+                        return true;
+                    }
+                    else
```
pranavkm: Could this be rewritten without the `else`s here and making the block inside the default fall through case?

````if (_diction?._dictionary != null)
{

}
else if (_dictionary?._properties != null)
{

}

Current = default(KeyValuePair<string, object>);
return false;
```

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -289,7 +302,126 @@ public bool TryGetValue(string key, out object value)
...
+                    {
+                        Current = default(KeyValuePair<string, object>);
+                        return false;
+                    }
+                }
+                else if (_dictionary?._properties != null)
+                {
+                    if (++_index < _dictionary._properties.Length)
```
pranavkm: Could this be rewritten to not look like Louis code?

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -289,7 +302,126 @@ public bool TryGetValue(string key, out object value)
...
+
+        private void EnsureWritable()
+        {
+            if (_dictionary == null && _properties == null)
+            {
+                _dictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+            }
+            else if (_dictionary == null)
+            {
+                _dictionary = new Dictionary<string, object>(_properties.Length + 1, StringComparer.OrdinalIgnoreCase);
```
pranavkm: Could you add a comment stating why it does a `+ 1`?

pranavkm: That said there are a bunch of scenarios where you wouldn't need the extra slot. `Clear` \ `Remove`, `Keys`, `Values` \ `ContainsKey` etc. Would it help to make it parameterized

src/Microsoft.AspNet.Routing/VirtualPathContext.cs
```diff
@@ -30,12 +32,12 @@ public class VirtualPathContext
 
         public IDictionary<string, object> ProvidedValues { get; set; }
 
-        public IDictionary<string, object> AmbientValues { get; private set; }
+        public RouteValueDictionary AmbientValues { get; private set; }
```
pranavkm: Can the setters be dropped?

pranavkm:
:watch:

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: No. That allocates a boxed enumerator.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: how many points. Actually wait, I'll convert this to a `=>`

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: yup.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: Yeah, I think I'll just drop the +1. I didn't find any examples of a dictionary being resized during testing. 

Dictionary doesn't use your capacity verbatim anyway.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: meh

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: Yeah that's not super-correct.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: Sure

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: Lou says he didn't write this.

src/Microsoft.AspNet.Routing/VirtualPathContext.cs
rynowak: Totes.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: Sure

rynowak:
updated

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -57,23 +56,23 @@ public VirtualPathData(IRouter router, string virtualPath)
...
-
-            _dataTokens = new RouteValueDictionary();
-            if (dataTokens != null)
-            {
-                foreach (var dataToken in dataTokens)
-                {
-                    _dataTokens.Add(dataToken.Key, dataToken.Value);
-                }
-            }
+            _dataTokens = dataTokens == null ? null : new RouteValueDictionary(dataTokens);
```
rynowak: Had to unbreak a test

test/Microsoft.AspNet.Routing.Tests/RouteValueDictionaryTests.cs
```diff
@@ -165,7 +165,11 @@ public void CreateFromObject_MixedCaseThrows()
 
             // Act & Assert
             ExceptionAssert.Throws<ArgumentException>(
-                () => new RouteValueDictionary(obj),
+                () =>
+                {
+                    var dictionary = new RouteValueDictionary(obj);
+                    dictionary.Add("Hi", "There");
+                },
```
rynowak: We're ok with the changes to this behavior. It won't throw on creation now.

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -43,46 +39,63 @@ public RouteValueDictionary()
...
-            if (values != null)
+            var otherDictionary = values as RouteValueDictionary;
+            if (otherDictionary != null)
             {
-                var keyValuePairCollection = values as IEnumerable<KeyValuePair<string, object>>;
-                if (keyValuePairCollection != null)
+                if (otherDictionary.InnerDictionary != null)
                 {
-                    foreach (var kvp in keyValuePairCollection)
+                    InnerDictionary = new Dictionary<string, object>(
```
pranavkm: Could you add a note to this saying this was done to prevent boxing the Dictionary enumerator?

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -43,46 +39,63 @@ public RouteValueDictionary()
...
+            if (values != null)
+            {
+                Properties = PropertyHelper.GetVisibleProperties(values);
+                Value = values;
+
+                return;
             }
         }
 
+        private Dictionary<string, object> InnerDictionary { get; set; }
```
pranavkm: :clap:

rynowak: I don't see how it's better, but I'll take the PranavPoints

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
```diff
@@ -43,46 +39,63 @@ public RouteValueDictionary()
         /// Only public instance non-index properties are considered.
         /// </remarks>
         public RouteValueDictionary(object values)
-            : this()
```
pranavkm: Do the summary or remarks need to be looked at?

pranavkm:
:shipit:

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: Sure

src/Microsoft.AspNet.Routing/RouteValueDictionary.cs
rynowak: No semantics of this have changed, we just made it better.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
Eilon: `??`?

src/Microsoft.AspNet.Routing/VirtualPathData.cs
```diff
@@ -57,23 +56,23 @@ public VirtualPathData(IRouter router, string virtualPath)
...
         /// <summary>
         /// Gets the collection of custom values for the <see cref="Router"/>.
         /// </summary>
-        public IDictionary<string, object> DataTokens
+        public RouteValueDictionary DataTokens
         {
-            get { return _dataTokens; }
+            get
+            {
+                if (_dataTokens == null)
```
Eilon: Given the codes above, will `_dataTokens` ever be `null`?

src/Microsoft.AspNet.Routing/VirtualPathData.cs
rynowak: I want the opposite of this. Leave it null if possible.

src/Microsoft.AspNet.Routing/VirtualPathData.cs
rynowak: Yes, that's the point of the `? :`

rynowak: Also, it's a total :trollface: to wait until 15 seconds after I checked this in to comment

src/Microsoft.AspNet.Routing/VirtualPathData.cs
Eilon: Oh duh I read it backwards...

rynowak: The common case here is that there aren't any data tokens and that no one will look at them, so we're alloc-free when that happens. If you are using data tokens then pay-for-play

src/Microsoft.AspNet.Routing/VirtualPathData.cs
Eilon: IE was having issues and by the time I got the PR back up, you had merged. :smile:

src/Microsoft.AspNet.Routing/VirtualPathData.cs
Eilon: Yup, totally makes sense.

PR #249, opened 2015-12-10
Add IRouteHandler, RouteBase, and IRoutingFeature

rynowak:
Adds IRouterHandler, an abstraction for endpoints in the routing system
that can't chain (example: delegates). The idea is that some kinds of
routes aren't really friendly to chaining. If you don't support chaining,
then accept IRouteHandler and work with that rather than IRouter.

There's one implementation of IRouteHandler, RouteHandler. It implements
both IRouter and IRouteHandler.

Adds RouteBase as a base class for routes based on our template syntax and
defaults/constraints/data-tokens. Updated a lot of signatures to be
get/set virtual and mutable to facilitate or bigger variety of usage
scenarios.

Renamed TemplateRoute to just Route, now inherits from RouteBase.

Adds IRoutingFeature for middleware scenarios where you don't have access
to the route context.

Also adds some basic extension methods for accessing route values.


src/Microsoft.AspNet.Routing/Internal/TaskCache.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Routing.Internal
+{
+    public static class TaskCache
+    {
+        public static readonly Task CompletedTask = Task.FromResult(0);
```
pranavkm: Isn't there a Task.Completed in CoreCLR?

rynowak: Yeah. I can put in an ifdef here.

src/Microsoft.AspNet.Routing.Abstractions/IRouteHandler.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.Routing
+{
+    public interface IRouteHandler
```
pranavkm: Docs

src/Microsoft.AspNet.Routing/Route.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            _target = target;
+        }
+
+        public string RouteTemplate => ParsedTemplate.TemplateText;
+
+        protected override async Task OnRouteMatched(RouteContext context)
+        {
+            context.RouteData.Routers.Add(_target);
+            await _target.RouteAsync(context);
```
pranavkm: `return _target.RouteAsync(context);`

src/Microsoft.AspNet.Routing/Tree/TreeRouteLinkGenerationEntry.cs
```diff
@@ -20,12 +20,12 @@ public class TreeRouteLinkGenerationEntry
         /// <summary>
         /// The route constraints.
```
pranavkm: Gets or sets

test/Microsoft.AspNet.Routing.Tests/RouteTest.cs
```diff
@@ -730,25 +472,12 @@ public void CreateTemplate_InlineConstraint_Regex_Malformed()
...
-            var context = CreateRouteContext("/Home");
-
-            // Act
-            await route.RouteAsync(context);
+            Assert.Null(context.Handler);
 
-            // Assert
-            Assert.False(context.IsHandled);
-            Assert.Empty(context.RouteData.Routers);
+            Assert.Single(context.RouteData.Values);
```
pranavkm:
```C#
var value = Assert.Single(context.RouteData.Values);
Assert.Equal("controller", value.Key);
Assert.Equal("Home", Assert.IsType<string>(value.Value));
```

pranavkm:
Can you add docs to all the new API in abstractions?

:shipit: once that is addressed.

PR #255, opened 2015-12-14
HttpMethodConstraint -> HttpMethodRouteConstraint

rynowak:
This is consistent with other constraint types in routing, and avoids a
naming conflict with MVC.

This is a change **away** from the names used in System.Web and
System.Web.Http.Routing, but it seems worth doing for consistency and
clarity.


src/Microsoft.AspNet.Routing/Constraints/HttpMethodRouteConstraint.cs
```diff
@@ -11,14 +11,14 @@ namespace Microsoft.AspNet.Routing.Constraints
     /// <summary>
     /// Constrains the HTTP method of request or a route.
     /// </summary>
-    public class HttpMethodConstraint : IRouteConstraint
```
rynowak: Note that the files already had the *right* name :laughing:

Eilon:
:shipit:

PR #258, opened 2015-12-15
Remove custom url encoding

### LINKED ISSUE #214: Get rid of custom URI encoding code
https://github.com/aspnet/Routing/blob/c6941e797f1585730b69d088a3b1b0c17ce6afa7/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs#L281

Routing isn't using our standard encoders.

pranavkm:
Fixes #214

### DIFF
--- a/src/Microsoft.AspNet.Routing/RouteBase.cs
+++ b/src/Microsoft.AspNet.Routing/RouteBase.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Routing.Internal;
@@ -115,7 +116,7 @@ namespace Microsoft.AspNet.Routing
         /// <inheritdoc />
         public virtual VirtualPathData GetVirtualPath(VirtualPathContext context)
         {
-            EnsureBinder();
+            EnsureBinder(context.HttpContext);
             EnsureLoggers(context.HttpContext);
 
             var values = _binder.GetValues(context.AmbientValues, context.Values);
@@ -237,11 +238,12 @@ namespace Microsoft.AspNet.Routing
             }
         }
 
-        private void EnsureBinder()
+        private void EnsureBinder(HttpContext context)
         {
             if (_binder == null)
             {
-                _binder = new TemplateBinder(ParsedTemplate, Defaults);
+                var urlEncoder = context.RequestServices.GetRequiredService<UrlEncoder>();
+                _binder = new TemplateBinder(ParsedTemplate, urlEncoder, Defaults);
             }
         }
 
--- a/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
@@ -414,20 +414,22 @@ namespace Microsoft.AspNet.Routing.Template
             private readonly StringBuilder _uri;
 
             // Holds the 'optional' parts of the uri. We need a secondary buffer to handle cases where an optional
-            // segment is in the middle of the uri. We don't know whether or not we need to write it out - if it's
+            // segment is in the middle of the uri. We don't know if we need to write it out - if it's
             // followed by other optional segments than we will just throw it away.
-            private readonly StringBuilder _buffer;
+            private readonly List<BufferValue> _buffer;
             private readonly UrlEncoder _urlEncoder;
             private readonly StringWriter _uriWriter;
 
             private bool _hasEmptySegment;
+            private int _lastValueOffset;
 
             public UriBuildingContext(UrlEncoder urlEncoder)
             {
                 _urlEncoder = urlEncoder;
                 _uri = new StringBuilder();
-                _buffer = new StringBuilder();
+                _buffer = new List<BufferValue>();
                 _uriWriter = new StringWriter(_uri);
+                _lastValueOffset = -1;
 
                 BufferState = SegmentState.Beginning;
                 UriState = SegmentState.Beginning;
@@ -458,7 +460,17 @@ namespace Microsoft.AspNet.Routing.Template
                     return false;
                 }
 
-                _uri.Append(_buffer);
+                for (var i = 0; i < _buffer.Count; i++)
+                {
+                    if (_buffer[i].RequiresEncoding)
+                    {
+                        _urlEncoder.Encode(_uriWriter, _buffer[i].Value);
+                    }
+                    else
+                    {
+                        _uri.Append(_buffer[i].Value);
+                    }
+                }
                 _buffer.Clear();
 
                 if (UriState == SegmentState.Beginning && BufferState == SegmentState.Beginning)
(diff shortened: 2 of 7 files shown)


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -471,13 +472,24 @@ public bool Accept(string value)
                 BufferState = SegmentState.Inside;
                 UriState = SegmentState.Inside;
 
-                _uri.Append(value);
+                // Allow the first segment to have a leading slash.
+                // This prevents the leading slash from PathString segments from being encoded.
+                if (_uri.Length == 0 && value.Length > 0 && value[0] == '/')
```
pranavkm: The hack I pinged you about.

src/Microsoft.AspNet.Routing/RouteBase.cs
```diff
@@ -237,11 +238,12 @@ public virtual VirtualPathData GetVirtualPath(VirtualPathContext context)
...
         }
 
-        private void EnsureBinder()
+        private void EnsureBinder(HttpContext context)
         {
             if (_binder == null)
             {
-                _binder = new TemplateBinder(ParsedTemplate, Defaults);
+                var urlEncoder = context.RequestServices.GetRequiredService<UrlEncoder>();
+                _binder = new TemplateBinder(ParsedTemplate, urlEncoder, Defaults);
```
rynowak: ugh. this is really ugly to have to have a service to build a URL

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
rynowak: Yeah, I see. I'm not super worried about this check, really you could just **never** encode `/` without the check. We've had a long-standing bug about not encoding `/` inside of segments, and we've decided multiple times not to address it.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -471,13 +472,24 @@ public bool Accept(string value)
...
+                    _urlEncoder.Encode(_uriWriter, value);
+                }
+
                 return true;
             }
 
             public void Remove(string literal)
             {
-                _uri.Length -= literal.Length;
+                _uri.Length -= _urlEncoder.Encode(literal).Length;
```
rynowak: I'm worried about all of these allocations from encoding things. We have a standing work item to optimize this, so if the encoding part doesn't get better, it's throw away work.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateBinderTests.cs
```diff
@@ -14,116 +16,97 @@ public class TemplateBinderTests
     {
         private readonly IInlineConstraintResolver _inlineConstraintResolver = GetInlineConstraintResolver();
 
-        public static IEnumerable<object[]> EmptyAndNullDefaultValues
-        {
-            get
+        public static TheoryData EmptyAndNullDefaultValues =>
+            new TheoryData<string, RouteValueDictionary, RouteValueDictionary, string>
```
rynowak: This is worse than what we had. revert

rynowak:
:watch:

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
pranavkm: I think I saw a test where they expect it to be encoded.

test/Microsoft.AspNet.Routing.Tests/Template/TemplateBinderTests.cs
pranavkm: Huh? It's much nicer!

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
pranavkm: I think we're almost always removing the last segment we accepted. We should be able to keep track of how long the encoded length was.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
rynowak: Ignore me then. I'm the WORST.

src/Microsoft.AspNet.Routing/RouteBase.cs
pranavkm: Are you complaining about the `TemplateBinder` type or the need to `GetRequiredService`?

rynowak: The fact that it shows up. The `GetRequiredService` is fine.

pranavkm:
:up: :date:

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -557,5 +584,18 @@ private enum SegmentState
...
+        {
+            public BufferValue(string value, bool requiresEncoding)
+            {
+                Value = value;
+                RequiresEncoding = requiresEncoding;
+            }
+
+            public bool RequiresEncoding { get; }
+
+            public string Value { get; }
```
rynowak: snazzy

test/Microsoft.AspNet.Routing.Tests/Template/TemplateBinderTests.cs
rynowak: This is literally the worst

rynowak:
:shipit:

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -534,15 +561,15 @@ internal void EndSegment()
...
             {
                 // We can ignore any currently buffered segments - they are are guaranteed to be 'defaults'.
-                return _uri.ToString();
+                return _uri;
             }
 
             private string DebuggerToString()
             {
-                return string.Format("{{Accepted: '{0}' Buffered: '{1}'}}", _uri.ToString(), _buffer.ToString());
+                return string.Format("{{Accepted: '{0}' Buffered: '{1}'}}", _uri, string.Join("", _buffer));
```
Eilon: Cool people use `string.Empty`

test/Microsoft.AspNet.Routing.Tests/Template/TemplateBinderTests.cs
dougbu: Gotta go w/ @rynowak on this one. Let's not churn the code to use C# 6, especially when the result is so ugly. Also makes it hard to see if anything actually needed to change here.

dougbu: .. BTW I do agree `IEnumerable<object[]>` -> `TheoryData<string, RouteValueDictionary, RouteValueDictionary, string>` is a real improvement. Just lose the personal preference syntax.

pranavkm: Ehh.

src/Microsoft.AspNet.Routing/RouteBase.cs
pranavkm: Does Routing need to add this service?

rynowak: it can't hurt... doesn't hosting add it?

pranavkm: @ryanbrandenburg had to add it here - https://github.com/aspnet/Mvc/commit/70f9431e7fd500583a84821cbcae4a97f64f78b2. Not sure why it would be missing in that case. I'll go ahead and do a TryAdd since we require it. It'd make it more reliable.

PR #259, opened 2015-12-15
Removed CatchAll and Fallback from DecisionTree

ajaybhargavb:
Related issue - https://github.com/aspnet/Mvc/issues/3430

@rynowak


rynowak:
:shipit:

PR #264, opened 2015-12-18
Reduce allocation in URL generation

rynowak:
This change optimizes our a per-operation dictionary that really can just
be cached for the whole app's lifetime.

### DIFF
--- a/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
@@ -14,14 +14,15 @@ namespace Microsoft.AspNet.Routing.Template
 {
     public class TemplateBinder
     {
-        private readonly IReadOnlyDictionary<string, object> _defaults;
+        private readonly RouteValueDictionary _defaults;
+        private readonly RouteValueDictionary _filters;
         private readonly RouteTemplate _template;
         private readonly UrlEncoder _urlEncoder;
 
         public TemplateBinder(
             RouteTemplate template,
             UrlEncoder urlEncoder,
-            IReadOnlyDictionary<string, object> defaults)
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -132,25 +141,22 @@ public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteV
 
             // Any default values that don't appear as parameters are treated like filters. Any new values
```
pranavkm: You could remove this comment.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -14,14 +14,15 @@ namespace Microsoft.AspNet.Routing.Template
...
-        private readonly IReadOnlyDictionary<string, object> _defaults;
+        private readonly RouteValueDictionary _defaults;
+        private readonly RouteValueDictionary _filters;
         private readonly RouteTemplate _template;
         private readonly UrlEncoder _urlEncoder;
 
         public TemplateBinder(
             RouteTemplate template,
             UrlEncoder urlEncoder,
-            IReadOnlyDictionary<string, object> defaults)
```
rynowak: Declaring this as `IReadOnlyDictionary<>` was also resulting in a boxed enumerator per operation.

pranavkm:
:shipit:

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
rynowak: Yeah, I still don't really see a reason to. No one understands this feature anyway.

PR #266, opened 2015-12-18
Pool state used to generate URL

rynowak:
/cc @pranavkm - working on numbers now

### DIFF
--- a/src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
+++ b/src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
@@ -1,0 +1,199 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Text.Encodings.Web;
+
+namespace Microsoft.AspNet.Routing.Internal
+{
+    [DebuggerDisplay("{DebuggerToString(),nq}")]
+    public class UriBuildingContext
+    {
+        // Holds the 'accepted' parts of the uri.
+        private readonly StringBuilder _uri;
+
+        // Holds the 'optional' parts of the uri. We need a secondary buffer to handle cases where an optional
+        // segment is in the middle of the uri. We don't know if we need to write it out - if it's
+        // followed by other optional segments than we will just throw it away.
+        private readonly List<BufferValue> _buffer;
+        private readonly UrlEncoder _urlEncoder;
+
+        private bool _hasEmptySegment;
+        private int _lastValueOffset;
+
+        public UriBuildingContext(UrlEncoder urlEncoder)
+        {
+            _urlEncoder = urlEncoder;
+            _uri = new StringBuilder();
+            _buffer = new List<BufferValue>();
+            Writer = new StringWriter(_uri);
+            _lastValueOffset = -1;
+
+            BufferState = SegmentState.Beginning;
+            UriState = SegmentState.Beginning;
+        }
+
+        public SegmentState BufferState { get; private set; }
+
+        public SegmentState UriState { get; private set; }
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
```diff
@@ -0,0 +1,199 @@
...
+            if (_buffer.Capacity == 8)
+            {
+                _buffer.Capacity = 8;
+            }
+
+            _hasEmptySegment = false;
+            _lastValueOffset = -1;
+            BufferState = SegmentState.Beginning;
+            UriState = SegmentState.Beginning;
+        }
```
rynowak: This is the only new code here, besides exposing the writer as a property.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -283,12 +296,22 @@ public string BindValues(RouteValueDictionary acceptedValues)
...
+                    wroteFirst = true;
+                }
+                else
+                {
+                    context.Writer.Write('&');
+                }
+
+                _urlEncoder.Encode(context.Writer, kvp.Key);
+                context.Writer.Write('=');
+                _urlEncoder.Encode(context.Writer, converted);
```
rynowak: Replaced usage of the `QueryBuilder`. We can encode directly to the writer, and we can more easily pool this way.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -350,7 +373,7 @@ private static bool IsRoutePartNonEmpty(object routePart)
         }
 
         [DebuggerDisplay("{DebuggerToString(),nq}")]
-        private class TemplateBindingContext
+        private struct TemplateBindingContext
```
rynowak: This doesn't have any important state, all the state is part of `_acceptedValues`

src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        public override string ToString()
```
pranavkm: Could we remove this since it's not intended to be used?

src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            _uri.Clear();
+            if (_uri.Capacity > 128)
+            {
+                // We don't want to retain too much memory if this is getting pooled.
+                _uri.Capacity = 128;
+            }
+
+            _buffer.Clear();
+            if (_buffer.Capacity == 8)
```
pranavkm: `> 8`

src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
pranavkm: Nevermind. We still use this.

pranavkm:
Changes look good. How are the allocation profiles?

src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
rynowak: :trollface:

src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
rynowak: doh.

While I'm at it, I'll add `var patience = 5787`

rynowak:
Hmm. I'll have to log on to my main machine tomorrow and find the data. I did some profiling after submitting this. I think it was worth about 1mb on MediumAPI

rynowak:
**Before**
image


**After**
image

pranavkm:
:shipit:

PR #269, opened 2015-12-30
Add support IEnumerable parameters for query string

grehovm:
Recreated a pull request
https://github.com/aspnet/Routing/pull/234

### DIFF
--- a/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
 using System.Diagnostics;
 using System.Globalization;
 using System.Text.Encodings.Web;
@@ -290,28 +291,34 @@ namespace Microsoft.AspNet.Routing.Template
                     continue;
                 }
 
-                var converted = Convert.ToString(kvp.Value, CultureInfo.InvariantCulture);
-                if (string.IsNullOrEmpty(converted))
+                var values = kvp.Value as IEnumerable;
+                if (values != null && !(values is string))
                 {
-                    continue;
-                }
-
-                if (!wroteFirst)
-                {
-                    context.Writer.Write('?');
-                    wroteFirst = true;
+                    foreach (var value in values)
+                    {
+                        wroteFirst |= AddParameterToContext(context, kvp.Key, value, wroteFirst);
+                    }
                 }
                 else
                 {
-                    context.Writer.Write('&');
+                    wroteFirst |= AddParameterToContext(context, kvp.Key, kvp.Value, wroteFirst);
                 }
+            }
+            return context.ToString();
+        }
 
-                _urlEncoder.Encode(context.Writer, kvp.Key);
+        private bool AddParameterToContext(UriBuildingContext context, string key, object value, bool wroteFirst)
+        {
+            var converted = Convert.ToString(value, CultureInfo.InvariantCulture);
+            if (!string.IsNullOrEmpty(converted))
+            {
+                context.Writer.Write(wroteFirst ? '&' : '?');
+                _urlEncoder.Encode(context.Writer, key);
                 context.Writer.Write('=');
                 _urlEncoder.Encode(context.Writer, converted);
+                return true;
             }
-
-            return context.ToString();
+            return false;
         }
 
         private TemplatePart GetParameter(string name)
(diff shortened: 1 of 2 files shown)


grehovm:
It seems, all conditions are satisfied. please accept pull request while nothing has changed.
@Eilon

test/Microsoft.AspNet.Routing.Tests/RouteTest.cs
```diff
@@ -631,6 +631,60 @@ public void GetVirtualPath_Fail()
...
+            Assert.Empty(pathData.DataTokens);
+        }
+        
+        [Fact]
+        public void GetVirtualPath_ForList_StringWorkaround()
+        {
+            // Arrange
+            var route = CreateRoute("{controller}/{action}");
+            var context = CreateVirtualPathContext(
+                new { page = 1, color = new List<string> { "red", "green", "blue" }, message = "bum!" },
```
pranavkm: Could you pick a text other than `bum!`? Our CI runs a tool on sources which looks for questionable content and this looks like something it would flag.

src/Microsoft.AspNet.Routing/Template/TemplateBinder.cs
```diff
@@ -7,6 +7,7 @@
 using System.Text.Encodings.Web;
 using Microsoft.AspNet.Routing.Internal;
 using Microsoft.Extensions.ObjectPool;
+using System.Collections;
```
pranavkm: Sort usings (`System` namespaces go first)

pranavkm:
I'll merge this in once this is ready to go.

grehovm:
I think I found a bug. I changed text in text with a space and the space has been converted twice. I pumped my main repository without my edits, and this was confirmed.

`Expected: /Home/Index%3Fpage%3D1%26message%3Dtext%20message`
`Actual:      /Home/Index%3Fpage%3D1%26message%3Dtext%2520message`
Can I do a test without a space and we will close pull request?

test/Microsoft.AspNet.Routing.Tests/RouteTest.cs
```diff
@@ -631,6 +631,60 @@ public void GetVirtualPath_Fail()
             Assert.Null(path);
         }
 
+        [Fact]
+        public void GetVirtualPath_ForList()
```
Eilon: Could you change this test to be `GetVirtualPath_ForListOfStrings` and add one more test for `GetVirtualPath_ForListOfInts` to show that string conversion happens for non-strings?

Eilon:
@grehovm where are you actually seeing this:

```
 Expected: /Home/Index%3Fpage%3D1%26message%3Dtext%20message 
 Actual: /Home/Index%3Fpage%3D1%26message%3Dtext%2520message 
```

Is that in one of these unit tests in this repo? Or from some other repo?

The reason I ask is that nothing in those strings should be URL encoded except for the space itself. And indeed none of the unit tests I see here have anything encoded...

grehovm:
This test fail in my repo
```C#
[Fact]
public void GetVirtualPath_ForList_StringWorkaround()
{
    // Arrange
    var route = CreateRoute("{controller}/{action}");
    var context = CreateVirtualPathContext(
        new { page = 1, color = new List<string> { "red", "green", "blue" }, message = "text message" },
        new { controller = "Home", action = "Index" });

    // Act
    var pathData = route.GetVirtualPath(context);

    // Assert
    Assert.Equal(new PathString("/Home/Index?page=1&color=red&color=green&color=blue&message=text message"), pathData.VirtualPath);
    Assert.Same(route, pathData.Router);
    Assert.Empty(pathData.DataTokens);
}
```
Then i clone https://github.com/aspnet/Routing.git. Add only this test. And got the same results.
Space in `text message` has been converted twice.

Eilon:
@pranavkm can you investigate this? This seems quite odd...

grehovm:
Shall i do something else?

pranavkm:
@grehovm let me have a look. I think your PR looks good as is for now.

Eilon:
@grehovm indeed the code looks great, and thanks for adding the tests!

pranavkm:
We represent generated virtual paths using `PathString` and `PathString.ToString` UrlEncodes segments. It almost looks like the type isn't designed to hold query string values. Your changes look fine, but yeah the double encoding looks suspicious. 

@Tratcher Should we change this data type - https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNet.Routing.Abstractions/VirtualPathData.cs#L86 - to not be a `PathString` if it can contain query components or is the `ToString` behavior in `PathString` incorrect?

cc @rynowak 

@grehovm I'll merge your changes in. Thanks for noticing the issue with the double encoding.

pranavkm:
Merged in https://github.com/aspnet/Routing/commit/a61a68defcce061191f2bbe5d4a01807908ebc26

Tratcher:
@pranavkm PathString is not designed to contain a query, it's for un-escaped path segments.

pranavkm:
Ok. I'll file a bug to change the data type.

PR #272, opened 2016-01-01
Optimize TemplateMatcher

rynowak:
Replaces a bunch of dictionary operations with indexing into an array by
doing some caching. Also eliminating an enumerator allocation by changing
from IReadOnlyDictionary to RouteValueDictionary.


rynowak:
Here's the data that I'm working with:

image


This should eliminate most of the 'ContainsKey` and `TryGetValue`

Eilon:
:shipit:

PR #273, opened 2016-01-04
Change the data type of VirtualPathData.VirtualPath

### LINKED ISSUE #270: Change the data type of VirtualPathData.VirtualPath
Based on discussions in https://github.com/aspnet/Routing/pull/269#issuecomment-168130032:
we store Url encoded path and query components in this field. However the type is designed to not contain non encoded path component. 

Alternatives to represent it:
* Vanilla string
* Two separate components `PathString` + `QueryString`
* Uri (think it might get treated as a relative path)

pranavkm:
Fixes #270

### DIFF
--- a/src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
+++ b/src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
@@ -188,6 +188,12 @@ namespace Microsoft.AspNet.Routing.Internal
         public override string ToString()
         {
             // We can ignore any currently buffered segments - they are are guaranteed to be 'defaults'.
+            if (_uri.Length > 0 && _uri[0] != '/')
+            {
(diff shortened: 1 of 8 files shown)


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Routing/Internal/UriBuildingContext.cs
```diff
@@ -188,6 +188,12 @@ public void Clear()
         public override string ToString()
         {
             // We can ignore any currently buffered segments - they are are guaranteed to be 'defaults'.
+            if (_uri.Length > 0 && _uri[0] != '/')
+            {
+                // Normalize generated paths so that they always contain a leading slash.
+                _uri.Insert(0, '/');
+            }
```
rynowak: :heart_eyes_cat:

rynowak:
:shipit:

Issue #276, opened 2016-01-11
Throw a nicer error when the routing services haven't been added to DI


DamianEdwards:
Like [other parts of the framework](https://github.com/aspnet/SignalR-Server/blob/da96c1d9abae49e6aadeefd30942ad0269a65996/src/Microsoft.AspNet.SignalR.Server/Hosting/BuilderExtensions.cs#L93) we should throw a nice helper exception when services are missing from DI, including the sample call the app needs to add the routing services to DI, e.g.
```
Routing services have not been registered. Register services by calling 'IServices.UseRouting()'.
```

DamianEdwards:
@rynowak @Eilon

Eilon:
MVC already does this. We can follow the same model.

rynowak:
Can we not follow the same model instead? It would be nice to do this on Startup instead of on first request

Eilon:
Hmm ok, do we have an idea of how to do that?

rynowak:
We can register a marker service and have the routing middleware resolve it when created.

DamianEdwards:
Ya, the code linked to above does this.

rynowak:
MVC does it on first request - we should think about changing it to do it when we add AddMvc.

Eilon:
@javiercn you got that? Let's add this to Routing, and then also change MVC to do the same. I logged https://github.com/aspnet/Mvc/issues/3907 and assigned to you.

javiercn:
@Eilon will do it for routing and mvc

PR #278, opened 2016-01-13
Fixes [#276] Throw a nicer error when the routing services haven't been added to DI

javiercn:


test/Microsoft.AspNet.Routing.Tests/BuilderExtensionsTest.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Builder;
+using Moq;
+using Xunit;
+
+namespace Microsoft.AspNet.Routing.Tests
```
dougbu: `Microsoft.AspNet.Builder` to match `BuilderExtensions` (plus: almost never `...Test` or `...Tests`)

test/Microsoft.AspNet.Routing.Tests/BuilderExtensionsTest.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            // Arrange
+            var applicationBuilderMock = new Mock<IApplicationBuilder>();
+            applicationBuilderMock
+                .Setup(s => s.ApplicationServices)
+                .Returns(Mock.Of<IServiceProvider>());
+
+            var router = Mock.Of<IRouter>();
+            
+            // Act
```
dougbu: `// Act & Assert` and get rid of extra `// Assert` line below. Same for other tests.

test/Microsoft.AspNet.Routing.Tests/BuilderExtensionsTest.cs
```diff
@@ -0,0 +1,36 @@
...
+            var error = Assert.Throws<InvalidOperationException>(
```
dougbu: We usually use `ex` or `exception` for similar variables.

dougbu:
:watch:

javiercn:
@dougbu updated

src/Microsoft.AspNet.Routing/RouteBuilder.cs
```diff
@@ -21,6 +23,14 @@ public RouteBuilder(IApplicationBuilder applicationBuilder, IRouter defaultHandl
                 throw new ArgumentNullException(nameof(applicationBuilder));
             }
 
+            if (applicationBuilder.ApplicationServices.GetService(typeof(RoutingMarkerService)) == null)
+            {
+                throw new InvalidOperationException(Resources.FormatUnableToFindServices(
```
dougbu: Forgot to ask last time: Do we really want the constructor to `throw` for a situation unrelated to its parameters? Realize `RouteBuilder` objects aren't passed into the system but am still hoping there's another place where we can cover the non-`UseRouting()` scenario.

test/Microsoft.AspNet.Routing.Tests/RouteBuilderTest.cs
```diff
@@ -0,0 +1,33 @@
...
+        [Fact]
+        public void Ctor_ThrowsInvalidOperationException_IfRoutingMarkerServiceIsNotRegistered()
+        {
+            // Arrange
+            var applicationBuilderMock = new Mock<IApplicationBuilder>();
+            applicationBuilderMock
+                .Setup(s => s.ApplicationServices)
+                .Returns(Mock.Of<IServiceProvider>());
+
+            // Act
```
dougbu: Missed an `// Act & Assert`

test/Microsoft.AspNet.Routing.Tests/RouteBuilderTest.cs
```diff
@@ -0,0 +1,33 @@
...
+            var error = Assert.Throws<InvalidOperationException>(() => new RouteBuilder(applicationBuilderMock.Object));
```
dougbu: `error` -> `exception` as in other new test

dougbu:
:watch: for small touch-ups and one [open question](https://github.com/aspnet/Routing/pull/278#discussion-diff-49766388).

javiercn:
@dougbu updated. @rynowak any comment?

src/Microsoft.AspNet.Routing/RouteBuilder.cs
dougbu: Caught @Eilon and he's :ok: with this. @rynowak has more to say however.

dougbu:
:shipit: but definitely get input from @rynowak.

src/Microsoft.AspNet.Routing/RouteBuilder.cs
rynowak: Hmm - what's the reason for this?  Like, what's the callstack that needs to throw that isn't throwing?

javiercn:
@rynowak Updated to include your suggestions

src/Microsoft.AspNet.Routing/BuilderExtensions.cs
```diff
@@ -29,7 +31,51 @@ public static IApplicationBuilder UseRouter(this IApplicationBuilder builder, IR
...
+
+            if (builder.ApplicationServices.GetService(typeof(RoutingMarkerService)) == null)
+            {
+                throw new InvalidOperationException(Resources.FormatUnableToFindServices(
+                    nameof(IServiceCollection),
+                    nameof(RoutingServiceCollectionExtensions.AddRouting),
+                    "ConfigureServices(...)"));
+            }
+
+            return new RouteBuilder(builder, defaultHandler);
```
rynowak: We should also add the middleware to the pipeline here. This is a `Use` method after all.

rynowak: Yeah let's leave this.

rynowak:
:shipit: for https://github.com/aspnet/Routing/commit/db34f85ea1ef69cf980a633db260ce7d889bf662

PR #280, opened 2016-02-02
Reduce defensive copies

rynowak:
Please ignore the changes to `project.json`

This change reduces allocations that happen on every routed request by eliminating defensive copies.

**Before**
image

**After**
image

Net improvement is about 900kb of allocations, which is pretty good on this workload where we're under 25mb on x86.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
@@ -12,6 +12,7 @@ namespace Microsoft.AspNetCore.Routing
     public class RouteData
     {
         private RouteValueDictionary _dataTokens;
+        private IList<IRouter> _routers;
         private RouteValueDictionary _values;
 
         /// <summary>
@@ -19,8 +20,7 @@ namespace Microsoft.AspNetCore.Routing
         /// </summary>
         public RouteData()
         {
-            // Perf: Avoid allocating DataTokens and RouteValues unless needed.
-            Routers = new List<IRouter>();
+            // Perf: Avoid allocating collections unless needed.
         }
 
         /// <summary>
@@ -34,9 +34,12 @@ namespace Microsoft.AspNetCore.Routing
                 throw new ArgumentNullException(nameof(other));
             }
 
-            Routers = new List<IRouter>(other.Routers);
-            
-            // Perf: Avoid allocating DataTokens and RouteValues unless we need to make a copy.
+            // Perf: Avoid allocating collections unless we need to make a copy.
+            if (other._routers != null)
+            {
+                _routers = new List<IRouter>(other.Routers);
+            }
+
             if (other._dataTokens != null)
             {
                 _dataTokens = new RouteValueDictionary(other._dataTokens);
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Routing/RouteBase.cs
```diff
@@ -19,6 +19,12 @@ public abstract class RouteBase : IRouter, INamedRouter
     {
         private TemplateMatcher _matcher;
         private TemplateBinder _binder;
+        private readonly IReadOnlyDictionary<string, IRouteConstraint> _constraints;
+        private readonly RouteValueDictionary _dataTokens;
+        private readonly RouteValueDictionary _defaults;
+        private readonly IRouter _target;
+        private readonly RouteTemplate _parsedTemplate;
+        private readonly string _routeTemplate;
```
rynowak: Bad merge, will undo.

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -173,31 +173,24 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
...
+                                snapshot.Restore(context.RouteData);
+                                continue;
+                            }
 
-                        context.RouteData = newRouteData;
+                            _logger.MatchedRoute(match.Entry.RouteName, match.Entry.RouteTemplate.TemplateText);
 
-                        try
-                        {
                             await match.Entry.Target.RouteAsync(context);
```
rynowak: Bad merge will fix

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -173,31 +173,24 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
...
-                        _logger.MatchedRoute(match.Entry.RouteName, match.Entry.RouteTemplate.TemplateText);
+                            if (!RouteConstraintMatcher.Match(
+                                    match.Entry.Constraints,
+                                    context.RouteData.Values,
+                                    context.HttpContext,
+                                    this,
+                                    RouteDirection.IncomingRequest,
+                                    _constraintLogger))
+                            {
+                                snapshot.Restore(context.RouteData);
```
rynowak: will remove

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
```diff
@@ -84,5 +85,99 @@ public RouteValueDictionary Values
...
+                _routerCount = routeData.Routers.Count;
+                _values = routeData._values;
+            }
+
+            public void Restore(RouteData routeData)
+            {
+                routeData._dataTokens = _dataTokens;
+                routeData._values = _values;
+
+                var routersToTrim = routeData._routers.Count - _routerCount;
```
Eilon: Is this just being very defensive? In a normal case wouldn't it always just need to remove exactly 1 router?

rynowak: In a normal case that's certainly true, but we need to handle **all** cases.

Eilon: Yup, got it.

Eilon:
:shipit:

PR #283, opened 2016-02-17
Minor fix to service collection extensions

kichalla:
@Eilon Following a pattern here as in other repos

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/DependencyInjection/RoutingServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.Routing/DependencyInjection/RoutingServiceCollectionExtensions.cs
@@ -15,15 +15,17 @@ namespace Microsoft.Extensions.DependencyInjection
     /// </summary>
     public static class RoutingServiceCollectionExtensions
     {
-        public static IServiceCollection AddRouting(this IServiceCollection services)
+        /// <summary>
+        /// Adds services required for routing requests.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Routing/DependencyInjection/RoutingServiceCollectionExtensions.cs
```diff
@@ -17,13 +17,11 @@ public static class RoutingServiceCollectionExtensions
     {
         public static IServiceCollection AddRouting(this IServiceCollection services)
```
pranavkm: Docs?

Eilon: Funny, I just logged a bug on that: https://github.com/aspnet/Routing/issues/284

@kichalla I'll assign to you and it's up to you whether to do in this PR or a new one.

Eilon:
The code looks :shipit: to me but would like @rynowak to confirm this makes sense.

rynowak:
These should return `void` - other than that looks fine

Eilon:
@rynowak ah, is `void`-returning part of the pattern? If so, @kichalla @javiercn please be aware of that as you do the cleanup.

rynowak:
Yep. Returning `void` from `AddXyz` methods specifically to remove chaining.

Some methods like `AddMvc()` return their own builder. We want that to stand out.

kichalla:
:up: :date:

Eilon:
:shipit:

PR #290, opened 2016-02-26
[Perf] Avoid Linq method and save its related allocations in TreeRouter

mnltejaswini:
TreeRouter.GenerateVirtualPath() has the following expression

var parameter = entry.Template.Parameters
             .FirstOrDefault(p => string.Equals(p.Name, kvp.Key, StringComparison.OrdinalIgnoreCase));

that is resulting in 3 allocations
1. Delegate for FirstOrDefault (System.Func)
2. Closure for TreeRouter
3. Enumerator for Parameters 
This expression is used in a loop for each of the Route values. 

By removing this Linq expression, there is 3.5% saving in total number of allocations and 2.5% in total number of bytes allocated. 

The measurements are against Performance/BasicViews test app with the View having 6 action links . Each of these actions href is generated using TagHelpers with **attribute routing**. And loadtest is run against the app with the following setting

loadtest --rps 100 -n 3000 -k <URL>

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
+++ b/src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
@@ -59,5 +59,24 @@ namespace Microsoft.AspNetCore.Routing.Template
         {
             return string.Join(SeparatorString, Segments.Select(s => s.DebuggerToString()));
         }
+
+        /// <summary>
+        /// Gets the parameter matching the given name.
+        /// </summary>
+        /// <param name="name">The name of the parameter to match.</param>
+        /// <returns>The matching parameter or <c>null</c> if no parameter matches the given name.</returns>
+        public TemplatePart GetParameter(string name)
+        {
+            for (var i = 0; i < Parameters.Count; i++)
+            {
+                var parameter = Parameters[i];
+                if (string.Equals(parameter.Name, name, StringComparison.OrdinalIgnoreCase))
+                {
+                    return parameter;
+                }
+            }
+
+            return null;
+        }
     }
 }
--- a/src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
+++ b/src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
@@ -386,8 +386,7 @@ namespace Microsoft.AspNetCore.Routing.Tree
             {
                 if (entry.RequiredLinkValues.ContainsKey(kvp.Key))
                 {
-                    var parameter = entry.Template.Parameters
-                        .FirstOrDefault(p => string.Equals(p.Name, kvp.Key, StringComparison.OrdinalIgnoreCase));
+                    var parameter = entry.Template.GetParameter(kvp.Key);
 
                     if (parameter == null)
                     {


src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
```diff
@@ -59,5 +59,24 @@ private string DebuggerToString()
         {
             return string.Join(SeparatorString, Segments.Select(s => s.DebuggerToString()));
         }
+
+        /// <summary>
+        /// Gets the parameter matching the given name 
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns>matching parameter or null if none matches</returns>
+        public TemplatePart GetParameter(string name)
```
SajayAntony: This is a public class and hence has API surface area. You might want to move the implementation closer to the callsite.

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -386,8 +386,7 @@ private VirtualPathData GenerateVirtualPath(VirtualPathContext context, TreeRout
             {
                 if (entry.RequiredLinkValues.ContainsKey(kvp.Key))
                 {
-                    var parameter = entry.Template.Parameters
-                        .FirstOrDefault(p => string.Equals(p.Name, kvp.Key, StringComparison.OrdinalIgnoreCase));
+                    var parameter = entry.Template.GetParameter(kvp.Key);
```
SajayAntony: Did you consider removing the enumerator as well for context.Values?

src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
rynowak: This isn't a public API we're worried about breaking changes for.

src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
```diff
@@ -59,5 +59,24 @@ private string DebuggerToString()
         {
             return string.Join(SeparatorString, Segments.Select(s => s.DebuggerToString()));
         }
+
+        /// <summary>
+        /// Gets the parameter matching the given name
```
rynowak: Period

src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
```diff
@@ -59,5 +59,24 @@ private string DebuggerToString()
         {
             return string.Join(SeparatorString, Segments.Select(s => s.DebuggerToString()));
         }
+
+        /// <summary>
+        /// Gets the parameter matching the given name 
+        /// </summary>
+        /// <param name="name"></param>
```
rynowak: Fill this in

src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
```diff
@@ -59,5 +59,24 @@ private string DebuggerToString()
...
+        /// <returns>matching parameter or null if none matches</returns>
```
rynowak: `The matching parameter or <c>null></name> if no parameter matches the given name.`

rynowak:
:shipit: after fixing doc comments

src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
SajayAntony: Ah in that case :see_no_evil:

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
mnltejaswini: I cannot because, RouteValueDictionary.Values is a ValueCollection on Dictionary

SajayAntony: Nice in that case ignore the comment.

SajayAntony:
:shipit:

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
rynowak: `context.Values` is a `RVD` which has a struct enumerator already. I don't think it's causing trouble.

src/Microsoft.AspNetCore.Routing/Template/RouteTemplate.cs
```diff
@@ -59,5 +59,24 @@ private string DebuggerToString()
         {
             return string.Join(SeparatorString, Segments.Select(s => s.DebuggerToString()));
         }
+
+        /// <summary>
+        /// Gets the parameter matching the given name.
+        /// </summary>
+        /// <param name="name">name of the parameter to match.</param>
```
rynowak: Should be "The name". Remember these are complete sentences (sometimes the subject is implied).

rynowak:
:shipit:

Issue #294, opened 2016-03-08
Not sure why but seems the parameter "defaultHandler" is useless here


4engkai:
https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing/RouteBuilder.cs

````
public RouteBuilder(IApplicationBuilder applicationBuilder, IRouter defaultHandler)
{
    if (applicationBuilder == null)
    {
        throw new ArgumentNullException(nameof(applicationBuilder));
    }

    if (applicationBuilder.ApplicationServices.GetService(typeof(RoutingMarkerService)) == null)
    {
        throw new InvalidOperationException(Resources.FormatUnableToFindServices(
            nameof(IServiceCollection),
            nameof(RoutingServiceCollectionExtensions.AddRouting),
            "ConfigureServices(...)"));
    }

    ApplicationBuilder = applicationBuilder;
    ServiceProvider = applicationBuilder.ApplicationServices;

    Routes = new List<IRouter>();
}
````

Eilon:
Ah, that looks like a bug! I think it just needs to set the `DefaultHandler` property to that value.

PR #295, opened 2016-03-10
Set DefaultHandler in RouteBuilder

ajaybhargavb:
Issue #294 

@pranavkm

### DIFF
--- a/test/Microsoft.AspNetCore.Routing.Tests/RouteBuilderTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/RouteBuilderTest.cs
@@ -3,6 +3,8 @@
 
 using System;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Routing.Internal;
+using Microsoft.Extensions.DependencyInjection;
 using Moq;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


davidfowl:
This so needs a test though

ajaybhargavb:
Added.

test/Microsoft.AspNetCore.Routing.Tests/RouteBuilderTest.cs
```diff
@@ -11,6 +12,27 @@ namespace Microsoft.AspNetCore.Routing
...
+        public void Ctor_SetsPropertyValues()
+        {
+            // Arrange
+            var applicationServicesMock = new Mock<IServiceProvider>();
+            applicationServicesMock.Setup(s => s.GetService(typeof(RoutingMarkerService))).Returns(Mock.Of<RoutingMarkerService>);
+            var applicationBuilderMock = new Mock<IApplicationBuilder>();
+            var applicationServices = applicationServicesMock.Object;
+            applicationBuilderMock.Setup(a => a.ApplicationServices).Returns(applicationServices);
+            var applicationBuilder = applicationBuilderMock.Object;
+            var defaultHandler = Mock.Of<IRouter>();
```
rynowak: Can we just use `IServiceCollection` here instead?

rynowak:
:shipit:

test/Microsoft.AspNetCore.Routing.Tests/RouteBuilderTest.cs
ajaybhargavb: updated the test

PR #298, opened 2016-03-15
Fix package metadata

Eilon:
@rynowak @ajaybhargavb @javiercn


src/Microsoft.AspNetCore.Routing/project.json
```diff
@@ -1,5 +1,9 @@
 {
-  "description": "ASP.NET 5 middleware for routing requests to application logic and for generating links.",
+  "description": "ASP.NET Core middleware for routing requests to application logic and for generating links.\r\nCommonly used types:\r\n Microsoft.AspNetCore.Routing.Route\r\nMicrosoft.AspNetCore.Routing.RouteCollection",
```
javiercn: How about ```ASP.NET Core middleware for routing requests and generating links``` the ```to application logic`` part sounds weird, but beware i'm not a native speaker.

javiercn:
:shipit:

src/Microsoft.AspNetCore.Routing/project.json
Eilon: Not a huge fan of it either, but it *is* correct, and I can't think of anything better :smile:

PR #299, opened 2016-03-23
Fixes issue of request path segment length greater than the wildcard template route

kichalla:
@rynowak @ajaybhargavb 

Fixes https://github.com/aspnet/Mvc/issues/4256

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
+++ b/src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
@@ -249,7 +249,17 @@ namespace Microsoft.AspNetCore.Routing.Tree
                 while (_stack.Count > 0)
                 {
                     var next = _stack.Pop();
-                    if (++_segmentIndex >= _tokenizer.Count)
+
+                    // In case of wild card segment, the request path segment length can be greater
+                    // Example:
+                    // Template:    a/{*path}
+                    // Request Url: a/b/c/d
+                    if (next.IsCatchAll && next.Matches.Count > 0)
+                    {
+                        Current = next;
+                        return true;
+                    }
+                    else if (++_segmentIndex >= _tokenizer.Count)
                     {
                         _segmentIndex--;
                         if (next.Matches.Count > 0)
--- a/src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
+++ b/src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
@@ -18,6 +18,8 @@ namespace Microsoft.AspNetCore.Routing.Tree
 
         public int Length { get; }
 
+        public bool IsCatchAll { get; set; }
+
         // These entries are sorted by precedence then template
         public List<TreeRouteMatchingEntry> Matches { get; }
 
--- a/test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
@@ -120,6 +120,39 @@ namespace Microsoft.AspNetCore.Routing.Tree
         }
 
         [Theory]
+        [InlineData("{*path}", "/a", "a")]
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
```diff
@@ -18,6 +18,8 @@ public UrlMatchingNode(int length)
 
         public int Length { get; }
 
+        public bool IsCatchAll { get; set; }
```
kichalla: This node has information of its 'next' nodes, like if they are literal or catch all etc, but does not have information about itself. We need this information to find if the current node is a catch all so that we can avoid the request path segment length check.

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -249,7 +249,19 @@ public bool MoveNext()
                 while (_stack.Count > 0)
                 {
                     var next = _stack.Pop();
-                    if (++_segmentIndex >= _tokenizer.Count)
+
+                    _segmentIndex += 1;
```
rynowak: `++` is objectively better :laughing:

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
```diff
@@ -120,6 +120,41 @@ public class TreeRouterTest
...
+            var attributeRoute = CreateAttributeRoute(next.Object, matchingRoutes, linkGenerationEntries);
+            var context = CreateRouteContext(requestPath);
+
+            // Act
+            await attributeRoute.RouteAsync(context);
+
+            // Assert
+            Assert.NotNull(context.Handler);
+            object value;
+            Assert.True(context.RouteData.Values.TryGetValue("path", out value));
```
ajaybhargavb: Is `TryGetValue` really needed if we expect it to be present?

rynowak: `TryGetValue` is never needed with RVD

src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
rynowak: Is this really more like - has catchall child?

ajaybhargavb:
Can you add some cases where you have multiple matching routes and check if the correct route is selected? For example, have the request match both catchall and a constrained catchall and see if the constrained one gets selected

ajaybhargavb:
:watch:

src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
kichalla: Its like 'am i a catchall node'. the catchall child is captured by `CatchAlls` property.

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
kichalla: sure, will remove

kichalla:
> Can you add some cases where you have multiple matching routes and check if the correct route is selected? For example, have the request match both catchall and a constrained catchall and see if the constrained one gets selected

This should cover it? https://github.com/aspnet/Routing/blob/dev/test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs#L41

ajaybhargavb:
> This should cover it? https://github.com/aspnet/Routing/blob/dev/test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs#L41

It does. Thanks

ajaybhargavb:
:shipit:

src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
rynowak: gotcha

rynowak:
:shipit:

PR #303, opened 2016-04-07
Remove dictionary alloc in routing

rynowak:
This changes TemplateMatcher to mutate RouteData.Values directly instead
of creating a new dictionary and then merging in values. This is one the
biggest single costs in routing in terms of both allocations and execution
time.

So Match now becomes TryMatch. This will dirty the state of the RVD, so
the caller needs to snapshot it before calling into it (handled
inside the TreeRouter or RouteCollection).

Some subtle changes were needed to how/when values are added to be
compatible with the existing tests. The general idea is that we add null
values for non-parameter defaults or catchalls, but only if they don't
trounce an existing value. This logic used to live in MergeValues but now
it's in TryMatch since TryMatch might be working from existing data.

Also fixed the .sln to avoid building a package that we use as shared
source.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
+++ b/src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
@@ -65,6 +65,11 @@ namespace Microsoft.AspNetCore.Routing.Template
 
         public bool TryMatch(PathString path, RouteValueDictionary values)
         {
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
             var i = 0;
             var pathTokenizer = new PathTokenizer(path);
 
--- a/src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
+++ b/src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
@@ -63,7 +63,7 @@ namespace Microsoft.AspNetCore.Routing.Template
 
         public RouteTemplate Template { get; }
 
-        public RouteValueDictionary Match(PathString path)
+        public bool TryMatch(PathString path, RouteValueDictionary values)
         {
             var i = 0;
             var pathTokenizer = new PathTokenizer(path);
@@ -84,7 +84,7 @@ namespace Microsoft.AspNetCore.Routing.Template
                 {
                     // If pathSegment is null, then we're out of route segments. All we can match is the empty
                     // string.
-                    return null;
+                    return false;
                 }
                 else if (routeSegment.IsSimple && routeSegment.Parts[0].IsLiteral)
                 {
@@ -92,7 +92,7 @@ namespace Microsoft.AspNetCore.Routing.Template
                     var part = routeSegment.Parts[0];
                     if (!requestSegment.Equals(part.Text, StringComparison.OrdinalIgnoreCase))
                     {
-                        return null;
+                        return false;
                     }
                 }
                 else if (routeSegment.IsSimple && routeSegment.Parts[0].IsCatchAll)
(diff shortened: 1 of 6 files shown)


rynowak:
#### Sampling Data from HelloWorldMvc (Plaintext)

**Before**
image


**After**
image


#### Allocation Data from HelloWorldMvc (Plaintext)

**Before**
image


**After**
image

I'm experimenting with using dotTrace's ETW collection for this, so the allocation numbers are only comparable as percentages not as totals (different number of requests). So **17.01%** before vs **4.79%** after.

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -166,36 +166,37 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
 
                 var treeEnumerator = new TreeEnumerator(root, tokenizer);
 
+                // Create a snapshot before processing the route. We'll restore this snapshot before running each
+                // to restore the state. This is likely an "empty" snapshot, which doesn't allocate.
+                var snapshot = context.RouteData.PushState(null, null, null);
```
kichalla: Which test in this PR exercises this part of the code? was difficult to find out..

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateMatcherTests.cs
```diff
@@ -773,17 +825,17 @@ public void GetRouteDataDoesMatchesExactLiteralMatch()
         }
 
         [Fact]
-        public void GetRouteDataWithWeirdParameterNames()
+        public void GetRouteDataWithWeimatchParameterNames()
```
kichalla: fix all the find/replace errors

src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
```diff
@@ -63,7 +63,7 @@ public class TemplateMatcher
 
         public RouteTemplate Template { get; }
 
-        public RouteValueDictionary Match(PathString path)
+        public bool TryMatch(PathString path, RouteValueDictionary values)
```
ajaybhargavb: Add a null check for `values`?

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
ajaybhargavb: Named params

rynowak: No new tests were needed for this there are existing tests that verify it, this change just made it more optimal.

src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
rynowak: :+1:

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateMatcherTests.cs
rynowak: Sure, will do a pass for names

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
```diff
@@ -1533,7 +1533,6 @@ public static IEnumerable<object[]> OptionalParamValues
             // Assert
             Assert.NotEqual(nestedValues, context.RouteData.Values);
 
-            // The new routedata is a copy
```
rynowak: This comment was stale, I removed it while I was debugging a failure in this test.

ajaybhargavb:
:shipit: after addressing minor comments

PR #309, opened 2016-04-19
Make RVD better

rynowak:
This change makes RouteValueDictionary into something incredible. Like a butterfly emerging from a chrysalis.

@Eilon are you terrified by this?


rynowak:
Sampling data from HelloWorldMvcCore (Techempower plaintext):
**Before**
image


**After**
image


This change significantly speeds up routing by using a list instead of a dictionary. Operations hashing a string (ordinal ignore case) are taking too much time, and we usually store 3-4 values.

javiercn:
Looks good to me, Is it this [Design]?

rynowak:
> Is it this [Design]?

Yeah, I suppose it is. I'll need to add tests for the state transitions as well as the methods which we previously got for free from dictionary.

javiercn:
Design looks good to me. To summarize the changes you basically have a RVD that starts with an empty storage and upgrades to a list storage when a value is inserted, isn't it? 

Also, the improvement will be 4.36-0.99 (set_item) 3.37?

rynowak:
> Design looks good to me. To summarize the changes you basically have a RVD that starts with an empty storage and upgrades to a list storage when a value is inserted, isn't it?

Yep. Next phase will do even more crazy stuff

> Also, the improvement will be 4.36-0.99 (set_item) 3.37?

It's not quite that straightforward... **everything** in RVD should be faster now.  If you look at the time proportionally it's spending 5.45% in RVD methods before and 1.42% after..

Note that the time measurements are **not** directly comparable, you have to look at proportions of the total time spent in MVC+Routing+App which was 75,610ms before and 70,256ms after.

**Before**
image


**After**
image

test/Microsoft.AspNetCore.Mvc.Routing.Abstractions.Tests/RouteValueDictionaryTests.cs
```diff
@@ -156,21 +156,6 @@ public void CreateFromObject_CopiesPropertiesFromRegularType_WithIndexerProperty
             Assert.Equal(0, dict.Count);
         }
 
-        [Fact]
-        public void CreateFromObject_MixedCaseThrows()
```
dougbu: We added this test because it could thoroughly confuse users if the code silently chose one value over the other. Or perhaps I did cause a customer issue...

Either way, we shouldn't lose this behaviour.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -35,62 +38,51 @@ public RouteValueDictionary()
...
                 {
-                    Properties = otherDictionary.Properties;
-                    Value = otherDictionary.Value;
+                    _storage = new ListStorage(listStorage);
                     return;
                 }
-                else
+
+                PropertyStorage propertyStorage;
+                if ((propertyStorage = dictionary._storage as PropertyStorage) != null)
```
dougbu: Strange to use `else if` in the outer level and not here.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -35,62 +38,51 @@ public RouteValueDictionary()
...
-                Properties = PropertyHelper.GetVisibleProperties(values);
-                Value = values;
-
+                _storage = new PropertyStorage(values, PropertyHelper.GetVisibleProperties(values));
+                return;
+            }
+            else
+            {
+                _storage = EmptyStorage.Instance;
                 return;
```
dougbu: Most of the `return` statements in this constructor are just noise and make things longer. Either use early exits everywhere (removing the `else if` bits) or remove excess `return`s.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -137,17 +132,33 @@ public ICollection<string> Keys
         {
             get
             {
-                EnsureWritable();
-                return InnerDictionary.Keys;
+                Upgrade();
+
+                var list = ((ListStorage)_storage);
```
dougbu: What are the outer parens for?

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -185,22 +214,36 @@ public void Add(string key, object value)
...
+            var list = ((ListStorage)_storage);
+            list.Inner.Add(new KeyValuePair<string, object>(key, value));
         }
 
         /// <inheritdoc />
         public void Clear()
         {
-            EnsureWritable();
-            InnerDictionary.Clear();
+            Upgrade();
```
dougbu: Should exit early if storage is currently `EmptyStorage`.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -185,22 +214,36 @@ public void Add(string key, object value)
...
+            var list = ((ListStorage)_storage);
+            list.Inner.Clear();
         }
 
         /// <inheritdoc />
         bool ICollection<KeyValuePair<string, object>>.Contains(KeyValuePair<string, object> item)
         {
-            EnsureWritable();
-            return ((ICollection<KeyValuePair<string, object>>)InnerDictionary).Contains(item);
+            Upgrade();
```
dougbu: Should exit early if storage is currently `EmptyStorage`. Particularly strange to upgrade for a read operation.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
dougbu: Applies to all delete-ish methods below e.g. `Remove()` should return `false` in `EmptyStorage` case.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -293,114 +344,269 @@ public bool TryGetValue(string key, out object value)
...
+
+                Current = default(KeyValuePair<string, object>);
+                _index = -1;
+            }
+
+            public KeyValuePair<string, object> Current { get; private set; }
+
+            object IEnumerator.Current => Current;
+
+            public void Dispose()
```
dougbu: What is this method for?

dougbu:
:watch:

test/Microsoft.AspNetCore.Mvc.Routing.Abstractions.Tests/RouteValueDictionaryTests.cs
rynowak: Yeah I agree, I think RVD will end up doing it's own cache of PropertyHelper objects. We don't want to lose this error message.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
rynowak: MSDN will [explain it better than I can](https://msdn.microsoft.com/en-us/library/system.idisposable.dispose(v=vs.110).aspx)

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -137,17 +132,33 @@ public ICollection<string> Keys
         {
             get
             {
-                EnsureWritable();
-                return InnerDictionary.Keys;
+                Upgrade();
```
Eilon: Why always upgrade? Why not have the Storage do this?

Eilon: Plus same comment in other similar scenarios. (e.g. generic Keys property below)

---
In reply to: [60490919](https://github.com/aspnet/Routing/pull/309#discussion_r60490919) [](ancestors = 60490919)

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
dougbu: Funny. What I was missing was that `IEnumerator<KeyValuePair<string, object>>` implements `IDisposable`.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
rynowak: There's little payoff to trying to produce optimized implementations of everything on `IDictionary<,>` and `ICollection<>`. All of these extra implementations require their own tests.

The eventual goal of all of this is that an app using a 'normal' MVC action selection scheme or attribute routing doesn't need to `Upgrade` on any of the main code paths.

Eilon:
Overall looks good to me. Just some Upgrade() questions.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
rynowak: Changing this code to use more `if` and early exits.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -35,62 +38,51 @@ public RouteValueDictionary()
         /// </remarks>
         public RouteValueDictionary(object values)
         {
-            var otherDictionary = values as RouteValueDictionary;
-            if (otherDictionary != null)
+            RouteValueDictionary dictionary;
```
javiercn: Do we intentionally want to allow null to be passed to this class? The code implies that we do, but i'm wondering if its intentional.

rynowak: yes.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -35,62 +38,51 @@ public RouteValueDictionary()
...
                 }
-            }
 
-            var keyValuePairCollection = values as IEnumerable<KeyValuePair<string, object>>;
-            if (keyValuePairCollection != null)
+                // If we get here, it's an EmptyStorage
+                _storage = EmptyStorage.Instance;
+                return;
+            }
+            else if ((keyValueEnumerable = values as IEnumerable<KeyValuePair<string, object>>) != null)
```
javiercn: if you have to do return; at the end of each block I think it's clearer if you just write the code as a sequence of ifs, rather than doing if else all around, i think its clearer that way.

rynowak: Why are any of you commenting on this in a [Design] PR?

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -293,114 +344,269 @@ public bool TryGetValue(string key, out object value)
...
-                    InnerDictionary.Add(property.Property.Name, property.ValueGetter(Value));
+                    _inner.Add(other._inner[i]);
                 }
             }
-        }
 
-        public struct Enumerator : IEnumerator<KeyValuePair<string, object>>
-        {
-            private readonly RouteValueDictionary _dictionary;
+            public override int Count => _inner.Count;
```
javiercn: 5 pranav points

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
javiercn: Fair enough :smile:

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
rynowak: There's a few operations on dictionary/collection that aren't commonly used. Typically anything that's explicitly implemented isn't really **ever** going to be called. I'm opting for simplicity in those cases and trying to avoid writing 7-8 tests for every method.

rynowak:
Updated. 

I've added tests, and addressed most of the feedback. I'm in the middle of doing some more thorough testing to see what implementations need to be optimized. I've done a pass and made many of the operations that call upgrade check the count first and early exit where possible.

rynowak:
Removing the design tag on this. I verified with some profiling that we're not really ever upgrading needlessly. All of our upgrades happen because we're writing to the dictionary.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -269,8 +314,30 @@ IEnumerator IEnumerable.GetEnumerator()
...
         bool ICollection<KeyValuePair<string, object>>.Remove(KeyValuePair<string, object> item)
         {
-            EnsureWritable();
-            return ((ICollection<KeyValuePair<string, object>>)InnerDictionary).Remove(item);
+            if (_storage.Count == 0)
+            {
+                return false;
+            }
+
+            if (_storage.Count == 0)
```
dougbu: Like the new `Count` checks. But I suspect you'll find the value hasn't changed since [line 317](https://github.com/aspnet/Routing/pull/309/files#diff-8a515f05675a880115920385ab225fcbR317) :trollface:

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
dougbu: :see_no_evil:

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -293,115 +376,306 @@ public bool TryGetValue(string key, out object value)
...
+            public override void Upgrade(ref Storage storage)
             {
+                // Do nothing.
             }
+        }
 
-            public bool MoveNext()
+        internal class PropertyStorage : Storage
+        {
+            private static readonly PropertyCache _propertyCache = new PropertyCache();
```
dougbu: Ick, another `static` cache on top of those in `PropertyHelper`. Any way we can avoid that? E.g. could expose a `virtual ValidateHelpers(PropertyHelper[] helpers)` method that `PropertyHelper.GetProperties(...)` calls before updating its cache.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -293,115 +376,306 @@ public bool TryGetValue(string key, out object value)
...
+                for (var i = 0; i < _properties.Length; i++)
+                {
+                    var property = _properties[i];
+                    storage.TrySetValue(property.Name, property.GetValue(_value));
+                }
+            }
+
+            private static void ValidatePropertyNames(Type type, PropertyHelper[] properties)
+            {
+                var names = new Dictionary<string, PropertyHelper>(StringComparer.OrdinalIgnoreCase);
```
dougbu: Minor but this could use the `Comparer` property couldn't it?

test/Microsoft.AspNetCore.Mvc.Routing.Abstractions.Tests/RouteValueDictionaryTests.cs
```diff
@@ -4,44 +4,148 @@
 using System;
 using System.Linq;
 using System.Collections.Generic;
-using Microsoft.AspNetCore.Testing;
 using Xunit;
+using Microsoft.AspNetCore.Testing;
```
dougbu: Was ordered correctly before.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -293,115 +376,306 @@ public bool TryGetValue(string key, out object value)
...
+                    names.Add(property.Name, property);
+                }
+            }
+        }
+
+        internal class EmptyStorage : Storage
+        {
+            public static readonly EmptyStorage Instance = new EmptyStorage();
+
+            public override int Count => 0;
```
dougbu: Suggest making the parameterless constructor `private` to keep future-Ryans honest.

test/Microsoft.AspNetCore.Mvc.Routing.Abstractions.Tests/RouteValueDictionaryTests.cs
```diff
@@ -153,75 +269,1121 @@ public void CreateFromObject_CopiesPropertiesFromRegularType_WithIndexerProperty
...
+                $"case-insensitive comparisons.";
+
+            // Act & Assert		
+            var exception = Assert.Throws<InvalidOperationException>(() =>
             {
                 var dictionary = new RouteValueDictionary(obj);
-                dictionary.Add("Hi", "There");
             });
+
+            // Ignoring case to make sure we're not testing reflection's ordering.
```
dougbu: :clap:

test/Microsoft.AspNetCore.Mvc.Routing.Abstractions.Tests/RouteValueDictionaryTests.cs
```diff
@@ -153,75 +269,1121 @@ public void CreateFromObject_CopiesPropertiesFromRegularType_WithIndexerProperty
...
-                yield return new object[] { routeValues };
-            }
+        [Fact]
+        public void IndexGet_ListStorage_Match_ReturnsValue()
+        {
+            // Arrange
+            var dict = new RouteValueDictionary(new List<KeyValuePair<string, object>>()
+            {
+                new KeyValuePair<string, object>("key", "value"),
+            });
```
dougbu: Just
``` c#
new RouteValueDictionary
{
    { "key", "value" },
};
```

dougbu:
:watch: hoping https://github.com/aspnet/Routing/pull/309#discussion-diff-60759496 can be improved.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
rynowak: Here's my proposal - let's add another overload/parameter to all of these https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.PropertyHelper.Sources/PropertyHelper.cs#L166 that accepts a `Func<PropertyInfo, PropertyHelper>`. 

This feature already exists in `PropertyHelper` - it's just not exposed as public.

rynowak: Actually this won't work - we can't let anyone write to our cache because they aren't following the rules that we are.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
rynowak: :+1:

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
rynowak: No, not from a static method in a nested class. :see_no_evil:

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
rynowak: sure

test/Microsoft.AspNetCore.Mvc.Routing.Abstractions.Tests/RouteValueDictionaryTests.cs
rynowak: :+1:

rynowak:
updated

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -293,115 +368,310 @@ public bool TryGetValue(string key, out object value)
...
             {
-                value = null;
-                return false;
+                Current = default(KeyValuePair<string, object>);
+                _index = -1;
             }
         }
 
-        private void EnsureWritable()
+        // Storage and it's subclasses are internal for testing.
```
Eilon: Your write about this comment :trollface:

Eilon:
:shipit: from me.

dougbu:
:shipit:

rynowak:
btw this change takes MVC plaintext from 205k -> 220k /cc

PR #310, opened 2016-04-21
TreeRouter cleanup

rynowak:


src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -73,43 +80,49 @@ public class TreeRouter : IRouter
...
-                    namedEntries.Add(entry.Name, entry);
+                    _namedEntries.Add(entry.RouteName, outboundMatch);
                 }
             }
 
-            _namedEntries = namedEntries;
-
             // The decision tree will take care of ordering for these entries.
-            _linkGenerationTree = new LinkGenerationDecisionTree(linkGenerationEntries.ToArray());
+            _linkGenerationTree = new LinkGenerationDecisionTree(outboundMatches.ToArray());
```
pranavkm: Presumably the `ToArray` isn't necessary since we're working with a list here.

rynowak: The decision tree wants it to be an array, I didn't see much reason to change this.

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
```diff
@@ -787,139 +673,31 @@ public void TreeRouter_GenerateLink_EnsuresStableOrder(string firstTemplate, str
...
-            // Arrange
-            string expectedExceptionMessage = "Two or more routes named 'NamedEntry' have different templates." +
-                Environment.NewLine +
-                "Parameter name: linkGenerationEntries";
-
-            var next = new Mock<IRouter>().Object;
-
-            // Act
-            var builder = new TreeRouteBuilder(next, NullLoggerFactory.Instance);
-            var exception = Assert.Throws<ArgumentException>(
```
pranavkm: :shipit: once these tests are revived.

PR #313, opened 2016-05-11
Optimize manipulation of RouteData.Routers

rynowak:
This change avoid calling `List.Clear()` and `new List(IEnumerable<T>)` which
both end up calling into native methods via the Array static class. These
methods are designed to be performant for large collections, and for our
needs this collection has at most 1-4 items. This is worth 2-3% in
techempower plaintext.

Sampling - MVC Core Plaintext - Notice the improvement in `PushState` and `Restore`
**Before** 
image

**After**
image

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNetCore.Routing
     public class RouteData
     {
         private RouteValueDictionary _dataTokens;
-        private IList<IRouter> _routers;
+        private List<IRouter> _routers;
         private RouteValueDictionary _values;
 
         /// <summary>
@@ -126,10 +126,22 @@ namespace Microsoft.AspNetCore.Routing
         /// <returns>A <see cref="RouteDataSnapshot"/> that captures the current state.</returns>
         public RouteDataSnapshot PushState(IRouter router, RouteValueDictionary values, RouteValueDictionary dataTokens)
         {
+            // Perf: this is optimized for small list sizes, in particular to avoid overhead of a native call in
+            // Array.CopyTo inside the List(IEnumerable<T>) constructor.
+            List<IRouter> routers = null;
+            if (_routers?.Count > 0)
+            {
+                routers = new List<IRouter>();
+                for (var i = 0; i < _routers.Count; i++)
+                {
+                    routers.Add(_routers[i]);
+                }
+            }
+
             var snapshot = new RouteDataSnapshot(
                 this,
                 _dataTokens?.Count > 0 ? new RouteValueDictionary(_dataTokens) : null, 
-                _routers?.Count > 0 ? new List<IRouter>(_routers) : null,
+                routers,
                 _values?.Count > 0 ? new RouteValueDictionary(_values) : null);
 
             if (router != null)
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
```diff
@@ -126,10 +126,22 @@ public RouteValueDictionary Values
         /// <returns>A <see cref="RouteDataSnapshot"/> that captures the current state.</returns>
         public RouteDataSnapshot PushState(IRouter router, RouteValueDictionary values, RouteValueDictionary dataTokens)
         {
+            // Perf: this is optimized for small list sizes, in particular to avoid overhead of a native call in
+            // Array.CopyTo inside the List(IEnumerable<T>) constructor.
+            List<IRouter> routers = null;
+            if (_routers?.Count > 0)
+            {
+                routers = new List<IRouter>();
```
pranavkm: Does it help to specify the capacity here?

rynowak: It won't make a difference for the kinds of sizes we usually ( < 4), but it can't hurt. I'll put it in.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
```diff
@@ -222,15 +234,41 @@ public void Restore()
                 }
                 else if (_routers == null)
                 {
-                    _routeData._routers.Clear();
+                    // Perf: this is optimized for small list sizes, in particular to avoid overhead of a native call in
+                    // Array.Clear inside the List.Clear() method.
+                    var routers = _routeData._routers;
+                    for (var i = routers.Count - 1; i >=0 ; i--)
```
rynowak: yeah I know spacing :trollface:

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
```diff
@@ -222,15 +234,41 @@ public void Restore()
                 }
                 else if (_routers == null)
                 {
-                    _routeData._routers.Clear();
+                    // Perf: this is optimized for small list sizes, in particular to avoid overhead of a native call in
```
pranavkm: Is this something we could contribute to CoreFx?

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
pranavkm: formatting `i >= 0;`

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
rynowak: It's a neat idea, but I'm not going to block on it.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
```diff
@@ -222,15 +234,41 @@ public void Restore()
...
+                    }
+
+                    for (; i < snapshotRouters.Count; i++)
+                    {
+                        routers.Add(snapshotRouters[i]);
+                    }
 
-                    for (var i = 0; i < _routers.Count; i++)
+                    // Trim excess
+                    for (i = routers.Count - 1; i >= snapshotRouters.Count; i--)
```
pranavkm: Can you add a comment here stating `RemoveRange` has the same overhead as `Clear`?

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
pranavkm: Oh sure, just a follow up since there's data backing the change.

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
rynowak: :+1:

PR #314, opened 2016-05-13
[Perf] Refactor TemplateBinder.GetValues to avoid an expensive copy of RouteValueDictionary

mnltejaswini:
Reference Issue https://github.com/aspnet/Mvc/issues/4593 

loadtest with 3000 requests for the home page of the MusicStore app

Before
-------
image

After
------

image

4.8% saving in the total bytes allocated.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
@@ -235,10 +235,6 @@ namespace Microsoft.AspNetCore.Routing.Template
                         // If it's a parameter, get its value
                         object value;
                         var hasValue = acceptedValues.TryGetValue(part.Name, out value);
-                        if (hasValue)
-                        {
-                            acceptedValues.Remove(part.Name);
-                        }
 
                         var isSameAsDefault = false;
                         object defaultValue;
@@ -292,6 +288,12 @@ namespace Microsoft.AspNetCore.Routing.Template
             var wroteFirst = false;
             foreach (var kvp in acceptedValues)
             {
+                if (_template.GetParameter(kvp.Key) != null)
+                {
+                    // We already added the template parameters.
+                    continue;
+                }
+
                 if (_defaults != null && _defaults.ContainsKey(kvp.Key))
                 {
                     // This value is a 'filter' we don't need to put it in the query string.
@@ -494,7 +496,7 @@ namespace Microsoft.AspNetCore.Routing.Template
 
             private string DebuggerToString()
             {
-                return string.Format("{{Accepted: '{0}'}}", string.Join(", ", _acceptedValues.Keys));
+                return string.Format("{{Accepted: '{0}'}}", string.Join(", ", AcceptedValues.Keys));
             }
         }
     }
--- a/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
@@ -80,8 +80,9 @@ namespace Microsoft.AspNetCore.Routing.Template
                 }
             }
 
+            var acceptedValues = GetAcceptedValues(ambientValues, values);
+
             // Validate that all required parameters have a value.
-            // Perf: Fail fast if the required parameter doesn't have a value.
             for (var i = 0; i < _template.Parameters.Count; i++)
             {
                 var parameter = _template.Parameters[i];
@@ -90,14 +91,45 @@ namespace Microsoft.AspNetCore.Routing.Template
                     continue;
                 }
 
-                if (!HasValue(parameter.Name, ambientValues, values))
+                if (!acceptedValues.ContainsKey(parameter.Name))
                 {
                     // We don't have a value for this parameter, so we can't generate a url.
                     return null;
                 }
             }
 
-            
+            // Add any ambient values that don't match parameters - they need to be visible to constraints
+            // but they will ignored by link generation.
+            // Perf: Lazily allocate RouteValueDictionary only when needed as in most cases combined is same as accepted values
+            RouteValueDictionary combinedValues = null;
+            if (ambientValues != null)
+            {
+                foreach (var kvp in ambientValues)
+                {
+                    if (IsRoutePartNonEmpty(kvp.Value))
+                    {
+                        var parameter = GetParameter(kvp.Key);
+                        if (parameter == null && !acceptedValues.ContainsKey(kvp.Key))
+                        {
+                            if (combinedValues == null)
+                            {
+                                combinedValues = new RouteValueDictionary(acceptedValues);
+                            }
+                            combinedValues.Add(kvp.Key, kvp.Value);
+                        }
+                    }
+                }
+            }
+
+            return new TemplateValuesResult()
+            {
+                AcceptedValues = acceptedValues,
+                CombinedValues = combinedValues ?? acceptedValues
+            };
+        }
+
+        private RouteValueDictionary GetAcceptedValues(RouteValueDictionary ambientValues, RouteValueDictionary values)
+        {
             var context = new TemplateBindingContext(_defaults, values);
 
             // Find out which entries in the URI are valid for the URI we want to generate.
@@ -127,12 +159,12 @@ namespace Microsoft.AspNetCore.Routing.Template
                     }
                 }
 
-                // If the parameter is a match, but the value is empty remove it from Accepted values.
+                // If the parameter is a match, add it to the list of values we will use for URI generation
                 if (hasNewParameterValue)
                 {
-                    if (!IsRoutePartNonEmpty(newParameterValue))
+                    if (IsRoutePartNonEmpty(newParameterValue))
                     {
-                        context.Remove(parameterName);
+                        context.Accept(parameterName, newParameterValue);
                     }
                 }
                 else
@@ -144,12 +176,12 @@ namespace Microsoft.AspNetCore.Routing.Template
                 }
             }
 
-            // Check the remaining values for Empty because we will use them for URI generation
+            // Add all remaining new values to the list of values we will use for URI generation
             foreach (var kvp in values)
             {
-                if (!IsRoutePartNonEmpty(kvp.Value))
+                if (IsRoutePartNonEmpty(kvp.Value))
                 {
-                    context.Remove(kvp.Key);
+                    context.Accept(kvp.Key, kvp.Value);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -72,7 +98,7 @@ public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteV
                 var parameterName = parameter.Name;
 
                 object newParameterValue;
-                var hasNewParameterValue = values.TryGetValue(parameterName, out newParameterValue);
+                var hasNewParameterValue = context.AcceptedValues.TryGetValue(parameterName, out newParameterValue);
```
rynowak: Is this change going to modify the dictionary passed in to routing? We can't do that.

mnltejaswini:
The following are the changes

1. TemplateBinder.GetValues() : Fail fast if any of the required parameter of the template doesn't have a value 
2. TemplateBindingContext() : Lazily Initialize the AcceptedValues of the context with the new values. AcceptedValues will be initialized when any of the values has to be updated.For most of the common cases, AcceptedValues is same as the values. In such cases, this optimization saves one RouteValueDictionary allocation per GetValues call.
3. The AcceptedValues will not be changed in TemplateBinder.BindValues

rynowak:
What's the data motivating this change? Is this about making an unsuccessful link generation operation more efficient?

mnltejaswini:
1. The MusicStore app has 3 route templates defined. For every action link when URLHelper.Action is called to get the URL for the link, it is noticed that TemplateBinder.GetValues is called two times for two different routes. For the first route, TemplateBinder returned null as there was no route value defined for required route template parameter "area" . This check failed at the end of the processing where it allocates RouteValueDictionary and copies all other values. We can refactor it to fail fast rather than determining at the end of the processing.

2. In most of the common cases, the AcceptedValues returned by TemplateBinder.BindValues is same as values. But in cases where the AcceptedValues is different, we don't want to update the values given by the user. We can be smarter here to optimize the common cases and postpone the creation of new RouteValueDictionary for AcceptedValues until it is different from values and has to be changed. 

3. TemplateBinder.GetValues, TemplateBinder.BindValues should not change the input route values that were passed to them. This is made so that we can use the optimization mentioned above in point 2.


The perf data corresponding to this change. 
For scenario reference: #4593 
 loadtest -k -n 3000 http://localhost:5000 (home page of the MusicStore app

Before
--------
image

The net benefit of this and https://github.com/aspnet/Mvc/pull/4683 is 
After
-----
image

Total savings is 5.5% in total bytes allocated. 
For MusicStore app home page scenario, saves 2 RVD, List (underlying arrays) per link to 0.

rynowak:
Thanks for the explanation, this is much clearer and a good improvement in the data!

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -388,42 +419,79 @@ private static bool IsRoutePartNonEmpty(object routePart)
...
             {
+                EnsureAccceptedValues();
                 if (!_acceptedValues.ContainsKey(key))
                 {
                     _acceptedValues.Add(key, value);
                 }
+                else
+                {
+                    var existingValue = _acceptedValues[key];
+                    if (!ReferenceEquals(existingValue, value))
```
rynowak: This seems awfully fragile? why not just just compare these as strings?

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -340,6 +344,33 @@ private TemplatePart GetParameter(string name)
...
+            {
+                object value;
+                if (values.TryGetValue(name, out value) && 
+                    IsRoutePartNonEmpty(value))
+                {
+                    return true;
+                }
+            }
+
+            if (ambientValues != null
```
rynowak: I'm going to have to take some time later to re-familiarize myself with this code... why does a `value` have to be non-empty but an `ambient value` or `default` doesn't?

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
mnltejaswini: I know, but I did it intentionally. This class is private to TemplateBinder and this method is called with either of two values (one that is existing in values or _defaults) So I took that liberty of comparing by reference :)

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -58,7 +58,47 @@ public class TemplateBinder
...
+            // Perf: Fail fast if the required parameter doesn't have a value.
+            for (var i = 0; i < _template.Parameters.Count; i++)
+            {
+                var parameter = _template.Parameters[i];
+                if (parameter.IsOptional || parameter.IsCatchAll)
+                {
+                    continue;
+                }
+
+                if (!HasValue(parameter.Name, ambientValues, values))
```
rynowak: This check is not correct because you have to take into account route-value-invalidation. Not all ambient values are eligible to be used in link generation.

Basically I would expect your changes to start generating links in cases where routing would not normally be able to.

Consider this:
```
route: {controller}/{action}/{id?}
ambient values: new { controller = "Home", action = "Index" }
values: new { controller = "Blog" }
```

In this case, the value for `controller` has changed, which means that any ambient values that appear lexically-after `controller` in the template should not be used, so the value of `action` in the ambient values cannot be used. This is handled [here](https://github.com/aspnet/Routing/pull/314/files#diff-46626dd8adef7c6d9adc039d27a7201dL85) in the existing code.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -87,12 +127,12 @@ public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteV
                     }
```
rynowak: I'm not convinced this check is doing the right thing with your changes, it would need to to remove things from the accepted values if they can't be used.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
rynowak: Basically the process here is:
1. Generate accepted values
2. Ensure the accepted values can generate a URL (filters, required parameters)
3. Run constraints
4. Generate URL

You're trying to optimize this code by doing 2 before doing 1, and the problem is that 1 is a fairly complex process. 

Perhaps what you could do is break out 1 into it's own method, with an emphasis on no-op-ing in the common case where the `values` RVD doesn't need to change. Then make sure the rest of the process treats accepted values as readonly.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
rynowak: These kinds of things always require a comment explaining it to future generations.

rynowak:
:watch: unless I'm missing something major here https://github.com/aspnet/Routing/pull/314#discussion_r64644375 this isn't correct in a fairly common case. If all the tests are passing that's likely because some work is being done twice (once incorrectly).

I think you need to evolve the code a bit more. If working with the existing structure and context isn't a good fit then improve it.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -388,47 +397,78 @@ private static bool IsRoutePartNonEmpty(object routePart)
...
+                _values = values;
+                // Perf: In most of the common cases, AcceptedValues is same as the values.
+                // But when they are different, values is readonly and hence we need to create a new
+                // RouteValueDictionary for AcceptedValues.
+                // However to optimize the common cases, we initially start with null acceptedValues.
+                // We will initialize it only, when we find that values is different from acceptedValues.
+                // One case where we know that acceptedValues will be different from values is, when there is
+                // an empty value in values dictionary. If that is the case, we right away intialize the acceptedValues to 
+                // a new RouteValueDictionary. 
+                // This optimization saves one RouteValueDictionary allocation per GetValues call in common cases.
```
rynowak: What's the common case where `values` has an empty value?

rynowak:
This seems overall much better.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
mnltejaswini: As per my understanding, for all the paths where the method GetValues is called internally in MVC doesn't have this case of empty value. But as this is open for extensibility and can be called outside of the internal code, this check may be necessary

Eilon:
@mnltejaswini @rynowak do we want to take this for 1.1.0?

rynowak:
Yes

Eilon:
Okie dokie, then, assigned to you.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -58,7 +58,79 @@ public class TemplateBinder
...
+                if (!acceptedValues.ContainsKey(parameter.Name))
+                {
+                    // We don't have a value for this parameter, so we can't generate a url.
+                    return null;
+                }
+            }
+
+            // Add any ambient values that don't match parameters - they need to be visible to constraints
+            // but they will ignored by link generation.
+            // Perf: Lazily allocate RouteValueDictionary only when needed as in most cases combined is same as accepted values
```
rynowak: I'd really like to see examples of this in the comments. What does the route and the route values look like for a case where they are the same? What about when they are not?

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
rynowak: This is still unresolved, do you have evidence of this?

rynowak:
Code changes here look reasonable, there's a few small things to hammer out. :watch:

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
mnltejaswini: I don't have any evidence

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
mnltejaswini: I looked at the tests and didn't see any such that hits this path. But this change existed even before this PR.

ajaybhargavb:
:up: :date:

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
rynowak: @ajaybhargavb 

> I'd really like to see examples of this in the comments. What does the route and the route values look like for a case where they are the same? What about when they are not?

ajaybhargavb: Missed this part. Will add :+1:

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -388,47 +394,71 @@ private static bool IsRoutePartNonEmpty(object routePart)
...
+
+                        _acceptedValues.Add(key, value);
+                    }
+                }
+                else if (_acceptedValues == null)
+                {
+                    // 'values' dictionary contains an empty value which should not be accepted.
+                    // Eg: The template 'foo/{controller}' with 'values' { "controller", "" }
+                    // should generate 'null'.
+                    _acceptedValues = new RouteValueDictionary();
```
rynowak: I think we really need to document that this is stateful.

Also, I think ordering is an issue here - like if I "see" `{ "controller", "" }` first this seems correct. If I see `{ "action", "Index" }` and then `{ "controller", "" }` it seems like this would be incorrect because `_acceptedValues` will be missing the value for `action`.

The more I think about this change the less I like it.

ajaybhargavb:
Closing this as we want to have this fix as part of a bigger rewrite https://github.com/aspnet/Routing/issues/348

PR #317, opened 2016-06-03
[Fixes #316] TreeRouter does not match a route with the correct length

### LINKED ISSUE #316: No route candidate coming back to Mvc when it should be resolvable
Ok given these routes, and this url `http://example.com/ptslug/slug` it should be clear that the bottom most route and accompanied action should be triggered.

```cs
        [HttpGet]
        public Task<IEnumerable<Model>> GetAll(...) { ... }

        [HttpGet("{ptSlug}")]
        public Task<IEnumerable<Model>> GetAllByPtSlug(...) { ... }

        [HttpGet("{id:int}")]
        public Task<Model> GetById(...) { ... }

        [HttpGet("{ptSlug}/{slug}")]
        public Task<Model> GetByPath(...) { ... }
```

However, as I have debugged the http://github.com/aspnet/Mvc project it does not even get to the point of selecting an action, which normally happens in the ActionSelector. Breakpointing this class revealed that it did not get entered.

Changing this routes' `Order` to any number (either putting it in front of all the other routes or behind them) fixes this.
```cs
        [...]

        [HttpGet("{ptSlug}/{slug}", Order=1)]
        public Task<Model> GetByPath(...) { ... }
```
In turn it enters the ActionSelector and onwards into my code. 

What is going on here?

javiercn:


src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -275,14 +272,24 @@ public bool MoveNext()
                         Current = next;
                         return true;
                     }
-                    else if (++_segmentIndex >= _tokenizer.Count)
+                    // Next template has the same length as the url we are trying to match
+                    // The only possible matching segments are either our current matches or
+                    // any catch-all segment after this segment in which the catch all is empty.
+                    else if (next.Length == _tokenizer.Count)
```
javiercn: This is the core of the fix for the bug

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -235,14 +235,11 @@ private struct TreeEnumerator : IEnumerator<UrlMatchingNode>
             private readonly Stack<UrlMatchingNode> _stack;
             private readonly PathTokenizer _tokenizer;
 
-            private int _segmentIndex;
```
javiercn: We don't need to keep a segment index because each UrlMatchingNode contains a Length property that indicates its depth inside the tree

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -174,7 +175,6 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
             foreach (var tree in _trees)
             {
                 var tokenizer = new PathTokenizer(context.HttpContext.Request.Path);
-                var enumerator = tokenizer.GetEnumerator();
```
javiercn: This wasn't being used.

javiercn:
```C#
             public bool MoveNext()
             {
                 if (_stack == null)
                 {
                     return false;
                 }
```
We want to throw in TreeEnumerator.MoveNext() if the stack is null. It means that the struct hasn't been properly initialized. (As we new up the stack in the constructor).

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
```diff
@@ -65,6 +65,65 @@ public class TreeRouterTest
...
+        [InlineData("/1/2", "{constrained1:int}/{constrained2:int}")]
+        [InlineData("/1/2/3", "{constrained1:int}/{constrained2:int}/{constrained3:int}")]
+        [InlineData("/1/2/3/4", "{constrained1:int}/{constrained2:int}/{constrained3:int}/{*constrainedCatchAll:int}")]
+        [InlineData("/1/2/3/CatchAll4", "{constrained1:int}/{constrained2:int}/{constrained3:int}/{*catchAll}")]
+        [InlineData("/parameter1", "{parameter1}")]
+        [InlineData("/parameter1/parameter2", "{parameter1}/{parameter2}")]
+        [InlineData("/parameter1/parameter2/parameter3", "{parameter1}/{parameter2}/{parameter3}")]
+        [InlineData("/parameter1/parameter2/parameter3/4", "{parameter1}/{parameter2}/{parameter3}/{*constrainedCatchAll:int}")]
+        [InlineData("/parameter1/parameter2/parameter3/CatchAll4", "{parameter1}/{parameter2}/{parameter3}/{*catchAll}")]
+        public async Task TreeRouter_RouteAsync_MatchesRouteWithTheRightLength(string url, string expected)
```
javiercn: The goal of this test is to ensure that the iteration logic enumerates entries in the right order and stops iterating at the right point in the presence of a more "complete" set of routes.

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -275,19 +273,23 @@ public bool MoveNext()
...
                     {
-                        _segmentIndex--;
                         if (next.Matches.Count > 0)
                         {
                             Current = next;
                             return true;
                         }
-                    }
-
-                    if (_tokenizer.Count == 0)
```
javiercn: This case is not required as it's covered by https://github.com/aspnet/Routing/pull/317/files#diff-dc035bd90601467ff505d414a73d85deR279

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
```diff
@@ -221,6 +272,36 @@ private void AddEntryToTree(UrlMatchingTree tree, InboundRouteEntry entry)
...
+             *                                     CatchAlls: {*parameters}
+             *                                                    Matches: '{Controller}/{Action}/{*parameters}'
+             * 
+             * When the tree router tries to match a route, it iterates the list of url matching trees
+             * in ascending order. For each tree it traverses each node starting from the root in the
+             * following order: Literals, constrained parameters, parameters, constrained catch all routes, catch alls.
+             * When it gets to a node of the same length as the route its trying to match, it simply looks at the list of
+             * candidates (which is in precence order) and tries to match the url against it.
+             */
+
```
javiercn: I've tried to explain the tree construction to the best of my knowledge. @rynowak if you can express it better or in a more accurate way, just use this comment as a base and tell me what to replace it with.

javiercn:
@rynowak

src/Microsoft.AspNetCore.Routing/Tree/InboundMatch.cs
```diff
@@ -1,14 +1,30 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Diagnostics;
 using Microsoft.AspNetCore.Routing.Template;
 
 namespace Microsoft.AspNetCore.Routing.Tree
 {
+    /// <summary>
+    /// A candidate route to match incomming URLs in a <see cref="TreeRouter"/>.
```
rynowak: incoming

src/Microsoft.AspNetCore.Routing/Tree/InboundMatch.cs
```diff
@@ -1,14 +1,30 @@
...
+        /// Gets or sets the <see cref="InboundRouteEntry"/>.
+        /// </summary>
         public InboundRouteEntry Entry { get; set; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="TemplateMatcher"/>.
+        /// </summary>
         public TemplateMatcher TemplateMatcher { get; set; }
+
+        internal string DebuggerToString()
```
rynowak: private

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
```diff
@@ -176,17 +203,37 @@ public class TreeRouteBuilder
...
 
+        /// <summary>
+        /// Builds a <see cref="TreeRouter"/> with the <see cref="InboundEntries"/>
+        /// and <see cref="OutboundEntries"/> defined in this <see cref="TreeRouteBuilder"/>.
+        /// </summary>
+        /// <param name="version">The version of the <see cref="TreeRouter"/>.</param>
+        /// <returns>The <see cref="TreeRouter"/>.</returns>
         public TreeRouter Build(int version)
         {
+            // Tree route builder build a tree for each of the different route orders defined by
```
rynowak: build => builds

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
rynowak: Don't use block style comments, I can't think of anywhere else in our codebase we do this.

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
rynowak: Discussed in person - the issue here was that we weren't correctly updating `_segementIndex` on all code paths. As @javiercn has discovered, we don't really need `_segmentIndex` we can just look at the current node's Length.

src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
```diff
@@ -16,21 +26,56 @@ public UrlMatchingNode(int length)
             Literals = new Dictionary<string, UrlMatchingNode>(StringComparer.OrdinalIgnoreCase);
         }
 
+        /// <summary>
+        /// Gets the length of the path to this node in the <see cref="UrlMatchingTree"/>.
+        /// </summary>
         public int Length { get; }
```
rynowak: Please rename this to `Depth` - that will make things clearer overall.

src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
```diff
@@ -16,21 +26,56 @@ public UrlMatchingNode(int length)
...
 
+        /// <summary>
+        /// Gets or sets a value indicating wether this node represents a catch all segment.
```
rynowak: whether

src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
```diff
@@ -16,21 +26,56 @@ public UrlMatchingNode(int length)
...
+        /// </remarks>
         public List<InboundMatch> Matches { get; }
 
+        /// <summary>
+        /// Gets the list of literal segments following this segment.
+        /// </summary>
         public Dictionary<string, UrlMatchingNode> Literals { get; }
 
+        /// <summary>
+        /// Gets or sets the list of <see cref="UrlMatchingNode"/> representing
```
rynowak: These are not lists

src/Microsoft.AspNetCore.Routing/Tree/UrlMatchingNode.cs
```diff
@@ -16,21 +26,56 @@ public UrlMatchingNode(int length)
...
+        /// </summary>
         public UrlMatchingNode ConstrainedCatchAlls { get; set; }
 
+        /// <summary>
+        /// Gets or sets the list of <see cref="UrlMatchingNode"/> representing
+        /// catch all parameter segments following this segment in the <see cref="TreeRouter"/>.
+        /// </summary>
         public UrlMatchingNode CatchAlls { get; set; }
+
+        internal string DebuggerToString()
```
rynowak: private

rynowak:
:shipit: after fixing minor issues. This fix has a nice simplification

src/Microsoft.AspNetCore.Routing/Tree/InboundMatch.cs
javiercn: We do internal on other occurrences for these types of methods (I basically copied one as a sample).

javiercn:
@rynowak updated, i'm merging, just FYI

PR #327, opened 2016-06-14
[Fixes #324] Default parameters are not taken into account on attribute routing

### LINKED ISSUE #324: Request not matching route with defaults
A request like `/api/test` is resulting in 404. A request like `/api/test/hello` works fine as expected.

```c#
public class TestController : Controller
{
    [Route("api/test/{seg1=foo}")]
    public string onevariablesegmentwithdefaults(string seg1 = null)
    {
        return $"onevariablesegmentwithdefaults:{seg1}";
    }
}
```

This is working fine with RC2 final packages, so looks like a regression

@rynowak

javiercn:


test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
```diff
@@ -125,6 +126,58 @@ public class TreeRouterTest
             Assert.Equal(expectedRouteGroup, context.RouteData.Values["test_route_group"]);
         }
 
+        public static TheoryData<string, object[]> MatchesRoutesWithDefaultsData =>
```
kichalla: Nice

kichalla: Since you mentioned about tests not present in this area, could you add a test/variation where there are parameters with constraints and also default values. Ex: `{seg1:int=1}/{seg2:int=2}`

Eilon:
@kichalla assigning to you for final review, though it sounds to me that you're waiting for @javiercn to update the tests.

javiercn:
@kichalla This is not a candidate fix anymore, added tests

kichalla:
:shipit:

rynowak:
:shipit:

Issue #332, opened 2016-06-24
Add an overload of UseRouter that takes Action<IRouteBuilder>


rynowak:
This would be simple to add and would make configuring routing standalone much less verbose.

```
app.UseRouter(routes =>
{
    routes.MapRoute(...);
});

davidfowl:
👍

rynowak:
4adc693

PR #337, opened 2016-06-30
Add an UseRouter that takes Action<IRouteBuilder>

Muchiachio:
- Added an overload of UseRouter that takes Action<IRouteBuilder>, to make configuring standalone routing much less verbose.

Addresses #332

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/BuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Routing/BuilderExtensions.cs
@@ -61,7 +61,7 @@ namespace Microsoft.AspNetCore.Builder
                 throw new ArgumentNullException(nameof(action));
             }
 
-            IRouteBuilder routeBuilder = new RouteBuilder(builder);
+            var routeBuilder = new RouteBuilder(builder);
             action(routeBuilder);
 
             return builder.UseRouter(routeBuilder.Build());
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Routing/BuilderExtensions.cs
```diff
@@ -41,5 +41,30 @@ public static IApplicationBuilder UseRouter(this IApplicationBuilder builder, IR
...
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (action == null)
+            {
+                throw new ArgumentNullException(nameof(action));
+            }
+
+            IRouteBuilder routeBuilder = new RouteBuilder(builder);
```
rynowak: use `var` here

rynowak:
Would be nice to see a test of this in action, even if it's not really a "unit test"

Muchiachio:
Like the ones in sample pages? Because I tried adding unit tests, but that configuration buried behind a lot of layers :smiley:

rynowak:
It shouldn't be hard to do using something like https://github.com/aspnet/Routing/blob/dev/test/Microsoft.AspNetCore.Routing.Tests/RequestDelegateRouteBuilderExtensionsTest.cs as a starting point. If this makes you barf let me know, I don't mind doing it as part of the merge.

Muchiachio:
Yes that would be great, sorry I couldn't add them myself :disappointed_relieved:

rynowak:
No problem, I'll take care of it.

rynowak:
4adc693cb52c3b70175721ed9c474cde46409f96 and 0978a7dd166472f30ac40138fcc71da4cfb1df61 merged. Thanks :+1:

PR #342, opened 2016-07-13
Optimize ListStorage in RouteValueDictionary

ajaybhargavb:
@rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
@@ -77,7 +77,7 @@ namespace Microsoft.AspNetCore.Routing
                         throw new ArgumentException(message, nameof(values));
                     }
 
-                    listStorage._inner.Add(kvp);
+                    listStorage.Add(kvp);
                 }
 
                 return;
@@ -96,7 +96,7 @@ namespace Microsoft.AspNetCore.Routing
                         throw new ArgumentException(message, nameof(values));
                     }
 
-                    listStorage._inner.Add(new KeyValuePair<string, object>(kvp.Key, kvp.Value));
+                    listStorage.Add(new KeyValuePair<string, object>(kvp.Key, kvp.Value));
                 }
 
                 return;
@@ -162,7 +162,7 @@ namespace Microsoft.AspNetCore.Routing
             {
                 Upgrade();
 
-                var list = ((ListStorage)_storage)._inner;
+                var list = (ListStorage)_storage;
                 var keys = new string[list.Count];
                 for (var i = 0; i < keys.Length; i++)
                 {
@@ -188,7 +188,7 @@ namespace Microsoft.AspNetCore.Routing
             {
                 Upgrade();
 
-                var list = ((ListStorage)_storage)._inner;
+                var list = (ListStorage)_storage;
                 var values = new object[list.Count];
                 for (var i = 0; i < values.Length; i++)
                 {
@@ -223,7 +223,7 @@ namespace Microsoft.AspNetCore.Routing
 
             Upgrade();
 
-            var list = ((ListStorage)_storage)._inner;
+            var list = (ListStorage)_storage;
             for (var i = 0; i < list.Count; i++)
             {
                 if (string.Equals(list[i].Key, key, StringComparison.OrdinalIgnoreCase))
@@ -246,7 +246,7 @@ namespace Microsoft.AspNetCore.Routing
 
             Upgrade();
 
-            var list = ((ListStorage)_storage)._inner;
+            var list = (ListStorage)_storage;
             list.Clear();
         }
 
@@ -260,7 +260,7 @@ namespace Microsoft.AspNetCore.Routing
 
             Upgrade();
 
-            var list = ((ListStorage)_storage)._inner;
+            var list = (ListStorage)_storage;
             for (var i = 0; i < list.Count; i++)
             {
                 if (string.Equals(list[i].Key, item.Key, StringComparison.OrdinalIgnoreCase))
@@ -305,7 +305,7 @@ namespace Microsoft.AspNetCore.Routing
 
             Upgrade();
 
-            var list = ((ListStorage)_storage)._inner;
+            var list = (ListStorage)_storage;
             list.CopyTo(array, arrayIndex);
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -670,6 +669,104 @@ public override void Upgrade(ref Storage storage)
             }
         }
 
+        internal class KeyValuePairList
```
rynowak: This shouldn't be here, just put the code you need inside `ListStorage`. This will save a pointer, which is significant in a lot of our usage of this

rynowak:
Perf data?

rynowak:
:watch:

ajaybhargavb:
Perf Data:

Before:

                            Method |        Median |     StdDev |
---------------------------------- |-------------- |----------- |
                         CreateRvd |    65.8707 ns |  1.6964 ns |
           CreateRvd_SetProp_Clear |   125.9218 ns |  4.1885 ns |
 CreateRvd_SetMultipleProps_Remove | 1,212.2968 ns | 28.6963 ns |



After:

                            Method |        Median |     StdDev |
---------------------------------- |-------------- |----------- |
                         CreateRvd |    60.2613 ns |  1.1405 ns |
           CreateRvd_SetProp_Clear |    92.1897 ns | 22.6009 ns |
 CreateRvd_SetMultipleProps_Remove | 1,317.1857 ns | 33.5729 ns |

ajaybhargavb:
:up: :date:

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -439,42 +439,107 @@ internal abstract class Storage
...
         internal class ListStorage : Storage
         {
-            internal readonly List<KeyValuePair<string, object>> _inner;
+            private KeyValuePair<string, object>[] _items;
+            private int _size;
 
             public ListStorage()
             {
-                _inner = new List<KeyValuePair<string, object>>();
+                _items = new KeyValuePair<string, object>[0];
```
rynowak: Don't do this, keep a `readonly static` empty array and use that. The empty array is immutable so it can be shared. This is one of the many optimizations `List<T>` does.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -439,42 +439,107 @@ internal abstract class Storage
...
+                        {
+                            newItems[i] = _items[i];
+                        }
+                        _items = newItems;
+                    }
+                }
+            }
 
-            public override KeyValuePair<string, object> this[int index] => _inner[index];
+            public override int Count => _size;
```
rynowak: rename this to `_count` for consistency

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -439,42 +439,107 @@ internal abstract class Storage
...
+
+            public override KeyValuePair<string, object> this[int index]
```
rynowak: Does this need to do bounds checking? Wouldn't hurt to add an assert here.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -439,42 +439,107 @@ internal abstract class Storage
...
+                _items = new KeyValuePair<string, object>[other.Capacity];
+                for (var i = 0; i < other.Count; i++)
                 {
-                    _inner.Add(other._inner[i]);
+                    this.Add(other[i]);
                 }
             }
 
-            public override int Count => _inner.Count;
+            public int Capacity
```
rynowak: Why does this property need to exist, can't you just do all of this in `EnsureCapacity`?

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -559,6 +632,10 @@ public PropertyStorage(PropertyStorage propertyStorage)
                     var property = _properties[index];
                     return new KeyValuePair<string, object>(property.Name, property.GetValue(_value));
                 }
+                set
+                {
+                    // PropertyStorage never sets a value.
```
rynowak: Should this throw? no-op seems bad

rynowak:
Let's review the perf numbers when you get in, I'd like to understand what's in these tests.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
ajaybhargavb: Added.

ajaybhargavb:
:up: :date: 

Added a test. Didn't add other tests for `RemoveAt` and `Clear` because there are tests already covering that code path.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -486,25 +559,25 @@ public override bool ContainsKey(string key)
...
                     if (string.Equals(key, kvp.Key, StringComparison.OrdinalIgnoreCase))
                     {
-                        _inner[i] = new KeyValuePair<string, object>(key, value);
+                        _items[i] = new KeyValuePair<string, object>(key, value);
                         return true;
                     }
                 }
 
-                _inner.Add(new KeyValuePair<string, object>(key, value));
+                this.Add(new KeyValuePair<string, object>(key, value));
```
rynowak: no need for `this.`

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -520,6 +593,21 @@ public override void Upgrade(ref Storage storage)
             {
                 // Do nothing.
             }
+
+            private void EnsureCapacity(int min)
+            {
+                if (_items.Length < min)
```
rynowak: This check is duplicated between this method and `Add`

ajaybhargavb:
:up: :date: 
Added a test for `Remove`.

rynowak:
:shipit:

PR #345, opened 2016-07-19
[Fixes #4991] Misleading MissingMethodException message for incorrect…

kichalla:
… routes

@rynowak Let me know if this is ok and I can add tests

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
+++ b/src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
@@ -69,12 +69,25 @@ namespace Microsoft.AspNetCore.Routing
 
             if (!typeof(IRouteConstraint).GetTypeInfo().IsAssignableFrom(constraintType.GetTypeInfo()))
             {
-                throw new InvalidOperationException(
+                throw new RouteCreationException(
                             Resources.FormatDefaultInlineConstraintResolver_TypeNotConstraint(
                                                         constraintType, constraintKey, typeof(IRouteConstraint).Name));
             }
 
-            return CreateConstraint(constraintType, argumentString);
+            try
+            {
+                return CreateConstraint(constraintType, argumentString);
+            }
+            catch (RouteCreationException)
+            {
+                throw;
+            }
+            catch (Exception exception)
+            {
+                throw new RouteCreationException(
+                    $"An error occurred while trying to create an instance of route constraint '{constraintType.FullName}'.",
+                    exception);
+            }
         }
 
         private static IRouteConstraint CreateConstraint(Type constraintType, string argumentString)
@@ -107,7 +120,7 @@ namespace Microsoft.AspNetCore.Routing
 
                 if (constructorMatches == 0)
                 {
-                    throw new InvalidOperationException(
+                    throw new RouteCreationException(
                                 Resources.FormatDefaultInlineConstraintResolver_CouldNotFindCtor(
                                                        constraintTypeInfo.Name, arguments.Length));
                 }
@@ -118,7 +131,7 @@ namespace Microsoft.AspNetCore.Routing
                 }
                 else
                 {
-                    throw new InvalidOperationException(
+                    throw new RouteCreationException(
                                 Resources.FormatDefaultInlineConstraintResolver_AmbiguousCtors(
                                                        constraintTypeInfo.Name, arguments.Length));
                 }
--- a/src/Microsoft.AspNetCore.Routing/RouteBase.cs
+++ b/src/Microsoft.AspNetCore.Routing/RouteBase.cs
@@ -46,12 +46,19 @@ namespace Microsoft.AspNetCore.Routing
             ConstraintResolver = constraintResolver;
             DataTokens = dataTokens ?? new RouteValueDictionary();
 
-            // Data we parse from the template will be used to fill in the rest of the constraints or
-            // defaults. The parser will throw for invalid routes.
-            ParsedTemplate = TemplateParser.Parse(template);
+            try
+            {
+                // Data we parse from the template will be used to fill in the rest of the constraints or
+                // defaults. The parser will throw for invalid routes.
+                ParsedTemplate = TemplateParser.Parse(template);
 
-            Constraints = GetConstraints(constraintResolver, ParsedTemplate, constraints);
-            Defaults = GetDefaults(ParsedTemplate, defaults);
+                Constraints = GetConstraints(constraintResolver, ParsedTemplate, constraints);
+                Defaults = GetDefaults(ParsedTemplate, defaults);
+            }
+            catch (Exception exception)
+            {
+                throw new RouteCreationException(Resources.FormatTemplateRoute_Exception(name, template), exception);
+            }
         }
 
         public virtual IDictionary<string, IRouteConstraint> Constraints { get; protected set; }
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
```diff
@@ -82,7 +82,16 @@ private static IRouteConstraint CreateConstraint(Type constraintType, string arg
...
             if (argumentString == null)
             {
-                return (IRouteConstraint)Activator.CreateInstance(constraintType);
+                try
+                {
+                    return (IRouteConstraint)Activator.CreateInstance(constraintType);
+                }
+                catch (Exception exception)
+                {
+                    throw new InvalidRouteException(
```
rynowak: @Eilon - are you ok with changing the type of exception thrown here for 1.1.0?

We need to catch this in MVC, and don't want to catch everything.

src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
```diff
@@ -82,7 +82,16 @@ private static IRouteConstraint CreateConstraint(Type constraintType, string arg
...
+                        $"An error occurred while trying to create an instance of route constraint '{constraintType.FullName}'.",
+                        exception);
+                }
```
rynowak: Assuming we're OK with this approach, we probably want to throw this in place of a few other exceptions here, and also want to surround the calls to user code

src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
Eilon: Switching away from MissingMethodException seems fine. Unlikely someone would have depended on that.

src/Microsoft.AspNetCore.Routing/InvalidRouteException.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class InvalidRouteException : Exception
```
Eilon: This doesn't seem to properly represent the exception case it's being used for. There's nothing wrong with the *route* as far as I can tell. The route is great. But something else in the system is busted: a particular constraint cannot be created for some unknown reason.

kichalla: I was thinking that if we might use this exception in some other cases we might need a more general name(assuming we might not prefer to create more custom exception types).

Eilon: I'd rather be more conservative and write code for stuff that we know happens. It sounds like @rynowak has some thoughts on other scenarios we care about. But this type name doesn't seem like a good match.

kichalla:
:up: :date:

src/Microsoft.AspNetCore.Routing/RouteProcessingException.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class RouteProcessingException : Exception
```
Eilon: Add a doc comment to this?

Eilon:
Just need a doc comment, then :shipit:

src/Microsoft.AspNetCore.Routing/RouteProcessingException.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public RouteProcessingException(string message) :
```
rynowak: colon should go on the next line

src/Microsoft.AspNetCore.Routing/RouteProcessingException.cs
rynowak: Yeah docs here would be good

rynowak:
Can you do a quick search for other uses of `throw` in routing? I think the ones thrown from invalid route templates should use the same exception

kichalla:
> Can you do a quick search for other uses of throw in routing? I think the ones thrown from invalid route templates should use the same exception

I did not use this exception in other scenarios because they provide enough contextual information for debugging. For example, an invalid route `[Route("/a//b")]` generates this error: https://jsfiddle.net/m89gqsa6/

rynowak:
Even if the error messages are better, we now have a "we couldn't create this route" exception type - ideally we'd use it for all routes that fall into that category.

kichalla:
:up: :date:

test/Microsoft.AspNetCore.Routing.Tests/RouteCollectionTest.cs
```diff
@@ -290,7 +290,7 @@ public void GetVirtualPath_AmbiguousRoutes_RequiresRouteValueValidation_Error()
             var virtualPathContext = CreateVirtualPathContext("Ambiguous");
 
             // Act & Assert
-            var ex = Assert.Throws<InvalidOperationException>(() => routeCollection.GetVirtualPath(virtualPathContext));
```
rynowak: Revert these - this shouldn't change

rynowak:
:watch:

src/Microsoft.AspNetCore.Routing/RouteCollection.cs
```diff
@@ -102,7 +102,7 @@ public virtual VirtualPathData GetVirtualPath(VirtualPathContext context)
                 if (namedRoutePathData != null && pathData != null)
                 {
                     var message = Resources.FormatNamedRoutes_AmbiguousRoutesFound(context.RouteName);
-                    throw new InvalidOperationException(message);
+                    throw new RouteCreationException(message);
```
kichalla: Good point.

kichalla:
:up: :date:

src/Microsoft.AspNetCore.Routing/RouteBase.cs
```diff
@@ -46,12 +46,19 @@ public abstract class RouteBase : IRouter, INamedRouter
...
+                // Data we parse from the template will be used to fill in the rest of the constraints or
+                // defaults. The parser will throw for invalid routes.
+                ParsedTemplate = TemplateParser.Parse(template);
 
-            Constraints = GetConstraints(constraintResolver, ParsedTemplate, constraints);
-            Defaults = GetDefaults(ParsedTemplate, defaults);
+                Constraints = GetConstraints(constraintResolver, ParsedTemplate, constraints);
+                Defaults = GetDefaults(ParsedTemplate, defaults);
+            }
+            catch (Exception exception)
```
rynowak: Can we tighten up the scope of this? For instance - this could rewrap a `RouteCreationException` thrown by `GetConstraints` - is that what we want?

kichalla: Yeah this was intentional because any RouteCreationException thrown from within constraints or other methods here would not give contextual information like route name and template and to accommodate that we might need to pass on those details around.

rynowak: Discussed this and it seems fine

rynowak:
:shipit:

PR #347, opened 2016-07-23
! RegexRouteConstraint only did a partial match

artiomchi:
Seems like the RegexRouteConstraint only did a partial match on route values (which is not what the docs suggested)

e.g. Having a route with a parameter like `{ssn:regex(d{3}-d{2}-d{4})}` would match `123-45-6789`, but would also match `alpha123-45-6789`

The proposed change enforces start and end of string characters in the pattern.

If a regular expression constraint with a partial match is required, I'd suggest adding a separate constraint for that (`PartialRegexRouteConstraint` with an inline constraint name of `partialregex`?)

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Constraints/RegexInlineRouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/Constraints/RegexInlineRouteConstraint.cs
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.Routing.Constraints
         /// </summary>
         /// <param name="regexPattern">The regular expression pattern to match.</param>
         public RegexInlineRouteConstraint(string regexPattern)
-            : base(regexPattern)
+            : base("^" + regexPattern + "$")
         {
         }
     }
--- a/src/Microsoft.AspNetCore.Routing/Constraints/PartialRegexInlineRouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/Constraints/PartialRegexInlineRouteConstraint.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Routing.Constraints
+{
+    /// <summary>
+    /// Represents a partial match regex constraint which can be used as an inlineConstraint.
+    /// </summary>
+    public class PartialRegexInlineRouteConstraint : RegexRouteConstraint
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RegexInlineRouteConstraint" /> class.
+        /// </summary>
+        /// <param name="regexPattern">The regular expression pattern to match.</param>
+        public PartialRegexInlineRouteConstraint(string regexPattern)
+            : base(regexPattern)
+        {
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Routing/Constraints/RegexRouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/Constraints/RegexRouteConstraint.cs
@@ -19,7 +19,7 @@ namespace Microsoft.AspNetCore.Routing.Constraints
                 throw new ArgumentNullException(nameof(regex));
             }
 
-            Constraint = regex;
+            Constraint = "^" + regex + "$";
         }
 
         public RegexRouteConstraint(string regexPattern)
@@ -30,7 +30,7 @@ namespace Microsoft.AspNetCore.Routing.Constraints
             }
 
             Constraint = new Regex(
-                regexPattern,
+                "^" + regexPattern + "$",
                 RegexOptions.CultureInvariant | RegexOptions.IgnoreCase,
                 RegexMatchTimeout);
         }
--- a/src/Microsoft.AspNetCore.Routing/Constraints/RegexRouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/Constraints/RegexRouteConstraint.cs
@@ -19,7 +19,7 @@ namespace Microsoft.AspNetCore.Routing.Constraints
                 throw new ArgumentNullException(nameof(regex));
             }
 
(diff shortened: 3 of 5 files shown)


artiomchi:
The docs page with misleading examples:

https://docs.asp.net/en/latest/fundamentals/routing.html#route-constraint-reference

src/Microsoft.AspNetCore.Routing/Constraints/RegexInlineRouteConstraint.cs
```diff
@@ -13,7 +13,7 @@ public class RegexInlineRouteConstraint : RegexRouteConstraint
         /// </summary>
         /// <param name="regexPattern">The regular expression pattern to match.</param>
         public RegexInlineRouteConstraint(string regexPattern)
-            : base(regexPattern)
+            : base("^" + regexPattern + "$")
```
Eilon: Unfortunately, this is a breaking change so we cannot take it.

Perhaps we could reverse the behavior. That is, keep this type unchanged, but have the other one be something like `FullMatchRegexInlineRouteConstraint`.

Thoughts?

artiomchi: I was thinking about that as well, initially..

After your comment and while typing a reply I did several tests on both the old and the new framework, and realised that inline regex constraints work differently to the constraints route parameter. So while the constraints route parameter creates a bounded regex match, the inline constraints does not. If I remember correctly, the inline constraints were initially added in WebAPI 2, and the behaviour there was the same as it is currently in ASP.NET Core.

To summarise, given the following routes:

    routes.MapRoute(
        name: "test1",
        template: "test/{value}",
        defaults: new { controller = "Test", action = "Index" },
        constraints: new { value = "val.*" });

    routes.MapRoute(
        name: "test3",
        template: "test2/{value:regex(val.*)}",
        defaults: new { controller = "Test", action = "Index" });

We have the following behaviour:

 * `/test/val123' => Test.Index()
 * `/test/xval123' => 404
 * `/test2/val123' => Test.Index()
 * `/test2/xval123' => Test.Index()

While I find this behaviour is potentially a cause for confusion and source of potentially bad routes, I also understand that it's been like this since WebAPI 2, so it's definitely not an easy change to make.

With that in mind, I think I understand why my original request might not be such a great idea, but at the same time I'm not very pleased with the above either. It would be great if there was a way to somehow warn the users of this difference in behaviour.

I'm not 100% sure what would be the best decision here (perhaps a `FullMatchRegexInlineRouteConstraint` might be a good solution after all), but at the very least it shows that the current doc page is wrongly describing how the inline regex constraint works, so even if the above pull request is rejected, the docs have to be fixed.

I can update the pull request with the reversed behaviour,

javiercn:
Does anchoring the regex with ^ and $ work? I would think that the current behavior is by design and that is the right choice given the fact that if you want a full match you can just anchor the regex and that if we provided that behavior out of the box, it would be impossible for a user to remove. ^ and $ clearly represent your intention with the regex and don't require an extra concept. Does it make sense?

artiomchi:
Well, yes, and I can kind of see the point in that. But at the same time, why would you ever want to add an unconstrained regex parameter?

First of all, as I provided in the example below - it ends up very error prone, as for years before that, the constraints were matching against the whole value, and this change is unexpected (and undocumented!)

Secondly, from previously being regular expression parameter match, this ends up _feeling_ like a regular expression "contains" match (if you know what I mean).

If the regex match enforced ^ and $, you're saying it would be impossible for the user to remove them, but that's not actually the case. If you want to create a rule that starts with abc you just use `abc.*` (which would be converted to `^abc.*$`. So that's not actually a problem.

The biggest issue (as I see it) is performance - ASP.NET core is promoted as a very **fast** engine, and a `^abc` regex will be intrinsically faster than `^abc.*`, which I completely understand, but even so, on the off case where you'd want a rule like that, I'd rather have a `partialregex` inline rule and a `regex` rule that enforced start and end of string characters.

Again, the docs at https://docs.asp.net/en/latest/fundamentals/routing.html#route-constraint-reference (and a myriad other places!!) provide examples like:

    {ssn:regex(d{3}-d{2}-d{4})}

and

    {action:regex(list|get|create)}

And these examples are just plain wrong, and will get people in massive trouble. The only reason I found out about them is because I like tinkering and was experimenting with the rules...

Ultimately, I still think that changing the `RegexInlineRouteConstraint` is a good idea, even if it is a breaking change for a small set of apps (and will most likely fix most people's routes that they didn't even know were broken)

artiomchi:
Of course I have no data to substantiate it, but I'd imagine that the very vast majority of rules using the inline regex constraint would in fact need to match against the whole value, and only a small subset of regex constraints would be doing a partial match.

artiomchi:
One last note - the benefit of this pull request is it doesn't compromise performance in the slightest, while at the same time ensuring that by default the users will get a safe match (consistent with the examples in all the docs and demos), and in the off chance where you want to do a partial match, you can do that as well by using the `partialregex()` inline constraint that would give you both the functionality you need and the same performance as the `regex` one

artiomchi:
Regarding @Eilon's note about not being able to make a breaking change - I understand the desire not to have breaking changes, but it's not like this has never happened before..

Software and framework releases often have changes in behaviour, and new versions come with release notes that include breaking changes. The main thing we have to make sure is that there is no regression in functionality (i.e. everything that was possible to do before the change should also be possible to do after the change), which in this case is not a problem.

Other than that - I think it's a very beneficial change (even if I know it might be a biased opinion)

Eilon:
@artiomchi unfortunately that is not compatible with the general principles of breaking changes. Breaking changes can be behavioral, compilation, or other. Behavioral breaking changes are often the most severe kind of breaking change because an app developer might not even notice it. The app might *appear* to keep working, but there is in fact a significant change in behavior that is not compatible.

In order to continue with this change we need to preserve the current behavior, and then at most add a new inline constraint that has new behavior.

artiomchi:
Welp, I can see your point..

I guess I'll just make a pull request to patch the docs instead then :)

PR #349, opened 2016-08-04
StringRouteConstraint implementation

iscifoni:
related aspnet/MVC#4846

@rynowak @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Constraints/StringRouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/Constraints/StringRouteConstraint.cs
@@ -8,7 +8,7 @@ using Microsoft.AspNetCore.Http;
 namespace Microsoft.AspNetCore.Routing.Constraints
 {
     /// <summary>
-    /// Constrains a route parameter to contain only a specifyed strign.
+    /// Constrains a route parameter to contain only a specified strign.
     /// </summary>
     public class StringRouteConstraint : IRouteConstraint
     {
--- a/src/Microsoft.AspNetCore.Routing/Constraints/StringRouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/Constraints/StringRouteConstraint.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Constraints
+{
+    /// <summary>
+    /// Constrains a route parameter to contain only a specifyed strign.
+    /// </summary>
+    public class StringRouteConstraint : IRouteConstraint
+    {
+        private string _value;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StringRouteConstraint"/> class.
+        /// </summary>
+        /// <param name="value">The constraint value to match.</param>
+        public StringRouteConstraint(string value)
+        {
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            _value = value;
+        }
+
+        /// <inheritdoc />
+        public bool Match(HttpContext httpContext, IRouter route, string routeKey, RouteValueDictionary values, RouteDirection routeDirection)
+        {
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            if (route == null)
+            {
+                throw new ArgumentNullException(nameof(route));
+            }
+
+            if (routeKey == null)
+            {
+                throw new ArgumentNullException(nameof(routeKey));
+            }
+
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            object routeValue;
+
+            if (values.TryGetValue(routeKey, out routeValue)
+                && routeValue != null)
+            {
+                var parameterValueString = Convert.ToString(routeValue, CultureInfo.InvariantCulture);
+
+                return parameterValueString.Equals(_value, StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
+        }
+    }
+}
--- a/test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNetCore.Routing/Constraints/StringRouteConstraint.cs
```diff
@@ -0,0 +1,67 @@
...
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            object routeValue;
+
+            if (values.TryGetValue(routeKey, out routeValue)
+                && routeValue != null)
+            {
+                var parameterValueString = Convert.ToString(routeValue, CultureInfo.InvariantCulture);
+
```
iscifoni: the if statement could be modified

```
            var routeValue = Convert.ToString(values[routeKey], CultureInfo.InvariantCulture);

            if (routeValue != null)
            {
                return routeValue.Equals(_value, StringComparison.OrdinalIgnoreCase);
            }
```
because the RouteValueDictionary makes the control of the key presence, but did not do it for consistency with the other constraints.

src/Microsoft.AspNetCore.Routing/Constraints/StringRouteConstraint.cs
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Constraints
+{
+    /// <summary>
+    /// Constrains a route parameter to contain only a specifyed strign.
```
rynowak: typo: `specified`

src/Microsoft.AspNetCore.Routing/Constraints/StringRouteConstraint.cs
rynowak: I think it's best the way it is in the other constraints :+1: 

The one gotcha here is that `Convert.ToString((object)null)` will return empty string when provided with null. So it would appear (to someone debugging) that `routeValue` was provided as the empty string when in reality it wasn't present.

test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing.Constraints;
+using Moq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing.Tests.Constraints
```
rynowak: should be the same as the product namespace `Microsoft.AspNetCore.Routing.Constraints`

test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    public class StringRouteConstraintTest
+    {
+        [Fact]
+        public void StringRouteConstraintSimpleTrueTest()
+        {
+            var constraint = new StringRouteConstraint("home");
```
rynowak: Add `// Arrange` before this line (for all of these tests)

test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            var values = new RouteValueDictionary(new { controller = "home" });
```
rynowak: Add `// Act` before this line (for all of these tests)

test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
```diff
@@ -0,0 +1,67 @@
...
+              routeKey: "controller",
+              values: values,
+              routeDirection: RouteDirection.IncomingRequest);
+
+            // Assert
+            Assert.True(match);
+        }
+
+        [Fact]
+        public void StringRouteConstraintKeyNotFoundTest()
```
rynowak: We should also cover the case where the value is in the dictionary but is `null`

test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
```diff
@@ -0,0 +1,67 @@
...
+                httpContext: Mock.Of<HttpContext>(),
+                route: new Mock<IRouter>().Object,
+                routeKey: "action",
+                values: values,
+                routeDirection: RouteDirection.IncomingRequest);
+
+            // Assert
+            Assert.False(match);
+        }
+
```
rynowak: It would be good to see a positive and negative test with `RouteDirection` is `UrlGeneration`. I know the code doesn't do anything specific with it right now, but it would be good to have for completeness

rynowak:
A few small comments, overall looks good :+1:

iscifoni:
:up: :date:

test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
iscifoni: Add 2 new test using RouteDirection.UrlGeneration

test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
iscifoni: Add new inlinedata with routevalue null to test StringRouteConstraintEscapingCaseSensitive and renamed to StringRouteConstraintEscapingCaseSensitiveAndRouteNullTest

test/Microsoft.AspNetCore.Routing.Tests/Constraints/StringRouteConstraintTest.cs
```diff
@@ -0,0 +1,117 @@
...
+                httpContext: Mock.Of<HttpContext>(),
+                route: new Mock<IRouter>().Object,
+                routeKey: "action",
+                values: values,
+                routeDirection: RouteDirection.UrlGeneration);
+
+            // Assert
+            Assert.False(match);
+        }
+
```
rynowak: We're also missing cases where the route-value does have a value, but the value doesn't match

iscifoni:
:up: :date:

rynowak:
:shipit: - looks good, I'll merge

rynowak:
88de3d50703dd9f697df9fd4654f2a1157bcd3a3

rynowak:
FYI - build 21795 and later will have this change

iscifoni:
Ok thanks

PR #354, opened 2016-08-17
[Fixes #346] Port fix for "Request not matching route with defaults" to 1.0.1

### LINKED ISSUE #346: Port fix for "Request not matching route with defaults" to 1.0.1
Original bug for 1.1.0: https://github.com/aspnet/Routing/issues/324
Original PR: for 1.1.0: https://github.com/aspnet/Routing/issues/324

javiercn:


rynowak:
:shipit: - make sure to do the no-op merge back to dev so we don't get conflicts.

PR #358, opened 2016-08-29
[Fixes #351] Error message incorrectly references RouteCollection whe…

### LINKED ISSUE #351: Error message incorrectly references RouteCollection when you don't set the default handler on RouteBuilder
If you don't set the default handler on RouteBuilder and then call MapRoute then you get the following error message:

A default handler must be set on the RouteCollection.

I think that should say RouteBuilder, not RouteCollection.

kichalla:
…n you don't set the default handler on RouteBuilder

@rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Resources.resx
+++ b/src/Microsoft.AspNetCore.Routing/Resources.resx
@@ -130,7 +130,7 @@
     <value>The supplied route name '{0}' is ambiguous and matched more than one route.</value>
   </data>
   <data name="DefaultHandler_MustBeSet" xml:space="preserve">
-    <value>A default handler must be set on the RouteCollection.</value>
+    <value>A default handler must be set on the {0}.</value>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Routing/Resources.resx
```diff
@@ -130,7 +130,7 @@
     <value>The supplied route name '{0}' is ambiguous and matched more than one route.</value>
   </data>
   <data name="DefaultHandler_MustBeSet" xml:space="preserve">
-    <value>A default handler must be set on the RouteCollection.</value>
+    <value>A default handler must be set on the RouteBuilder.</value>
```
pranavkm: Could you pass this instead? Type names can be passed in to since they don't get localized.

kichalla:
:up: :date:

pranavkm:
:shipit:

PR #361, opened 2016-09-07
Use TaskCache class from Microsoft.Extensions.TaskCache.Sources

pakrym:
Instead of Task.FromResult(0)

@pranavkm


test/Microsoft.AspNetCore.Routing.Tests/RoutingBuilderExtensionsTest.cs
```diff
@@ -6,8 +6,8 @@
 using Microsoft.AspNetCore.Builder.Internal;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
-using Microsoft.AspNetCore.Routing.Internal;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Internal;
```
pranavkm: Could we use `Task.FromResult` in tests? It avoids type conflict issues if we ever add an `InternalsVisibleTo` in the source project.

pranavkm:
:watch: for the question.

pakrym:
Switched tests to use `Task.FromResult()`

pranavkm:
:shipit:

Issue #357, opened 2016-08-25
Fix behavior when optional parameters are defined within a route and not at the end


javiercn:
https://github.com/aspnet/Routing/issues/326

We create the incorrect tree. We should remove the optionality for route matching when we create the attribute route tree.

javiercn:
After doing some further investigation the E2E behavior is expected here. The optional parameter will be ignored during route matching. And for route generation it will only be considered as a default value. So for example, given the template

```/a/b/{c=3}/d```
It will only match routes like
```/a/b/c/d```
```/a/b/e/d```

and at link generation time, it will generate the route
```/a/b/3/d```

The reason for this to happen is that while TreeRouteBuilder creates the incorrect tree, TemplateRouteMatcher actually checks all the segments in the template and fails to match requests like /a/b when it gets to the d segment and determines that its not optional.

Given that the current behavior is what we expected the fix for this bug to be, we don't have to do anything here other than add the relevant tests to document the behavior here.

javiercn:
@rynowak @Eilon

rynowak:
I don't get what you're saying - are you saying that the behavior in 1.0.0 is correct or incorrect?

Eilon:
I think this bug just needs to be closed, no?

javiercn:
@rynowak still wants to fix the behavior on treeroutebuilder.

rynowak:
We build the wrong data structure, and then rely on the implementation details of another part of the code to bail us out.

Eilon:
Well that sounds just dandy 😄 Indeed, let's clean that up... And we're sure we have enough test coverage for this now?

PR #365, opened 2016-09-26
Fixes [#357] Fix behavior when optional parameters are defined within a route and not at the end

javiercn:


src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -134,6 +133,8 @@ public class TreeRouter : IRouter
         /// </summary>
         public int Version { get; }
 
+        public IEnumerable<UrlMatchingTree> MatchingTrees => _trees;
```
javiercn: probably want to move this to internal

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
```diff
@@ -333,7 +333,8 @@ private void AddEntryToTree(UrlMatchingTree tree, InboundRouteEntry entry)
                     continue;
                 }
 
-                if (part.IsParameter && (part.IsOptional || part.IsCatchAll || part.DefaultValue != null))
+                if (part.IsParameter && IsOptionalOrCatchAllOrHasDefaultValue(part) &&
+                    RemainingSegmentsAreOptional(entry.RouteTemplate.Segments, i))
```
rynowak: Formatting could be better here

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
```diff
@@ -392,5 +393,25 @@ private void AddEntryToTree(UrlMatchingTree tree, InboundRouteEntry entry)
                 return result == 0 ? x.Entry.RouteTemplate.TemplateText.CompareTo(y.Entry.RouteTemplate.TemplateText) : result;
             });
         }
+
+        private static bool RemainingSegmentsAreOptional(IList<TemplateSegment> segments, int currentParameterIndex)
+        {
+            for (var j = currentParameterIndex + 1; j < segments.Count; j++)
```
rynowak: it's a little more customary to use `i` here :laughing:

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
rynowak: You could also just start your loop at `j` instead of `j+1` and then you don't need the check for `IsOptionalOrCatchAllOrHasDefaultValue` here

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
```diff
@@ -333,7 +333,8 @@ private void AddEntryToTree(UrlMatchingTree tree, InboundRouteEntry entry)
                     continue;
                 }
 
-                if (part.IsParameter && (part.IsOptional || part.IsCatchAll || part.DefaultValue != null))
+                if (part.IsParameter && IsOptionalOrCatchAllOrHasDefaultValue(part) &&
```
rynowak: This could use an example or two in the comments

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
```diff
@@ -392,5 +393,25 @@ private void AddEntryToTree(UrlMatchingTree tree, InboundRouteEntry entry)
                 return result == 0 ? x.Entry.RouteTemplate.TemplateText.CompareTo(y.Entry.RouteTemplate.TemplateText) : result;
             });
         }
+
+        private static bool RemainingSegmentsAreOptional(IList<TemplateSegment> segments, int currentParameterIndex)
+        {
+            for (var j = currentParameterIndex + 1; j < segments.Count; j++)
+            {
+                if (!(segments[j].IsSimple &&
```
rynowak: weird negation - this consider just breaking this up into a series of `if`s with some comments

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
rynowak: Yes please

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouteBuilderTest.cs
```diff
@@ -70,19 +71,160 @@ public void TreeRouter_BuildDoesNotThrow_RoutesWithTheSameNameAndSameTemplates()
...
+        {
+            var resolverMock = new Mock<IInlineConstraintResolver>();
+            resolverMock.Setup(o => o.ResolveConstraint("int")).Returns(new IntRouteConstraint());
+            resolverMock.Setup(o => o.ResolveConstraint("range(1,20)")).Returns(new RangeRouteConstraint(1, 20));
+            resolverMock.Setup(o => o.ResolveConstraint("alpha")).Returns(new AlphaRouteConstraint());
+            resolverMock.Setup(o => o.ResolveConstraint(@"regex(^\d{3}-\d{3}-\d{4}$)")).Returns(
+                new RegexInlineRouteConstraint(@"^\d{3}-\d{3}-\d{4}$"));
+            resolverMock.Setup(o => o.ResolveConstraint(@"regex(^\d{1,2}\/\d{1,2}\/\d{4}$)")).Returns(
+                new RegexInlineRouteConstraint(@"^\d{1,2}\/\d{1,2}\/\d{4}$"));
+            return resolverMock.Object;
```
rynowak: Do we need all this? Why not just use the 'real one'?

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouteBuilderTest.cs
```diff
@@ -70,19 +71,160 @@ public void TreeRouter_BuildDoesNotThrow_RoutesWithTheSameNameAndSameTemplates()
             builder.Build();
         }
 
+        [Fact]
+        public void TreeRouter_BuildDoesNotAddIntermediateMatchingNodes_ForRoutesWithIntermediateParametersWithDefaultValues()
```
rynowak: We need some tests with an arity `!= 3`

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
```diff
@@ -275,6 +274,27 @@ public class TreeRouterTest
         }
 
         [Fact]
+        public async Task TreeRouter_RouteAsync_DoesNotMatchRoutesWithIntermediateDefaultRouteValues()
```
rynowak: Would like to see a few more cases longer/shorter or combinations of default value + optional or default value + catchall

javiercn:
Updated

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
```diff
@@ -392,5 +398,36 @@ private void AddEntryToTree(UrlMatchingTree tree, InboundRouteEntry entry)
...
+                {
+                    // /{parameter}
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private static bool IsOptionalOrCatchAllOrHasDefaultValue(TemplatePart part)
```
rynowak: This could just be merged into the above. I know you stylistically **love** long method names, but this isn't abstracting anything that a local variable doesn't do.

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouteBuilderTest.cs
rynowak: Still waiting for a reply

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
javiercn: Fair enough, this method made sense when the factoring of the code was different. I don't think it hurts, but I'll make you happy. This is something that could also bubble to the RouteSegment class in a different flavor, like IsRequired or something like that.

javiercn:
Addressed the remaining feedback, replaced the constraint resolver mock with the default implementation, and removed the helper method.

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouteBuilderTest.cs
javiercn: done

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
javiercn: done

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
javiercn: done

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
javiercn: done

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
javiercn: done

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
javiercn: done

rynowak:
:shipit:

PR #366, opened 2016-09-27
[Fixes #359] Routing is matching empty segments to parameters and defaults are wrong

### LINKED ISSUE #359: Routing is matching empty segments to parameters and defaults are wrong
Consider the following:
```
    public class ValuesController : Controller
    {
        [HttpGet("{a=a}/{b=b}/{c=c}")]
        public string Get(string a, string b, string c)
        {
            return $"{a} - {b} - {c}";
        }
    }
```

This should match a url like `/` or `/10` or `/10/20` or `/10/20/30` 

However with a URL like `///hi` this will match. And the output will be `b - c - hi`. That is, the parameter `a` gets the default value `b` and is able to match an empty string.

javiercn:


src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
```diff
@@ -72,7 +72,6 @@ public bool TryMatch(PathString path, RouteValueDictionary values)
 
             var i = 0;
             var pathTokenizer = new PathTokenizer(path);
-
```
dougbu: Suggest leaving blank lines before comments alone.

src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
```diff
@@ -82,20 +81,25 @@ public bool TryMatch(PathString path, RouteValueDictionary values)
             //
             // On the second pass, we're almost certainly going to match the URL, so go ahead and allocate the 'values'
             // and start capturing strings. 
-            foreach (var requestSegment in pathTokenizer)
+            foreach (var pathSegment in pathTokenizer)
```
dougbu: Why the rename? `requestSegment` seems less ambiguous w.r.t. template versus request.

src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
```diff
@@ -82,20 +81,25 @@ public bool TryMatch(PathString path, RouteValueDictionary values)
...
             {
+                if (pathSegment.Length == 0)
+                {
+                    return false;
+                }
```
rynowak: Is this the only substantive change?

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateMatcherTests.cs
```diff
@@ -991,6 +991,107 @@ public void TryMatch_MultipleOptionalParameters()
...
         }
 
+        [Theory]
+        [InlineData("///")]
+        [InlineData("/a//")]
+        [InlineData("/a/b//")]
+        [InlineData("//b//")]
+        [InlineData("///c")]
+        [InlineData("///c/")]
+        public void TryMatch_MultipleOptionalParameters_WithEmptyIntermediateSegmentsDoesNotMatch(string url)
```
rynowak: I'd like to see a few tests at a higher level than `TemplateMatcher` - ie tests that cover `Route` and attribute routing

src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
javiercn: Yep. It was a matter of not taking into account pathSegment.Length == 0 which can only happen at the start or in the middle of a request path if it contains any number of consecutive //. For trailing `/` the PathTokenizer will return false

javiercn:
Addressed feedback

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateMatcherTests.cs
javiercn: Added tests for attribute routing, I think we don't need tests for route as it's pretty straightforward what we do in the RouteBase implementation. If you think otherwise, I think the right thing would be to create a bunch of tests that all route implementations can reuse instead of having the same kind of tests duplicated across the code base, but I think we are covered for this bug

src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
javiercn: We usually refer to path vs route across the routing code base, requestSegment and routeSegment are actually pretty similar and it was causing me trouble to follow through the code, so I went and made it clearer

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateMatcherTests.cs
rynowak: I do think otherwise, which is why I specifically mentioned both. We don't need a test for every possible scenario, just a few.

PR #368, opened 2016-10-23
Add DoesNotContainConstraint to Routing Constraings

ilya-git:
This constraint is very useful for the modern SPA applications.  For example if we want a SPA page for all request apart from static files and "/api" route we can use the following snippet:

```
app.UseMvc(routes =>
{
    routes.MapRoute(
        name: "api",
        template: "api/{controller}/{action}");

    routes.MapRoute(
        name: "angular",
        template: "{*url}",
        defaults: new {controller = "Home", action = "Index"},
        constraints: new {url = new DoesNotContainConstraint(".", "api/") });                
});
```

If we don't have such a constraint it will result in all requests to non-existing files (like e.g. .map.js files that Chrome does, etc...) and to non-existing api calls (mistype, etc...) result in the home page of the application returned which is not quite ideal situation. RegexConstraint can possibly be used, but it's too complex and most likely slower for such a simple and commonly required case.

### DIFF
--- a/src/Microsoft.AspNet.Routing/Constraints/DoesNotContainConstraint.cs
+++ b/src/Microsoft.AspNet.Routing/Constraints/DoesNotContainConstraint.cs
@@ -1,0 +1,44 @@
+public class DoesNotContainConstraint : IRouteConstraint
+    {
+        private readonly string m_Substring;
+
+        public DoesNotContainConstraint(string substring)
+        {
+            m_Substring = substring;
+        }
+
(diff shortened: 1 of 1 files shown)


Eilon:
@mkArtakMSFT / @rynowak  - any further action planned on this PR?

natemcmaster:
Hi @ilya-git, it appears this PR has gone stale. We're closing because we are archiving this repo. If this is something you would still like to contribute, please open a new PR against https://github.com/aspnet/AspNetCore.

PR #371, opened 2016-12-12
[Design][Fixes #367] Add extensions on WebHostBuilder for super simple HTTP s…

### LINKED ISSUE #367: Add extensions for super simple HTTP service application building
Some people want to have as little code as possible in a single file (`Program.cs`) to stand up very simple HTTP services (no, I won't say "micro-services").

In other platforms (Go, Node, etc.) it's generally very easy to get an HTTP server up with some basic dispatching to various request-handling functions based on URL path. Doing this in ASP.NET Core today requires slightly more ceremony (Using `IApplicationBuilder`, `RouteBuilder`, a `Startup` class, etc.) but this can be very easily fixed with a couple of new extension methods in Routing.

Super simple HTTP service application:

``` c#
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;

namespace NotAMicroService
{
    public class Program
    {
        public static void Main(string[] args)
        {
            var host = new WebHostBuilder()
                .UseKestrel()
                .UseUrls("http://+:8080/") // optional, we default to listening on http://localhost:5000/
                .Configure(app => app
                    .UseRouter(router => router
                        .MapGet("hello/{*name}", (req, res, data) => res.WriteAsync($"Hello, {data.Values["name"]}!"))
                        .MapGet("goodbye/{*name}", (req, res, data) => res.WriteAsync($"Goodbye, {data.Values["name"]}!"))
                        .MapGet("{*name}", (req, res, data) => res.WriteAsync($"Welcome, {data.Values["name"]}!"))
                    )
                )
                .Build();

            host.Run();
        }
    }
}
```

Extension methods we need to add to `Microsoft.AspNetCore.Routing` to make this work:

``` c#
using System;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;

namespace Microsoft.AspNetCore.Hosting
{
    public static class WebHostExtensions
    {
        public static IWebHostBuilder UseRouter(this IWebHostBuilder builder, Action<RouteBuilder> routeBuilder)
        {
            return builder
                .ConfigureServices(services => services.AddRouting())
                .Configure(app =>
                {
                    var router = new RouteBuilder(app);
                    routeBuilder(router);
                    app.UseRouter(router.Build());
                });
        }
    }
}
```

``` c#
using System;
using System.Threading.Tasks;

namespace Microsoft.AspNetCore.Hosting
{
    public static class RouteBuilderExtensions
    {
        public static IRouteBuilder MapGet(this IRouteBuilder builder, string template, Func<HttpRequest, HttpResponse, RouteData, Task> handler)
        {
            builder.MapGet(template, context =>
            {
                return handler(context.Request, context.Response, context.GetRouteData());
            });
            return builder;
        }
    }
}
```

kichalla:
…ervice application building

@DamianEdwards @rynowak


src/Microsoft.AspNetCore.Routing/WebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public static class WebHostExtensions
+    {
+        /// <summary>
+        /// Adds routing related services and a <see cref="RouterMiddleware"/> middleware.
+        /// </summary>
+        /// <param name="builder">The <see cref="IWebHostBuilder"/>.</param>
+        /// <param name="configureRoutes">An <see cref="Action{RouteBuilder}"/> to configure the provided <see cref="RouteBuilder"/>.</param>
+        /// <returns></returns>
+        public static IWebHostBuilder UseRouter(this IWebHostBuilder builder, Action<RouteBuilder> configureRoutes)
```
davidfowl: Ugh, I thought we agree this should hang off `IApplicationBuilder`.

src/Microsoft.AspNetCore.Routing/project.json
```diff
@@ -20,6 +20,7 @@
     "xmlDoc": true
   },
   "dependencies": {
+    "Microsoft.AspNetCore.Hosting": "1.2.0-*",
```
davidfowl: You shouldn't need this because we should be running off `IApplicationBuilder`

kichalla: Sure, will update the PR.

kichalla: @davidfowl There is an already an existing extension on IApplicationBuilder which can be used. I updated the PR.

samples/RoutingSample.Web/TestHttpServices.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace RoutingSample.Web
+{
+    public class TestHttpServices
+    {
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                .UseKestrel()
+                .ConfigureServices(services => services.AddRouting())
```
kichalla: @davidfowl example of how it looks like now

samples/RoutingSample.Web/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+                .Configure(app => app.UseRouter(routes =>
+                        {
+                            routes.DefaultHandler = new RouteHandler((httpContext) =>
+                            {
+                                var request = httpContext.Request;
+                                return httpContext.Response.WriteAsync($"Verb =  {request.Method.ToUpperInvariant()} - Path = {request.Path} - Route values - {string.Join(", ", httpContext.GetRouteData().Values)}");
+                            });
+
+                            routes.MapGet("api/get/{id}", (request, response, routeData) => response.WriteAsync($"API Get {routeData.Values["id"]}"))
+                                  .MapRoute("api/middleware", (IApplicationBuilder fork) => fork.Use((httpContext, next) => httpContext.Response.WriteAsync("Middleware!")))
```
rynowak: Had some discussion with @DamianEdwards and @Eilon lipton about these overloads (the ones that take `IApplicationBuilder`). We should **rename** these for 2.0.0 which would be a breaking change.

The problem is that lambda type inference can't work well because there are two overloads that take a single-parameter lambda.

Might want to double check with @Eilon - that could be either part of this PR or not (your choice).

rynowak: Suggestions for names, I'm pitching `MapMiddleware<Verb>`

src/Microsoft.AspNetCore.Routing/RequestDelegateRouteBuilderExtensions.cs
```diff
@@ -75,6 +76,22 @@ public static IRouteBuilder MapDelete(this IRouteBuilder builder, string templat
...
+        /// <paramref name="template"/>, and <paramref name="handler"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IRouteBuilder"/>.</param>
+        /// <param name="template">The route template.</param>
+        /// <param name="handler">The route handler.</param>
+        /// <returns>A reference to the <paramref name="builder"/> after this operation has completed.</returns>
+        public static IRouteBuilder MapDelete(
+            this IRouteBuilder builder,
+            string template,
+            Func<HttpRequest, HttpResponse, RouteData, Task> handler)
```
rynowak: @DamianEdwards - should this be `RouteData` or `RouteValueDictionary`?

samples/RoutingSample.Web/Program.cs
Eilon: If we agree we want to do a full "cleansing" of this group of APIs, I'm fine doing it all in one (but don't care either way).

src/Microsoft.AspNetCore.Routing/RequestDelegateRouteBuilderExtensions.cs
Eilon: Good question. You certainly don't want the DataTokens (there aren't any) or the Routers (super-not-interesting here). So perhaps RVD is all that's even useful anyway?

src/Microsoft.AspNetCore.Routing/RequestDelegateRouteBuilderExtensions.cs
```diff
@@ -88,16 +145,52 @@ public static IRouteBuilder MapGet(this IRouteBuilder builder, string template,
...
+        /// <para>
+        ///     Adds a route to the <see cref="IRouteBuilder"/> that only matches HTTP GET requests for the given
+        ///     <paramref name="template"/>, and <paramref name="action"/>.
+        /// </para>
+        /// </summary>
+        /// <param name="builder">The <see cref="IRouteBuilder"/>.</param>
+        /// <param name="template">The route template.</param>
+        /// <param name="action">The action to apply to the <see cref="IApplicationBuilder"/>.</param>
+        /// <returns>A reference to the <paramref name="builder"/> after this operation has completed.</returns>
+        [Obsolete("This method is obsolete and will be removed in a future version. The recommended alternative is Microsoft.AspNetCore.Routing.RequestDelegateRouteBuilderExtensions.MapMiddlewareGet.")]
```
Eilon: Per discussion, we should probably get rid of these methods entirely to avoid the IntelliSense issue where the overloads are ambiguous at first.

Need to post a breaking-change announcements for this too.

cc @rynowak

kichalla:
🆙 📅

src/Microsoft.AspNetCore.Routing/RequestDelegateRouteBuilderExtensions.cs
```diff
@@ -41,7 +42,7 @@ public static IRouteBuilder MapRoute(this IRouteBuilder builder, string template
         /// <param name="template">The route template.</param>
         /// <param name="action">The action to apply to the <see cref="IApplicationBuilder"/>.</param>
         /// <returns>A reference to the <paramref name="builder"/> after this operation has completed.</returns>
-        public static IRouteBuilder MapRoute(this IRouteBuilder builder, string template, Action<IApplicationBuilder> action)
+        public static IRouteBuilder MapMiddleware(this IRouteBuilder builder, string template, Action<IApplicationBuilder> action)
```
rynowak: This should probably be `MapMiddlewareRoute`

test/Microsoft.AspNetCore.Routing.FunctionalTests/RoutingSampleTest.cs
```diff
@@ -1,30 +1,36 @@
...
+        private readonly HttpClient _client;
+        private readonly TestServer _testServer;
+
+        public RoutingSampleTest()
         {
-            Client = fixture.Client;
+            var webHostBuilder = RoutingSample.Web.Program.GetWebHostBuilder();
+            _testServer = new TestServer(webHostBuilder);
+            _client = _testServer.CreateClient();
+            _client.BaseAddress = new Uri("http://localhost");
```
rynowak: Why?

PR #375, opened 2016-12-30
[Fixes #370] Raw route values should be restored after template binder failing binding values when generating a url

### LINKED ISSUE #370: [Routing]Raw route values should be restored after template binder failing binding values when generating a url
_From @anobaka on December 6, 2016 7:20_

### Minimal repro steps
1. Create a sample .net core web application.
2. Modify the `app.UseMvc...` line to:
```
app.UseMvc(routes =>
{
    routes.MapRoute(
        name: "test",
        template: "{area?}/{controller=Home}/{action=Index}/{id?}"
    );
    routes.MapRoute(
        name: "default",
        template: "{controller=Home}/{action=Index}/{id?}");
});
```
3. Add `TestController` and an `Index` action in it.
4. Add `var a = Url.Action("Index", "Test");` in `Index` Action of `HomeController`.
5. Run app and visit `/`.

### Expected result
local variable `a` should be `/Test`

### Actual result
the value of `a` is `/`

### Further technical details
I checked the source of `Routing` and found that the raw route values will not be restored after template binder failed binding values. So when the `default` router was tring to generate url, there was `action:Index` only left in route values, so it used the default controller `Home`.
ref:
[https://github.com/aspnet/Routing/blob/rel/1.1.0/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs#L238](https://github.com/aspnet/Routing/blob/rel/1.1.0/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs#L238)
[https://github.com/aspnet/Routing/blob/rel/1.1.0/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs#L279](https://github.com/aspnet/Routing/blob/rel/1.1.0/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs#L279)
I think the logic is that all routers should try to bind raw route values to generate url, but not the route values modified by previous router, so the raw route values should be restored after each router binding or routers should use the copy of the raw route values.

Thanks.

_Copied from original issue: aspnet/Home#1860_

kichalla:
@rynowak

### DIFF
--- a/test/Microsoft.AspNetCore.Routing.Tests/RouteCollectionTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/RouteCollectionTest.cs
@@ -363,76 +363,58 @@ namespace Microsoft.AspNetCore.Routing
             Assert.Empty(pathData.DataTokens);
         }
 
-        public static TheoryData<RestoreRouteDataVariation> RestoresRouteDataForEachRouterData
+        public static IEnumerable<object[]> RestoresRouteDataForEachRouterData
         {
             get
             {
-                var data = new TheoryData<RestoreRouteDataVariation>();
-
                 // Here 'area' segment doesn't have a value but the later segments have values. This is an invalid
                 // route match and the url generation should look into the next available route in the collection.
-                var variation = new RestoreRouteDataVariation();
-                variation.Routes.Add(new RouteInfo("1", "{area?}/{controller=Home}/{action=Index}/{id?}"));
-                variation.Routes.Add(new RouteInfo("2", "{controller=Home}/{action=Index}/{id?}"));
-                variation.Values = new RouteValueDictionary(new { controller = "Test", action = "Index" });
-                variation.ExpectedUrl = "/Test";
-                variation.ExpectedRouteToMatch = "2";
-                data.Add(variation);
-
-                variation = new RestoreRouteDataVariation();
-                variation.Routes.Add(new RouteInfo("1", "{category:int}/{controller=Home}/{action=Index}/{id?}"));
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Routing.Tests/RouteCollectionTest.cs
```diff
@@ -363,6 +363,79 @@ public static IEnumerable<object[]> IntegrationTestData
...
+                variation = new RestoreRouteDataVariation();
+                variation.Routes.Add(new RouteInfo("1", "{a}/{b?}/{c}"));
+                variation.Routes.Add(new RouteInfo("2", "{a=Home}/{b=Index}"));
+                variation.Values = new RouteValueDictionary(new { a = "Test", c = "Foo" });
+                variation.ExpectedUrl = "/Test?c=Foo";
+                variation.ExpectedRouteToMatch = "2";
+                data.Add(variation);
+
+                return data;
+            }
```
rynowak: Why is all this needed? I don't understand why we need to do something so complicated to test this, it should just be two mocks. 

Is there anything in the bug that says anything about particular routes or values?

kichalla: I added a new TemplateBinder test and simplified the existing tests. I didn't remove the existing tests because they are integration tests and more easy to follow through as a scenario than the targeted template binder test.

kichalla:
🆙 📅

kichalla:
4bfd663c451281acc4e78a9a84e105bdb4343800

Issue #373, opened 2016-12-17
Using square brackets in regex constraints


vaindil:
I have a controller that accepts either an ID or a string at a given endpoint. I have the ID function set correctly using `api/departments/{id:long}`, but I cannot figure out how to route to the string endpoint. The `alpha` constraint is not enough as the string can contain numbers.

I'm trying to use this route to accept any valid string but require that the first character be a letter.

```
[Route("api/departments/{deptName:regex(^[a-zA-Z]{1}[a-zA-Z0-9_]*$)}")]
```

This does not work, however:

> For action: 'Namespace.DepartmentsNameController.Get (Namespace)'
Error: While processing template 'api/departments/{deptName:regex(^[a-zA-Z]{1}[a-zA-Z0-9_]*$)}', a replacement value for the token 'a-zA-Z' could not be found. Available tokens: 'action, controller'.

What is the correct way to approach this?

dougbu:
Escape (double) the special characters: `^[[a-zA-Z]]{{1}}[[a-zA-Z0-9_]]*$`. Since matches are not case-sensitive and `{1}` is unnecessary, `^[[a-z]][[a-z0-9_]]*$` works too.

That said, we should improve the message in this case. An un-escaped curly brace leads to an informative message:
```
In a route parameter, '{' and '}' must be escaped with '{{' and '}}'
```
`AttributeRouteModel` currently ignores curly braces when handling square brackets and the
```
An unescaped '[' token is not allowed inside of a replacement token. Use '[[' to escape.
```
message appears only when square brackets appear within square-bracketed tokens.

Nit: The curly-brace message should end with a period 😈

vaindil:
Oh that's awesome. Not sure how I looked at it so many times and didn't see the `{1}` being unnecessary, heh, and I didn't realize even regex matches are not case-sensitive for routing. Thank you so much, especially for helping on a Saturday night!

Would it be helpful to include this in the routing docs? I can open an issue if it would help. I scoured that page to find the info; this issue was my last resort.

Eilon:
Yeah definitely something we can improve.

I think the fixes are:

* [ ] Add a period to the end of the `In a route parameter, '{' and '}'  ...` message
* [ ] For the `While processing template 'foo', a replacement value ...` message, add a blurb about how to escape `[` and `]`. Use something similar (but not identical) to the `An unescaped '[' token` message (because you can't know at this time that it's actually unescaped, so it's just informative, not accusative).

[...]

PR #376, opened 2017-01-03
Fixed error message related to escaping curly braces.

kichalla:
Related to https://github.com/aspnet/Routing/issues/373

### DIFF
--- a/test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateParserTests.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateParserTests.cs
@@ -521,7 +521,7 @@ namespace Microsoft.AspNetCore.Routing.Template.Tests
             // Act and Assert
             ExceptionAssert.Throws<ArgumentException>(
                 () => TemplateParser.Parse(template),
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -554,8 +554,8 @@ public void Parse_ComplexSegment_OptionalParametersSeperatedByPeriod_Invalid(str
             // Act and Assert
             ExceptionAssert.Throws<ArgumentException>(
                 () => TemplateParser.Parse(template),
-                "In the complex segment '"+ template +"',  the optional parameter 'p2' is preceded by an invalid " + 
-                "segment '" + parameter +"'. Only valid literal to precede an optional parameter is a period (.)." + 
+                "In the complex segment '"+ template +"',  the optional parameter 'p2' is preceded by an invalid " +
+                "segment '" + parameter +"'. Only valid literal to precede an optional parameter is a period (.)." +
```
Eilon: Argh, since we're here, can we fix the grammar here too? `**The** only valid literal **that can** precede an...`

PR #382, opened 2017-02-10
Remove makefile.shade and sources.nuspec

natemcmaster:
Remove files that are not needed anymore once https://github.com/aspnet/KoreBuild/pull/166 is merged.


rynowak:
:shipit: - this does what you say it does 👍

PR #395, opened 2017-04-10
Make TreeRouter restore RouteData snapshots consistently. Fixes #394.

### LINKED ISSUE #394: ArgumentException: An element with the key '(key)' already exists in the RouteValueDictionary.
Hello, I'm trying to use attribute routing and either things are broken or I'm not getting it. Here is a sample controller

```
namespace WebApplication1.Controllers
{
    [Route("/api")]
    public class TestController : Controller
    {
        [HttpPost("environment('{environment:alpha}')/copy('{newEnvironment:alpha}')")]
        public bool Clone([FromRoute(Name = "environment")]string env, [FromRoute(Name = "newEnvironment")] string env2)
        {
            return true;
        }


        [HttpGet("environment('{environment:alpha}')/promotions('{name:alpha}')")]
        public bool GetPromotion([FromRoute(Name = "environment")] string env, [FromRoute(Name = "name")] string name)
        {
            return true;
        }
    }
}
```

If I make a request to any endpoint, I see this error:
```
System.ArgumentException: An element with the key 'environment' already exists in the RouteValueDictionary.
Parameter name: key
   at Microsoft.AspNetCore.Routing.RouteValueDictionary.Add(String key, Object value)
   at Microsoft.AspNetCore.Routing.Template.TemplateMatcher.MatchComplexSegmentCore(TemplateSegment routeSegment, String requestSegment, IReadOnlyDictionary`2 defaults, RouteValueDictionary values, Int32 indexOfLastSegmentUsed)
   at Microsoft.AspNetCore.Routing.Template.TemplateMatcher.MatchComplexSegment(TemplateSegment routeSegment, String requestSegment, IReadOnlyDictionary`2 defaults, RouteValueDictionary values)
   at Microsoft.AspNetCore.Routing.Template.TemplateMatcher.TryMatch(PathString path, RouteValueDictionary values)
   at Microsoft.AspNetCore.Routing.Tree.TreeRouter.<RouteAsync>d__13.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Routing.RouteCollection.<RouteAsync>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Builder.RouterMiddleware.<Invoke>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.<Invoke>d__7.MoveNext()
```

SteveSandersonMS:
Fix for #394.

The issue was that `TreeRouter`'s `RouteAsync` logic was not using its own `snapshot.Restore` feature consistently in between attempting each `matcher`. If a `matcher` failed to match because its 2nd or subsequent complex segment didn't match, then any `RouteValueDictionary` entries populated by the N-1 preceding complex segments were left in place and leaked over into subsequent `matcher` attempts. Depending on whether a subsequent `matcher` tried to use the same key names or not, you'd either get a "*key already present*" exception, or would just leak the internal state from preceding matchers into subsequent ones, neither of which are desirable behaviours.

This fix is pretty trivial: it just ensures that `snapshot.Restore()` is called after each non-matching `matcher.TryMatch`.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
+++ b/src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
@@ -192,13 +192,14 @@ namespace Microsoft.AspNetCore.Routing.Tree
                     {
                         var entry = item.Entry;
                         var matcher = item.TemplateMatcher;
-                        if (!matcher.TryMatch(context.HttpContext.Request.Path, context.RouteData.Values))
-                        {
-                            continue;
-                        }
 
                         try
                         {
+                            if (!matcher.TryMatch(context.HttpContext.Request.Path, context.RouteData.Values))
+                            {
+                                continue;
+                            }
+
                             if (!RouteConstraintMatcher.Match(
                                 entry.Constraints,
                                 context.RouteData.Values,
--- a/test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
@@ -1844,6 +1844,43 @@ namespace Microsoft.AspNetCore.Routing.Tree
             Assert.Equal(next.Object.GetType(), nestedRouters[0].GetType());
         }
 
+        [Fact]
+        public async Task TreeRouter_SnapshotsRouteData_ResetsBeforeMatchingEachRouteEntry()
+        {
+            // This test replicates a scenario raised as issue https://github.com/aspnet/Routing/issues/394
+            // The RouteValueDictionary entries populated while matching route entries should not be left
+            // in place if the route entry turns out not to match, because that would leak unwanted state
+            // to subsequent route entries and might cause "An element with the key ... already exists"
+            // exceptions.
+
+            // Arrange
+            RouteValueDictionary nestedValues = null;
+            var next = new Mock<IRouter>();
+            next
+                .Setup(r => r.RouteAsync(It.IsAny<RouteContext>()))
+                .Callback<RouteContext>(c =>
+                {
+                    nestedValues = new RouteValueDictionary(c.RouteData.Values);
+                    c.Handler = NullHandler;
+                })
+                .Returns(Task.FromResult(0));
+
+            var builder = CreateBuilder();
+            MapInboundEntry(builder, "cat_{category1}/prod1_{product}"); // Matches on first segment but not on second
+            MapInboundEntry(builder, "cat_{category2}/prod2_{product}", handler: next.Object);
+            var route = builder.Build();
+
+            var context = CreateRouteContext("/cat_examplecategory/prod2_exampleproduct");
+
+            // Act
+            await route.RouteAsync(context);
+
+            // Assert
+            Assert.Equal("examplecategory", nestedValues["category2"]);
+            Assert.Equal("exampleproduct", nestedValues["product"]);
+            Assert.DoesNotContain(nestedValues, kvp => kvp.Key == "category1");
+        }
+
         private static RouteContext CreateRouteContext(string requestPath)
         {
             var request = new Mock<HttpRequest>(MockBehavior.Strict);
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -192,13 +192,14 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
...
                         var entry = item.Entry;
                         var matcher = item.TemplateMatcher;
-                        if (!matcher.TryMatch(context.HttpContext.Request.Path, context.RouteData.Values))
-                        {
-                            continue;
-                        }
 
                         try
                         {
+                            if (!matcher.TryMatch(context.HttpContext.Request.Path, context.RouteData.Values))
```
Eilon: I don't know this code at all, but even though this specific code change looks correct to me (it's now properly implemented as `try { do; } finally { cleanup; }`), why does `matcher.TryMatch()` throw the offending exception to begin with? Is that also a bug that should be fixed?

I'm worried that this change has the correct behavior, but is perhaps masking some other bug.

SteveSandersonMS: It's not throwing any exception - having it in the `try` block just means that `finally` will run even when it invokes `continue;` on line 200.

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
```diff
@@ -1844,6 +1844,43 @@ public static IEnumerable<object[]> OptionalParamValues
...
+            RouteValueDictionary nestedValues = null;
+            var next = new Mock<IRouter>();
+            next
+                .Setup(r => r.RouteAsync(It.IsAny<RouteContext>()))
+                .Callback<RouteContext>(c =>
+                {
+                    nestedValues = new RouteValueDictionary(c.RouteData.Values);
+                    c.Handler = NullHandler;
+                })
+                .Returns(Task.FromResult(0));
```
pranavkm: `Task.CompletedTask`

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
```diff
@@ -1844,6 +1844,43 @@ public static IEnumerable<object[]> OptionalParamValues
...
+
+            var context = CreateRouteContext("/cat_examplecategory/prod2_exampleproduct");
+
+            // Act
+            await route.RouteAsync(context);
+
+            // Assert
+            Assert.Equal("examplecategory", nestedValues["category2"]);
+            Assert.Equal("exampleproduct", nestedValues["product"]);
+            Assert.DoesNotContain(nestedValues, kvp => kvp.Key == "category1");
```
pranavkm: Could you throw in a `Assert.NoNull(nestedValues);` here? A bit nicer than having the test fail with a null ref.

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
Eilon: OIC. I should learn C# better 😄

rynowak: Yeah - the question of whether or not to do cleanup on exception has multiple answers. I decided to just go ahead and clean up when it throws even though no one will look at the data.

I felt like coding it this way asks fewer questions than the other way, as people are used to seeing the try/finally question.

rynowak:
:shipit:

test/Microsoft.AspNetCore.Routing.Tests/Tree/TreeRouterTest.cs
SteveSandersonMS: Fixed

SteveSandersonMS:
Thanks - merged

PR #402, opened 2017-04-28
Fix aspnet/Mvc#6218

rynowak:
This fixes the case described in the comments in TemplateBinder.

This case is much more common for pages which is why we're only seeing it
now. We've had this issue for all of 1.0.0 in both conventional and
attribute routing.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
@@ -64,6 +64,9 @@ namespace Microsoft.AspNetCore.Routing.Template
             // If the URI had ordered parameters a="1", b="2", c="3" and the new values
             // specified that b="9", then we need to invalidate everything after it. The new
             // values should then be a="1", b="9", c=<no value>.
+            //
+            // We also handle the case where a parameter is optional but has no value - we shouldn't
+            // accept additional parameters that appear *after* that parameter.
             for (var i = 0; i < _template.Parameters.Count; i++)
             {
                 var parameter = _template.Parameters[i];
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Routing.Tests/RouteTest.cs
```diff
@@ -1372,6 +1372,103 @@ public void GetVirtualPath_OptionalParameter_InSimpleSegment()
             Assert.Empty(pathData.DataTokens);
         }
 
+        [Fact]
+        public void GetVirtualPath_TwoOptionalParameters_OneValueFromAmbientValues()
```
ajaybhargavb: Duplicate test. Remove.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -87,6 +90,27 @@ public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteV
                     }
                 }
 
+
+                if (!hasNewParameterValue && 
+                    !hasCurrentParameterValue && 
+                    _defaults != null &&
```
ajaybhargavb: Is there a case when `_defaults` is null but we still want to break?

rynowak: Yeah actually that's a good catch, this should be an or.

rynowak:
> #YOLO

-- @Eilon

test/Microsoft.AspNetCore.Routing.Tests/RouteTest.cs
```diff
@@ -1372,6 +1372,103 @@ public void GetVirtualPath_OptionalParameter_InSimpleSegment()
...
+            var pathData = route.GetVirtualPath(context);
+
+            // Assert
+            Assert.NotNull(pathData);
+            Assert.Equal("/a/15/17", pathData.VirtualPath);
+            Assert.Same(route, pathData.Router);
+            Assert.Empty(pathData.DataTokens);
+        }
+
+
```
javiercn: nit: extra line

PR #404, opened 2017-05-06
Migration

JunTaoLuo:
This is breaking the build. I'm guessing @pranavkm just missed the src projects and this is just an oversight?


JunTaoLuo:
FYI only, will merge now but let me know if keeping the src projects as netstandard1.3 was deliberate.

Issue #407, opened 2017-05-12
Investigate skipped TemplateParser tests


natemcmaster:
There are two tests in TemplateParserTests that are skipped on all tests runs because they fail.

- Parse_ComplexSegment_OptionalParameterFollowingPeriod_LastSegment
- Parse_ComplexSegment_OptionalParametersSeperatedByPeriod_Invalid

SteveSandersonMS:
Both of these were fixed in #408 and are no longer skipped.

rynowak:
Cool beans!

PR #408, opened 2017-05-22
Fix failing tests

juunas11:
Fixes issue #407.

One test was failing because the expectation was wrong as far as I can see. I changed the expected part to non-optional instead of optional as that is how it is specified in the template.
The other test that was failing was the xUnit theory with parameters `"{p1}{p2?}", "{p1}"`. It completely crashed the test runner. I found that if I eliminated the `Debug.Assert` in `TemplateParser`, it would go through. The problem was that the preceding part was not a literal at all. I made the assumption that the only thing that can precede an optional part is a literal (which must be a period).

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Resources.resx
+++ b/src/Microsoft.AspNetCore.Routing/Resources.resx
@@ -187,7 +187,7 @@
     <value>In a route parameter, '{' and '}' must be escaped with '{{' and '}}'.</value>
   </data>
   <data name="TemplateRoute_OptionalParameterCanbBePrecededByPeriod" xml:space="preserve">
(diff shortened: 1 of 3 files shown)


rynowak:
This looks great! I'll take care of merging

rynowak:
BTW in case you ever need to, our `Resources.Designer.cs` files get updated by running `build.cmd /t:resx` at the command line.

juunas11:
Thanks, good to know now lol.

rynowak:
f457c7b9d8ab5fba01d1ca8b2ab17d61c381a115
a18f59ab88f34213c96b571f7a43f4b82477eb13
9a6cb89b8da13266fd589b71a978677628099253

rynowak:
It's not showing as merged because of the rebase, but those are the commits in dev.

PR #409, opened 2017-05-23
Target .NET Standard 2.0

ryanbrandenburg:


build/common.props
```diff
@@ -18,8 +18,7 @@
     <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(InternalAspNetCoreSdkVersion)" PrivateAssets="All" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework' AND '$(OutputType)'=='library'">
-    <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework'">
+    <PackageReference Include="NETStandard.Library.NETFramework" Version="$(NETStandardLibraryNETFrameworkVersion)" />
   </ItemGroup>
-
```
pranavkm: Keep this

samples/RoutingSample.Web/RoutingSample.Web.csproj
```diff
@@ -3,7 +3,8 @@
   <Import Project="..\..\build\common.props" />
```
JunTaoLuo: We shouldn't depend on common in samples. Import dependency.props instead and add the explicity reference to NETStandard.Library.NETFramework

pranavkm: If it already works, why change it?

JunTaoLuo: We don't want samples bringing in everything in common.props such as the sdk, signing, ...

dougbu:
@pranavkm 
> Revive the two API check files 

Why? The projects target only .NET Standard 2.0, never .NET Framework AnyWhere.AnyHow.

pranavkm:
Yup, shouldn't need those.

samples/RoutingSample.Web/RoutingSample.Web.csproj
ryanbrandenburg: This sample is referenced by the FunctionalTests, which complain if it's not strongnamed when I replace common.props with dependencies.props. I vote that we leave it as is to get this PR through and unblock the PR's below me, and we can file a bug to not reference common.props

PR #413, opened 2017-06-09
Remove TreatWarningsAsErrors from common.props

mikeharder:
- Property is automatically set by Internal.AspNetCore.Sdk


rynowak:
:shipit:  thanks!

PR #414, opened 2017-06-26
React to options

HaoK:


test/Microsoft.AspNetCore.Routing.Performance/RoutingBenchmark.cs
```diff
@@ -35,7 +35,7 @@ public RoutingBenchmark()
                 NullLoggerFactory.Instance,
                 UrlEncoder.Default,
                 new DefaultObjectPool<UriBuildingContext>(new UriBuilderContextPooledObjectPolicy(UrlEncoder.Default)),
-                new DefaultInlineConstraintResolver(new OptionsManager<RouteOptions>(Enumerable.Empty<IConfigureOptions<RouteOptions>>())));
+                new DefaultInlineConstraintResolver(new OptionsManager<RouteOptions>(new OptionsFactory<RouteOptions>(Enumerable.Empty<IConfigureOptions<RouteOptions>>(), Enumerable.Empty<IPostConfigureOptions<RouteOptions>>()))));
```
rynowak: Would it be possible to add a default constructor to this? We have a million places in MVC where we've done stuff like this

rynowak: Like **literally** a million

HaoK: Are these all test cases that are bypassing DI?

rynowak: There are literally a million places that MVC constructs an options manager

HaoK: Well taking a look at MVC now

HaoK: It doesn't look THAT bad, WebAPIShim is the only one not using TestOptionsMonitor, which is how things are supposed to be used.

Issue #427, opened 2017-08-29
Add a basic sample that selects a RequestDelegate


rynowak:
Add some code to the sample that will select a `RequestDelegate` and *tunnel* it through some middleware in between.

It's OK to start this with all code in the sample. Do this without copying any routing code into the project. We don't need it to do something *real* just to see the abstractions.

PR #430, opened 2017-08-30
Add abstractions and sample for selecting a RequestDelegate

jbagga:
Addresses #426 and #427 

cc @rynowak

### DIFF
--- a/samples/DispatcherSample/Startup.cs
+++ b/samples/DispatcherSample/Startup.cs
@@ -39,13 +39,13 @@ namespace DispatcherSample
                         }
                 },
                 {
-                    "/taytay",
+                    "/example2",
                     new DispatcherFeature
                         {
-                            Endpoint = new DispatcherEndpoint("taylor"),
+                            Endpoint = new DispatcherEndpoint("example2"),
                             RequestDelegate = async (context) =>
                             {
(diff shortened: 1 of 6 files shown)


samples/DispatcherSample/Startup.cs
```diff
@@ -16,9 +19,61 @@ public void ConfigureServices(IServiceCollection services)
...
+                        {
+                            Endpoint = new DispatcherEndpoint("example"),
+                            RequestDelegate = async (context) =>
+                            {
+                                await context.Response.WriteAsync("Hello from the example!");
+                            }
+                        }
+                },
+                {
+                    "/taytay",
```
NTaylorMullen: 👍

samples/DispatcherSample/Startup.cs
```diff
@@ -16,9 +19,61 @@ public void ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
+            app.Use(async (context, next) =>
+            {
+                await context.Response.WriteAsync("<p>Middleware 1</p>");
+                await next.Invoke();
```
rynowak: How does this work? Doesn't this need a parameter? I'd expect to see `next(context)`

rynowak:
:shipit:

Issue #428, opened 2017-08-29
Add some code to the sample that will 'select' an address and generate a URL


rynowak:
Add some code to the sample that will 'select' an address and generate a URL

It's OK to start this with all code in the sample. Do this without copying any routing code into the project. We don't need it to do something *real* just to see the abstractions.

PR #434, opened 2017-09-01
Add sample to "select" an address and generate a URL

jbagga:
Addresses #428 

Naive logic to select the address from a table. Looks for exact match.

cc @rynowak


samples/DispatcherSample/AddressTable.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Routing;
+
+namespace DispatcherSample
+{
+    public class RouteValueAddressTable
```
rynowak: Add this and the UrlGenerator to services

rynowak:
This looks great. Add the UrlGenerator and the RVAT to services as singletons so that they can be retrieved that way

jbagga:
@rynowak Updated

jbagga:
@rynowak Updated

Issue #440, opened 2017-09-06
Build initial statup experience 


rynowak:
```
public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddDispatcher();
    }

    public void Configure(IApplicationBuilder app)
    {
        app.UseDispatcher()

        app.UseCors();

        app.UseSomething();

        // hidden terminator middleware here
    }
}
```
Step 1:
- Implement IStartupFilter https://andrewlock.net/exploring-istartupfilter-in-asp-net-core/
- In UseDispatcher - add a 'magic' token to the app.Properties
- In your startup filter, look for the magic token and then add terminator middleware if is present

Issue #437, opened 2017-09-06
Move the two middleware into .Dispatcher


rynowak:
Move the dispatcher middleware and end middleware into the .Dispatcher project

PR #442, opened 2017-09-11
Startup experience

jbagga:
Addresses #437 and #440 

cc @rynowak

### DIFF
--- a/samples/DispatcherSample/Startup.cs
+++ b/samples/DispatcherSample/Startup.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Dispatcher;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
@@ -16,7 +15,7 @@ namespace DispatcherSample
         {
             services.AddSingleton<UrlGenerator>();
             services.AddSingleton<RouteValueAddressTable>();
-            services.AddSingleton<IStartupFilter, DispatcherEndpointStartupFilter>();
+            services.AddDispatcher();
         }
 
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
--- a/samples/DispatcherSample/Startup.cs
+++ b/samples/DispatcherSample/Startup.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Dispatcher;
 using Microsoft.AspNetCore.Hosting;
@@ -17,6 +16,7 @@ namespace DispatcherSample
         {
             services.AddSingleton<UrlGenerator>();
             services.AddSingleton<RouteValueAddressTable>();
+            services.AddSingleton<IStartupFilter, DispatcherEndpointStartupFilter>();
         }
 
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
@@ -27,47 +27,7 @@ namespace DispatcherSample
                 await next.Invoke();
             });
 
-            var dictionary = new Dictionary<string, DispatcherFeature>
-            {
-                {
-                    "/example",
-                    new DispatcherFeature
-                        {
-                            Endpoint = new DispatcherEndpoint("example"),
-                            RequestDelegate = async (context) =>
-                            {
-                                await context.Response.WriteAsync("Hello from the example!");
-                            }
-                        }
-                },
-                {
-                    "/example2",
-                    new DispatcherFeature
-                        {
-                            Endpoint = new DispatcherEndpoint("example2"),
-                            RequestDelegate = async (context) =>
-                            {
-                                await context.Response.WriteAsync("Hello from the second example!");
-                            }
-                        }
-                },
-            };
-
-            app.Use(async (context, next) =>
-            {
-                if (dictionary.TryGetValue(context.Request.Path, out var value))
-                {
-                    var dispatcherFeature = new DispatcherFeature
-                    {
-                        Endpoint = value.Endpoint,
-                        RequestDelegate = value.RequestDelegate
-                    };
-
-                    context.Features.Set<IDispatcherFeature>(dispatcherFeature);
-                    await context.Response.WriteAsync("<p>Dispatch</p>");
-                    await next.Invoke();
-                }
-            });
+            app.UseDispatcher();
 
             app.Use(async (context, next) =>
             {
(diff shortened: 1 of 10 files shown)


samples/DispatcherSample/Startup.cs
```diff
@@ -17,6 +16,7 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddSingleton<UrlGenerator>();
             services.AddSingleton<RouteValueAddressTable>();
+            services.AddSingleton<IStartupFilter, DispatcherEndpointStartupFilter>();
```
rynowak: You should create an 'AddDispatcher()` method and have that responsible for adding the startup filter

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+            if (dictionary.TryGetValue(httpContext.Request.Path, out var value))
+            {
+                var dispatcherFeature = new DispatcherFeature
+                {
+                    Endpoint = value.Endpoint,
+                    RequestDelegate = value.RequestDelegate
+                };
+
+                httpContext.Features.Set<IDispatcherFeature>(dispatcherFeature);
+                await httpContext.Response.WriteAsync("<p>Dispatch</p>");
```
rynowak: This line shouldn't be here  😆

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+                    new DispatcherFeature
+                        {
+                            Endpoint = new DispatcherEndpoint("example2"),
+                            RequestDelegate = async (context) =>
+                            {
+                                await context.Response.WriteAsync("Hello from the second example!");
+                            }
+                        }
+                },
+            };
```
rynowak: I think this is fine for now. I'll log a follow-up item to move this out.

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddlewareExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Builder;
+
+namespace Microsoft.AspNetCore.Dispatcher
```
rynowak: This should be in the `Microsoft.AspNetCore.Builder` namespace. We do this so you don't need to a million namespaces to configure middleware.

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddlewareExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public static class DispatcherMiddlewareExtensions
```
rynowak: DispatcherApplicationBuilderExtensions

samples/DispatcherSample/Startup.cs
rynowak: Add a new `DispatcherServiceCollectionExtensions` class in the `Microsoft.Extensions.DependencyInjection` namespace

src/Microsoft.AspNetCore.Dispatcher/EndpointMiddleware.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            var feature = context.Features.Get<IDispatcherFeature>();
+            await feature.RequestDelegate(context);
+            await _next(context);
+        }
```
rynowak: The logic here needs to make a bit more sense.

This should check `feature.RequestDelegate` for null, and call `_next` only if it's null.

src/Microsoft.AspNetCore.Dispatcher/EndpointMiddleware.cs
```diff
@@ -18,8 +18,12 @@ public EndpointMiddleware(RequestDelegate next)
...
         {
             var feature = context.Features.Get<IDispatcherFeature>();
-            await feature.RequestDelegate(context);
-            await _next(context);
+            var requestDelegate = feature.RequestDelegate(context);
+            await requestDelegate;
+            if (requestDelegate == null)
+            {
+                await _next(context);
+            }
```
rynowak: This is still not right. Think about what would happen if `feature.RequestDelegate` is null

jbagga:
@rynowak Updated

src/Microsoft.AspNetCore.Dispatcher/Microsoft.AspNetCore.Dispatcher.csproj
```diff
@@ -12,6 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Hosting" />
```
davidfowl: This should be Microsoft.AspNetCore.Hosting.Abstractions

src/Microsoft.AspNetCore.Dispatcher/EndpointMiddleware.cs
```diff
@@ -0,0 +1,31 @@
...
+        public async Task Invoke(HttpContext context)
+        {
+            var feature = context.Features.Get<IDispatcherFeature>();
+            if (feature.RequestDelegate == null)
+            {
+                await _next(context);
+            }
+            else
+            {
+                await feature.RequestDelegate(context);
```
davidfowl: It might be a problem that the RequestDelegate can't get a handle on `next`.

/cc @rynowak

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
davidfowl: This should be part of the options no?

samples/DispatcherSample/Startup.cs
```diff
@@ -27,47 +26,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
-                        Endpoint = value.Endpoint,
-                        RequestDelegate = value.RequestDelegate
-                    };
-
-                    context.Features.Set<IDispatcherFeature>(dispatcherFeature);
-                    await context.Response.WriteAsync("<p>Dispatch</p>");
-                    await next.Invoke();
-                }
-            });
+            app.UseDispatcher();
```
davidfowl: Why is this being called when there's a Startup filter?

src/Microsoft.AspNetCore.Dispatcher/Microsoft.AspNetCore.Dispatcher.csproj
jbagga: But I want to use `IStartupFilter` which is in this namespace https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting.Abstractions/IStartupFilter.cs#L7

src/Microsoft.AspNetCore.Dispatcher/EndpointMiddleware.cs
rynowak: I don't understand why that would be desirable. It's called an **end**point

samples/DispatcherSample/Startup.cs
jbagga: The Startup filter is to call the hidden (terminator) endpoint middleware if this is called. This is where the user is specifying that the Dispatcher should be used. 

/cc @rynowak

src/Microsoft.AspNetCore.Dispatcher/EndpointMiddleware.cs
davidfowl: @rynowak Because it restricts the types of things you can build with this system. I can't handle /foo and pass through to the next middleware.

Issue #438, opened 2017-09-06
Make the Dispatcher Middleware use Route Templates


rynowak:
Use the RouteTemplate and TemplateMatcher classes in the dispatcher middleware

PR #443, opened 2017-09-14
Use RouteTemplate and TemplateMatcher

jbagga:
Addresses #438 

Design for using RouteTemplate, TemplateParser and TemplateMatcher

cc @rynowak

### DIFF
--- a/samples/DispatcherSample/Startup.cs
+++ b/samples/DispatcherSample/Startup.cs
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Dispatcher;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Template;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace DispatcherSample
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
```diff
@@ -1,60 +1,93 @@
...
 
         public async Task Invoke(HttpContext httpContext)
         {
-            var dictionary = new Dictionary<string, DispatcherFeature>
+            foreach (var entry in _options.DispatcherEntryList)
             {
+                var parsedTemplate = entry.RouteTemplate;
+                var defaults = GetDefaults(parsedTemplate);
+                var templateMatcher = new TemplateMatcher(parsedTemplate, defaults);
+                var values = new RouteValueDictionary();
```
rynowak: This doesn't get used, it gets overwritten pretty much immediately.

rynowak: Just stick to `out var` at line 34

samples/DispatcherSample/Startup.cs
```diff
@@ -13,6 +16,60 @@ public class Startup
...
+                options.DispatcherEntryList = new List<DispatcherEntry>()
+                {
+                    new DispatcherEntry
+                    {
+                        RouteTemplate = TemplateParser.Parse("{Endpoint=example}"),
+                        Endpoints = new []
+                        {
+                            new RouteValuesEndpoint("example")
+                            {
+                                RouteValueDictionary = new RouteValueDictionary(new { Endpoint = "First" }),
```
rynowak: This isn't a great name, suggestion `Values` or `RequiredValues`

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
```diff
@@ -1,60 +1,93 @@
...
-                },
-            };
+                    result.Add(parameter.Name, parameter.DefaultValue);
+                }
+            }
+
+            return result;
+        }
+
+        private bool MatchURLToEndpoint(RouteValueDictionary currentURLRouteValueDictionary, RouteValueDictionary endpointRouteValueDictionary)
```
rynowak: Names need love. 

1. `MatchURLToEndpoint` doesn't describe what's happening here. This is matching the route values to the endpoints
2. Suggested names for parameters `values`, `current values`

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
```diff
@@ -1,60 +1,93 @@
...
+        {
+            var endpointMatch = new RouteValueDictionary();
```
rynowak: This isn't use for anything other than maintaining a count, so we just use an int if you need that

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
```diff
@@ -1,60 +1,93 @@
...
+                    endpointMatch[key] = endpointRouteValueDictionary[key];
+                }
             }
+
+            if (endpointMatch.Count == currentURLRouteValueDictionary.Count)
+            {
+                return true;
+            }
+
+            return false;
```
rynowak: Suggestion on the algorithm.

Psuedocode
```
for each kvp in requiredValues
{
   if (string.IsNullOrEmpty(kvp.Value))
   {
      if (values.TryGetValue(kvp.Key, out var routeValue) && !string.IsNullOrEmpty(routeValue))
      {
          return false;
      }
   }
   else
   {
        if (!values.TryGetValue(kvp.Key, out var routeValue)  || string.Equals(kvp.Value, routeValue, OrdinalIgnoreCase))
        {
            return false;
         }
    }
}

return true;
```

Fun Facts:
1. All route values need to be converted to strings so they can be compared (usually they already are strings)
2. All route value strings are compared ignoring case
3. Null and empty string are considered equivalent
5. All of the values in the endpoint's route value collection that have a null/empty value require that the corresponding value from the route is null/empty/not-present
6. All of the values in the endpoint's route value collection that have a non-null value require that the corresponding value from the route is equal
7. The route values provided by the route can and will have more values (`.Count`) than the route values required by the endpoint

src/Microsoft.AspNetCore.Dispatcher/DispatcherOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.AspNetCore.Routing.Template;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    public class DispatcherOptions
+    {
+        public IList<DispatcherEntry> DispatcherEntryList { get; set; }
+    }
+
+    public class DispatcherEntry
```
rynowak: This gets its own file

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
jbagga: But it isn't an out parameter

rynowak: OH. You're right by golly. You need to clear it then if the route matches, but none of the endpoints do. TryMatch will dirty the dictionary

jbagga:
@rynowak Do we want to work on the route value matching algorithm as part of this PR?

src/Microsoft.AspNetCore.Dispatcher/DispatcherMiddleware.cs
```diff
@@ -1,60 +1,97 @@
...
-                            RequestDelegate = async (context) =>
+                            var dispatcherFeature = new DispatcherFeature
                             {
-                                await context.Response.WriteAsync("Hello from the second example!");
-                            }
+                                Endpoint = endpoint,
+                                RequestDelegate = endpoint.RequestDelegate
+                            };
+
+                            httpContext.Features.Set<IDispatcherFeature>(dispatcherFeature);
```
rynowak: You also need to a `break` here. You don't want to process multiple entries - we should move on once we have a match

rynowak:
I think it's OK if you fix the algorithm in a follow up. You need to fix this one https://github.com/aspnet/Routing/pull/443#discussion_r139012358

jbagga:
@rynowak Added `break` to the loop

PR #445, opened 2017-09-17
Relayer implementation

rynowak:
This refactor introduces two major changes

1. Now creating the 'handler' delegate happens inside the endpoint
middleware. This allows you to short circuit even harder, AND to create
endpoint funcs that capture and use 'next' to rejoin the middleware
pipeline.

2. Relayered the implementation to have routing plug into the dispatcher.
It wasn't immediately apparent to me that this was the right thing to do,
but I think we will need to do things this way to deliver the kind of
back-compat experience we need to do.

The idea that I have is that 'attribute routing' will be the 'default'
entry in the dispatcher. Adding additional conventional routes or other
IRouter-based extensibility will be possible through adapters - but the
default experience will be to add items to the 'attribute route'.

So. We will need to port the attribute routing infrastructure to the
dispatcher library.

We may also need to make RVD into a subclass of something in the
dispatcher assembly.


samples/DispatcherSample/Startup.cs
```diff
@@ -1,73 +1,79 @@
...
         {
             services.Configure<DispatcherOptions>(options =>
             {
-                options.DispatcherEntryList = new List<DispatcherEntry>()
-                {
-                    new DispatcherEntry
-                    {
-                        RouteTemplate = TemplateParser.Parse("{Endpoint=example}"),
-                        Endpoints = new []
+                options.Dispatchers.Add(CreateDispatcher(
```
rynowak: @jbagga - this is the same set of endpoint and routes you used, I just made it gross-er.

samples/DispatcherSample/Startup.cs
```diff
@@ -99,5 +105,11 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 await next.Invoke();
             });
         }
+
+        private static RequestDelegate CreateDispatcher(string routeTemplate, RouteValuesEndpoint endpoint, params RouteValuesEndpoint[] endpoints)
+        {
+            var dispatcher = new RouterDispatcher(new Route(new RouterEndpointSelector(new[] { endpoint }.Concat(endpoints)), routeTemplate, ConstraintResolver));
+            return dispatcher.InvokeAsync;
```
rynowak: Yeah, I had to do **horrible** things.

src/Microsoft.AspNetCore.Dispatcher.Abstractions/EndpointDelegateFactory.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// A delegate which attempts to create a <see cref="RequestDelegate"/> for the selected <see cref="Endpoint"/>.
+    /// </summary>
+    /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
+    /// <param name="requestDelegate">The next middleware in the pipeline. Usage of optional.</param>
+    /// <param name="endpoint">The <see cref="Endpoint"/> selected by the dispatcher.</param>
+    /// <returns>
+    /// A <see cref="RequestDelegate"/> that invokes the operation represented by the <see cref="Endpoint"/>, or <c>null</c>.
+    /// </returns>
+    public delegate RequestDelegate EndpointDelegateFactory(HttpContext httpContext, RequestDelegate requestDelegate, Endpoint endpoint);
```
rynowak: @davidfowl - I know we wanted to stick to delegate/simple signatures but this is horrible - need some suggestions 😆 

One idea is to not pass the endpoint. Then this ends up being the same as one of the `Use(...)` signatures. Buuuuuuuuuuuttt we expecte you to need the endpoint.... hmmm 🤔

src/Microsoft.AspNetCore.Dispatcher/DispatcherOptions.cs
```diff
@@ -1,12 +1,15 @@
...
 
+using Microsoft.AspNetCore.Http;
 using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Dispatcher
 {
     public class DispatcherOptions
     {
-        public IList<DispatcherEntry> DispatcherEntryList { get; set; }
+        public IList<RequestDelegate> Dispatchers { get; } = new List<RequestDelegate>();
```
rynowak: I think that this will end up needing to be a custom collection type or a builder. Right now the delegate is the primitive.. that's all you need. **BUT** I want to have objects so we can do diagnostics and things like route debugging properly.

It needs to be possible to go bare metal, but our implementations should have rich diagnostics and you only get that by having a richer set of interfaces.

src/Microsoft.AspNetCore.Dispatcher/Microsoft.AspNetCore.Dispatcher.csproj
```diff
@@ -9,12 +9,13 @@
...
-    <ProjectReference Include="..\Microsoft.AspNetCore.Routing\Microsoft.AspNetCore.Routing.csproj" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Hosting" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" />
     <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Options" />
   </ItemGroup>
```
rynowak: @jbagga trimming the set of dependencies to the real minimal set

src/Microsoft.AspNetCore.Dispatcher/SimpleEndpoint.cs
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    public class SimpleEndpoint : Endpoint
```
rynowak: Currently unused by the sample, but this is plumbing for a `Use(...)`/`Map(...)` like experience.

src/Microsoft.AspNetCore.Routing/Dispatcher/RouterDispatcher.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public async override Task InvokeAsync(HttpContext httpContext)
+        {
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            var routeContext = new RouteContext(httpContext);
+            await _router.RouteAsync(routeContext);
```
rynowak: This isn't really **all** we need to do to plug this in. We have to full-on fake the `RouteData` including a fake route collection. We also have to create the router feature.

I'm feeling pretty good about this though. I think it won't be super complicated to plug in routing in a super compatible way. Then anyone using `IRouter` based exensibility doesn't have to throw it away.

src/Microsoft.AspNetCore.Routing/Dispatcher/RouterEndpointSelector.cs
```diff
@@ -0,0 +1,110 @@
...
+            var handler = GetRequestHandler(context.HttpContext, context.RouteData);
+            if (handler != null)
+            {
+                context.Handler = handler;
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private bool CompareRouteValues(RouteValueDictionary values, RouteValueDictionary requiredValues)
```
rynowak: This is currently a very silly implementation of something like MVC's action selection. @jbagga is going to tune it up.

src/Microsoft.AspNetCore.Dispatcher.Abstractions/EndpointDelegateFactory.cs
davidfowl: Make it return the middleware signature instead Func\<RequestDelegate, RequestDelegate\> does ignore need ge httpContext immediately? Can turn return a the middleware signature passing in the endpoint only?

rynowak: The reason I passed in the HttpContext originally is that I think you want to get access to features for things like route values, but maybe not. I'll go with your suggestion, and we can see if it works out or not.

rynowak: Also services.

rynowak:
I took your suggestion, I have some concerns that we'll run into issues in the future with handler factories needing access to context/services but I think this is better than what I had. In general everywhere we've tried something like this in MVC (extensibility point without access to services) someone finds a case where they need it. Building more stuff in the future will prove whether this is good or bad I think 👍 

I'm going to go ahead and merge this now - if you think more changes are needed right now lets figure it out during the week.

davidfowl:
Maybe it should be an interface? We'll know more soon I guess

PR #446, opened 2017-09-18
Two DI experiments

rynowak:
Two big ideas here:

1. Make it less breaky when we provide a **base class** that also
requires services. This introduces a separate composition root that we
can change. Today we can't really add services to things we expect users
to subclass because it's always a breaking change.

2. Make it easy to register things in DI that require services and have
that show up in options.

Interested in feedback on these ideas. I think these are both things that
have some value as an experiment but aren't really required.

I think there are still some issues with construting 'routes' that
aren't resolved by adding these two pieces.


samples/DispatcherSample/Startup.cs
```diff
@@ -99,7 +99,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 
         private static RequestDelegate CreateDispatcher(string routeTemplate, RouteValuesEndpoint endpoint, params RouteValuesEndpoint[] endpoints)
         {
-            var dispatcher = new RouterDispatcher(new Route(new RouterEndpointSelector(new[] { endpoint }.Concat(endpoints)), routeTemplate, ConstraintResolver));
+            var dispatcher = new RouterDispatcher(null, new Route(new RouterEndpointSelector(new[] { endpoint }.Concat(endpoints)), routeTemplate, ConstraintResolver));
```
rynowak: HACK

src/Microsoft.AspNetCore.Dispatcher/DependencyInjection/DefaultDispatcherConfigureOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Dispatcher;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    internal class DefaultDispatcherConfigureOptions : IConfigureOptions<DispatcherOptions>
```
rynowak: This thing will auto-configure `DispatcherOptions` based on DI. The separate interfaces are needed because because the delegate types we use on dispatcher options aren't DI friendly.

Basically you need to register a class, but we need to retrieve a delegate. Extra interfaces are needed.

rynowak: This would work pretty well for something like MVC's model binders and formatters where the options are really doing a job that DI could already do.

src/Microsoft.AspNetCore.Dispatcher/DependencyInjection/DispatcherServiceCollectionExtensions.cs
```diff
@@ -17,6 +17,14 @@ public static IServiceCollection AddDispatcher(this IServiceCollection services)
             }
 
             services.AddSingleton<IStartupFilter, DispatcherEndpointStartupFilter>();
+
+            services.AddSingleton<DispatcherBaseServices, DefaultDispatcherBaseServices>();
+            services.AddSingleton<HandlerFactoryBaseServices, DefaultHandlerFactoryBaseServices>();
+
+            services.ConfigureOptions<DefaultDispatcherConfigureOptions>();
+
+            services.AddTransient<IHandlerFactoryOptionsValueProvider, SimpleEndpointHandlerFactory>();
```
rynowak: Example, here's an endpoint factory that's always available. But it also comes from DI.

src/Microsoft.AspNetCore.Dispatcher/DispatcherBase.cs
```diff
@@ -1,13 +1,42 @@
...
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Dispatcher
 {
-    public abstract class DispatcherBase
+    public abstract class DispatcherBase : IDispatcherOptionsValueProvider
```
rynowak: Convenience base class for dispatchers. It provides you with some services (assuming we add more services) and lets you register it with DI.

src/Microsoft.AspNetCore.Dispatcher/HandlerFactoryBaseServices.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    public abstract class HandlerFactoryBaseServices
+    {
+        public virtual ILoggerFactory LoggerFactory { get; protected set; }
+    }
```
rynowak: We can add additional services to this without it being a binary breaking change. Anyone that's mocking this for tests will need to respond if their tests cover those code paths.

src/Microsoft.AspNetCore.Dispatcher/SimpleEndpointHandlerFactory.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    public class SimpleEndpointHandlerFactory : HandlerFactoryBase
+    {
+        public SimpleEndpointHandlerFactory(HandlerFactoryBaseServices services) 
+            : base(services)
+        {
+        }
```
rynowak: One nice thing about this is that VS will generate this constructor for you verbatim

src/Microsoft.AspNetCore.Dispatcher/SimpleEndpointHandlerFactory.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        public override Func<RequestDelegate, RequestDelegate> CreateHandler(Endpoint endpoint)
+        {
+            return (endpoint as SimpleEndpoint).DelegateFactory;
```
rynowak: should be `?.` lol

src/Microsoft.AspNetCore.Routing/Dispatcher/DefaultRouterDispatcherFactory.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Dispatcher;
+
+namespace Microsoft.AspNetCore.Routing.Dispatcher
+{
+    internal class DefaultRouterDispatcherFactory : RouterDispatcherFactory
```
rynowak: This needs more love/thought. Routes require services today so newing them up is a pain.

rynowak:
Closing this PR, no plans to take these changes at this time.

PR #448, opened 2017-09-21
Split RVD in twain

rynowak:


src/Microsoft.AspNetCore.Dispatcher.Abstractions/DispatcherValueCollection.cs
```diff
@@ -0,0 +1,790 @@
...
+using System.Diagnostics;
+using Microsoft.AspNetCore.Dispatcher.Abstractions;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    /// <summary>
+    /// An <see cref="IDictionary{String, Object}"/> type for dispatcher values.
+    /// </summary>
+    public class DispatcherValueCollection : IDictionary<string, object>, IReadOnlyDictionary<string, object>
```
rynowak: Copied almost verbatim from RVD, no need to review.

rynowak:
sure

rynowak:
The things that would break btw would be error messages that now include a different name

PR #449, opened 2017-09-22
Add endpoint disambiguation

rynowak:
- Better sample of metadata
- Sample shows how conventional routing would work
- Added endpoint disambiguation


samples/DispatcherSample/HttpMethodEndpointSelector.cs
```diff
@@ -0,0 +1,86 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Dispatcher;
+
+namespace DispatcherSample
+{
+    public class HttpMethodEndpointSelector : EndpointSelector
```
rynowak: This is an off the cuff implementation from memory of MVC's http method selection + backtracking

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    public class DispatcherValueEndpointSelector : EndpointSelector
```
rynowak: This is a brute force (and wrong) implementation. I wanted to save the real thing for @jbagga

src/Microsoft.AspNetCore.Routing/Dispatcher/RouteTemplateDispatcher.cs
```diff
@@ -0,0 +1,157 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Dispatcher;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing.Template;
+
+namespace Microsoft.AspNetCore.Routing.Dispatcher
+{
+    public class RouteTemplateDispatcher : DispatcherBase
```
rynowak: This is mostly a copy-paste of all the boilerplate from `RouteBase`

src/Microsoft.AspNetCore.Dispatcher.Abstractions/IDispatcherFeature.cs
```diff
@@ -10,5 +10,7 @@ public interface IDispatcherFeature
         Endpoint Endpoint { get; set; }
 
         RequestDelegate RequestDelegate { get; set; }
+
+        DispatcherValueCollection Values { get; set; }
```
davidfowl: This looks suspect

rynowak: Let's discuss this. We need to make a call about whether or not route values are first class

Issue #444, opened 2017-09-15
URL disambiguation/mapping to endpoints for Dispatcher


jbagga:
Implement URL to endpoint disambiguation algorithm.

cc @rynowak

jbagga:
@rynowak Should this be closed? Since we are not going forward with this work item

rynowak:
Yes, close please

Issue #439, opened 2017-09-06
Add logging


rynowak:
We should add logging to:
- [ ] DispatcherMiddleware
- [ ] EndpointMiddleware
- [ ] DispatcherBase + all subclasses
- [ ] EndpointSelector + all subclasses

Take a stab at this and try to figure out the ideal things to log - use routing as an inspiration.

anurse:
One the one, as referenced in various issues (https://github.com/aspnet/SignalR/issues/876 https://github.com/aspnet/Hosting/issues/1205 https://github.com/aspnet/Mvc/issues/6818): make sure that when you specify an Event ID, you use `new EventId([number], "[name]")` rather than just an `int`. Having names associated with the events makes it much easier to filter and avoids having to look up obscure `int`s. The names should be unique but only within a particular `ILogger` category.

anurse:
That and: Log lots, use Trace if you're worried about logging too much :).

jbagga:
@rynowak Should the `EndpointSelectorContext` have a logger? Then `HttpMethodEndpointSelector` can use context.Logger inside the `SelectAsync` [method](https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Dispatcher/HttpMethodEndpointSelector.cs#L12). Is it correct to get that from DI [here](https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Dispatcher/EndpointSelectorContext.cs#L16)?

rynowak:
> Should the EndpointSelectorContext have a logger? 

No, we don't usually pass services this way. Let's discuss this at the meeting tomorrw

PR #451, opened 2017-09-26
Selectors initialization and DispatcherBase logging

jbagga:
Addresses #444 

I added `IEndpointCollection` to track `Version` of the cached endpoints, but after a conversation with @rynowak that may not be required anymore.


src/Microsoft.AspNetCore.Dispatcher.Abstractions/IEndpointCollection.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    public interface IEndpointCollection
```
rynowak: This was a fine idea, but I don't think we need it.

src/Microsoft.AspNetCore.Dispatcher/DispatcherBase.cs
```diff
@@ -78,17 +82,17 @@ public virtual IList<EndpointSelector> Selectors
                     return;
                 }
 
-                var selectorContext = new EndpointSelectorContext(httpContext, Endpoints.ToList(), Selectors);
+                var selectorContext = new EndpointSelectorContext(httpContext, Endpoints, Selectors);
```
rynowak: Be wary of changes that remove a `ToList()` or similar. This was an intentional defensive copy

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,262 @@
...
+                if (cache != null && cache.Version == endpoints.Version)
+                {
+                    return cache;
+                }
+
+                cache = new Cache(endpoints);
+                Volatile.Write(ref _cache, cache);
+                return cache;
+            }
+        }
```
rynowak: I'd like to see what this code looks like without the updating feature.

Can you instead add an `public virtual void Intialize(IEndpointCollection)` to the `EndpointSelector` class and do the initialization in there?

Call it from inside of `DispatcherBase`

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,262 @@
...
+            var dispatcherFeature = _endpointSelectorContext.HttpContext.Features.Get<IDispatcherFeature>();
 
-            for (var i = context.Endpoints.Count - 1; i >= 0; i--)
+            var candidates = SelectCandidates(dispatcherFeature.Values);
+            var selectedEndpoint = SelectBestCandidate(candidates);
+            if (selectedEndpoint == null)
             {
-                var endpoint = context.Endpoints[i] as IDispatcherValueSelectableEndpoint;
-                if (!CompareRouteValues(dispatcherFeature.Values, endpoint.Values))
+                // To do: Add logger
```
rynowak: No todos. If you need services, initialize them on the first request

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,262 @@
...
+                    return cache;
+                }
+
+                cache = new Cache(endpoints);
+                Volatile.Write(ref _cache, cache);
+                return cache;
+            }
+        }
+
+        public override Task SelectAsync(EndpointSelectorContext endpointSelectorContext)
```
rynowak: `context`

src/Microsoft.AspNetCore.Dispatcher/DispatcherLoggerExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+            "No endpoints matched the current request. Dispatcher values: {DispatcherValueCollection}");
+
+        public static void AmbiguousEndpoints(this ILogger logger, string ambiguousEndpoints)
+        {
+            _ambiguousEndpoints(logger, ambiguousEndpoints, null);
+        }
+
+        public static void EndpointMatched(this ILogger logger, Endpoint endpoint)
+        {
+            _endpointMatched(logger, endpoint, null);
```
jbagga: @rynowak I wanted to use the Endpoint.DisplayName property here but I realized that it may sometimes not be set/be null.
What's a sensible logging message in that case?

jbagga:
@rynowak Updated. I will add logging in DispatcherBase as part of #439

src/Microsoft.AspNetCore.Dispatcher/DispatcherLoggerExtensions.cs
rynowak: DisplayName is 100% the thing to use. If someone set it to null they can cry about it.

src/Microsoft.AspNetCore.Dispatcher/AmbiguousEndpointException.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    /// <summary>
+    /// An exception which indicates multiple matches in endpoint selection.
+    /// </summary>
+    [Serializable]
+    public class AmbiguousEndpointException : InvalidOperationException
```
rynowak: Let's not subclass `InvalidOperationException`, it's a little wierd. yeah the old code might have done this but we don't need it.

src/Microsoft.AspNetCore.Dispatcher/DispatcherBase.cs
```diff
@@ -117,6 +121,16 @@ protected virtual Task<bool> TryMatchAsync(HttpContext httpContext)
...
 
+            if (!_selectorsInitialized)
+            {
+                foreach (var selector in Selectors)
+                {
+                    selector.Initialize(this);
+                }
+
+                _selectorsInitialized = true;
+            }
```
rynowak: I think it would be better here to use LazyInitializer here: https://msdn.microsoft.com/en-us/library/dd642260(v=vs.110).aspx

You can just have it return null.

The problem with this pattern is that it can run multiple times when multiple threads come in before initialization completes. We've seen this in routing on ARM and I don't want it to start happening in the new codebase

src/Microsoft.AspNetCore.Dispatcher/DispatcherBase.cs
```diff
@@ -129,8 +143,15 @@ protected virtual Task<bool> TryMatchAsync(HttpContext httpContext)
                     return selectorContext.Endpoints[0];
 
                 default:
-                    throw new InvalidOperationException("Ambiguous bro!");
```
rynowak: 😀

src/Microsoft.AspNetCore.Dispatcher/DispatcherLoggerExtensions.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Dispatcher.Internal
```
rynowak: You can drop the `.Internal` from this. The class is actually-internal so it doesn't really do anything

src/Microsoft.AspNetCore.Dispatcher/DispatcherLoggerExtensions.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
```
rynowak: Copyright header

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
     {
+        private static readonly IReadOnlyList<SimpleEndpoint> EmptyEndpoints = Array.Empty<SimpleEndpoint>();
+
+        private EndpointSelectorContext _context;
+        private Cache _cache;
+
+        private ILogger<DispatcherValueEndpointSelector> _logger;
+        private bool _servicesInitialized;
+
+        public override void Initialize(IEndpointCollectionProvider endpointCollectionProvider)
```
rynowak: Just call this `endpointProvider`

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
+
         public override Task SelectAsync(EndpointSelectorContext context)
         {
             if (context == null)
             {
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var dispatcherFeature = context.HttpContext.Features.Get<IDispatcherFeature>();
+            _context = context;
```
rynowak: This should not be here. This is a big red flag.

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
+            if (!_servicesInitialized)
+            {
+                EnsureServices(context.HttpContext);
```
rynowak: EnsureServices should have the `if` inside of it.

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
rynowak: You can't cache this because you need to handle multiple requests at a time

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
+            }
 
-            for (var i = context.Endpoints.Count - 1; i >= 0; i--)
+            var dispatcherFeature = _context.HttpContext.Features.Get<IDispatcherFeature>();
+            var candidates = SelectCandidates(dispatcherFeature.Values);
+            _context.Endpoints.Clear();
```
rynowak: ???

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
+
+            foreach (var candidate in candidates)
+            {
+                _context.Endpoints.Add(candidate);
+            }
+
+            return _context.InvokeNextAsync();
+        }
+
+        public IReadOnlyList<Endpoint> SelectCandidates(DispatcherValueCollection dispatcherValueCollection)
```
rynowak: call this `values`

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
+                    _logger.EndpointMatched(candidates[0].DisplayName);
+                    break;
+
+                default:
+                    var endpointNames = string.Join(
+                            Environment.NewLine,
+                            candidates.Select(a => a.DisplayName));
+                    _logger.AmbiguousEndpoints(endpointNames);
+                    break;
+            }
```
rynowak: I'm not sure how this code made it into here.. There should be one block like this and it should be in `DispatcherBase`.

These policies run in a sequence, so it's not ambiguous if we have multiple endpoints - the ambiguity could be resolved by a subsequent selector

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
-        private bool CompareRouteValues(DispatcherValueCollection values, DispatcherValueCollection requiredValues)
+
+        private class Cache
         {
-            foreach (var kvp in requiredValues)
+            public Cache(IReadOnlyList<Endpoint> endpoints)
             {
-                if (string.IsNullOrEmpty(kvp.Value.ToString()))
+                // We need to build two maps for all of the route values.
+                OrdinalEntries = new Dictionary<string[], List<SimpleEndpoint>>(StringArrayComparer.Ordinal);
```
rynowak: Is it possible to make these types `Endpoint`?

rynowak: Instead of `SimpleEndpoint`

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
+                OrdinalIgnoreCaseEntries = new Dictionary<string[], List<SimpleEndpoint>>(StringArrayComparer.OrdinalIgnoreCase);
+
+                // We need to first identify of the keys that action selection will look at (in route data). 
+                // We want to only consider conventionally routed actions here.
```
rynowak: Some of these comments need to be cleaned up a bit - these refer to some MVC concepts that aren't part of the dispatcher

src/Microsoft.AspNetCore.Dispatcher/DispatcherValueEndpointSelector.cs
```diff
@@ -2,54 +2,244 @@
...
+
+                return hash.CombinedHash;
             }
+        }
 
-            return true;
+        private void EnsureServices(HttpContext context)
+        {
+            var loggerFactory = (ILoggerFactory)context.RequestServices.GetService(typeof(ILoggerFactory));
+            _logger = loggerFactory.CreateLogger<DispatcherValueEndpointSelector>();
```
rynowak: Use the `GetRequiredService<T>` method.

You can get `ILogger<DispatcherValueEndpointSelector>` directly

src/Microsoft.AspNetCore.Dispatcher/EndpointSelector.cs
```diff
@@ -8,5 +8,9 @@ namespace Microsoft.AspNetCore.Dispatcher
     public abstract class EndpointSelector
     {
         public abstract Task SelectAsync(EndpointSelectorContext context);
+
+        public virtual void Initialize(IEndpointCollectionProvider endpointCollectionProvider)
```
rynowak: call it `collectionProvider`

src/Microsoft.AspNetCore.Dispatcher/Microsoft.AspNetCore.Dispatcher.csproj
```diff
@@ -15,7 +15,23 @@
...
     <PackageReference Include="Microsoft.Extensions.Options" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Update="Resources.Designer.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>Resources.resx</DependentUpon>
+    </Compile>
+  </ItemGroup>
```
rynowak: You don't need this. We do our resources in a special way

rynowak: Look at one of the other projects as an example

src/Microsoft.AspNetCore.Dispatcher/Resources.resx
```diff
@@ -0,0 +1,124 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="DefaultEndpointSelector_AmbiguousEndpoints" xml:space="preserve">
+    <value>Multiple endpoints matched. The following endpoints matched route data:{0}{0}{1}</value>
```
rynowak: `route data` -> `the request`

jbagga:
@rynowak Undid changes to `DispatcherValueEndpointSelector`

src/Microsoft.AspNetCore.Dispatcher.Abstractions/Endpoint.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.AspNetCore.Dispatcher
 {
-    [DebuggerDisplay("{DisplayName,nq}")]
+    [DebuggerDisplay("{GetType()}{DisplayName,nq}")]
```
rynowak: put a space here. this is going to look like `Some.Type.Namedkfdfdkbydk`

I'd suggest formatting like like `{GetType().Name} - '{DisplayName,nq}'` or something similar. Make it look reasonable

src/Microsoft.AspNetCore.Dispatcher/EndpointSelector.cs
```diff
@@ -8,5 +8,10 @@ namespace Microsoft.AspNetCore.Dispatcher
     public abstract class EndpointSelector
     {
         public abstract Task SelectAsync(EndpointSelectorContext context);
+
+        public virtual Task Initialize(IEndpointCollectionProvider endpointProvider)
```
rynowak: When/why did this become async? We need this to be synchronous because it's going to happen under a lock

src/Microsoft.AspNetCore.Dispatcher/DispatcherBase.cs
```diff
@@ -117,20 +137,62 @@ protected virtual Task<bool> TryMatchAsync(HttpContext httpContext)
...
+
+                    var message = Resources.FormatAmbiguousEndpoints(
+                        Environment.NewLine,
+                        endpointNames);
+
+                    throw new AmbiguousEndpointException(message);
+            }
+        }
+
+        private async Task InitializeServices()
```
rynowak: This should be `InitializeSelectors`

jbagga:
@rynowak Updated

src/Microsoft.AspNetCore.Dispatcher/EndpointSelector.cs
```diff
@@ -8,5 +8,10 @@ namespace Microsoft.AspNetCore.Dispatcher
     public abstract class EndpointSelector
     {
         public abstract Task SelectAsync(EndpointSelectorContext context);
+
+        public virtual object Initialize(IEndpointCollectionProvider endpointProvider)
```
rynowak: This should return void

rynowak:
Looks good. Approve x2

PR #461, opened 2017-09-29
Making endpoints and addresses easier

rynowak:
This makes endpoints and addresses easier to work with by dropping the
'metadata first' approach for the the things that are really at the core
of the dispatcher.


rynowak:
@davidfowl thoughts?

/cc @jbagga please have a look as well

src/Microsoft.AspNetCore.Dispatcher/ITemplateAddress.cs
```diff
@@ -3,8 +3,10 @@
 
 namespace Microsoft.AspNetCore.Dispatcher
 {
-    public interface IDispatcherValueSelectableEndpoint
+    public interface ITemplateAddress
```
rynowak: What do you need to generate a URL?

src/Microsoft.AspNetCore.Dispatcher/ITemplateEndpoint.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    public interface ITemplateEndpoint
```
rynowak: What do you need to build a route table?

src/Microsoft.AspNetCore.Dispatcher/ITemplateEndpoint.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        string HttpMethod { get; }
```
rynowak: We could also put this on another interface, but it feels like a core concept to me

rynowak: It's not strictly required

src/Microsoft.AspNetCore.Dispatcher/TemplateEndpoint.cs
```diff
@@ -0,0 +1,127 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    public class TemplateEndpoint : Endpoint, ITemplateEndpoint
```
rynowak: There are a lot of constructors here. Don't think of this as something that's going to be in the face of most users. We will have a more polished startup experience.

src/Microsoft.AspNetCore.Routing/Dispatcher/RouteTemplateUrlGenerator.cs
```diff
@@ -37,14 +37,13 @@ public string GenerateUrl(HttpContext httpContext, object values)
                 throw new ArgumentNullException(nameof(values));
             }
 
-            var address = _addressSelector.SelectAddress(new DispatcherValueCollection(values));
+            var address = _addressSelector.SelectAddress(new DispatcherValueCollection(values)) as ITemplateAddress;
```
rynowak: Contract of the `TemplateAddressSelector` returns an `ITemplateAddress + Address`. We don't have sum types tho. Sad panda

src/Microsoft.AspNetCore.Routing/Dispatcher/TreeDispatcher.cs
```diff
@@ -142,16 +142,16 @@ private Cache CreateCache()
...
                 {
                     continue;
                 }
 
-                if (!groups.TryGetValue(new Key(metadata.Order, metadata.RouteTemplate), out var group))
+                if (!groups.TryGetValue(new Key(0, templateEndpoint.Template), out var group))
                 {
                     group = new List<Endpoint>();
-                    groups.Add(new Key(metadata.Order, metadata.RouteTemplate), group);
+                    groups.Add(new Key(0, templateEndpoint.Template), group);
```
rynowak: ignoring order for now, I think we can get rid of that concept

rynowak: By making it separate dispatchers with a single tree

src/Microsoft.AspNetCore.Dispatcher/TemplateEndpoint.cs
davidfowl: Maybe a builder pattern would be better.

src/Microsoft.AspNetCore.Dispatcher/ITemplateEndpoint.cs
```diff
@@ -0,0 +1,14 @@
...
+
+        string Template { get; }
+
+        DispatcherValueCollection Values { get; }
```
davidfowl: Still not sure why this is available on the end point

samples/DispatcherSample/Startup.cs
```diff
@@ -32,27 +31,27 @@ public void ConfigureServices(IServiceCollection services)
...
-                        new SimpleEndpoint(Home_About, new object[] { new RouteTemplateMetadata("{controller=Home}/{action=Index}/{id?}"), }, new { controller = "Home", action = "About", }, "Home:About()"),
-                        new SimpleEndpoint(Admin_Index, new object[] { new RouteTemplateMetadata("{controller=Home}/{action=Index}/{id?}"), }, new { controller = "Admin", action = "Index", }, "Admin:Index()"),
-                        new SimpleEndpoint(Admin_GetUsers, new object[] { new RouteTemplateMetadata("{controller=Home}/{action=Index}/{id?}"), new HttpMethodMetadata("GET"), new AuthorizationPolicyMetadata("Admin"), }, new { controller = "Admin", action = "Users", }, "Admin:GetUsers()"),
-                        new SimpleEndpoint(Admin_EditUsers, new object[] { new RouteTemplateMetadata("{controller=Home}/{action=Index}/{id?}"),  new HttpMethodMetadata("POST"), new AuthorizationPolicyMetadata("Admin"), }, new { controller = "Admin", action = "Users", }, "Admin:EditUsers()"),
+                        new TemplateEndpoint("{controller=Home}/{action=Index}/{id?}", new { controller = "Home", action = "Index", }, Home_Index, "Home:Index()"),
```
davidfowl: Why are we still specifying defaults?

src/Microsoft.AspNetCore.Dispatcher/TemplateEndpoint.cs
rynowak: I built two different versions of that yesterday. I wanted to understand your ideas about application model before going further with it.

src/Microsoft.AspNetCore.Dispatcher/ITemplateEndpoint.cs
rynowak: These are the values that need to be present to select the endpoint. We can go further here by just doing token replacement

rynowak: Even if we do token replacement this property still has to be here. Attribute routing treats these like defaults, that's required for ambient values to work

samples/DispatcherSample/Startup.cs
rynowak: Even if we do token replacement this property still has to be here. Attribute routing treats these like defaults, that's required for ambient values to work

rynowak:
Discussed this in person, I'm going to go ahead and merge for now since this is cleanup.

Issue #452, opened 2017-09-27
Productize the HttpMethodEndpointSelector


rynowak:
Move https://github.com/aspnet/Routing/blob/dev/samples/DispatcherSample/HttpMethodEndpointSelector.cs and apply some polish to it

PR #463, opened 2017-10-05
Productize HttpMethodEndpointSelector

jbagga:
Addresses #452


test/Microsoft.AspNetCore.Dispatcher.FunctionalTest/ApiAppStartup.cs
```diff
@@ -62,15 +62,25 @@ public void ConfigureDispatcher(DispatcherOptions options)
             {
                 Endpoints =
                 {
-                    new TemplateEndpoint("api/products", Products_Get),
+                    new TemplateEndpoint("api/products", new { controller = "Products", action = "Get", }, "GET", Products_Get),
+                    new TemplateEndpoint("api/products", new { controller = "Products", action = "Post", }, "POST", Products_Post),
+                    new TemplateEndpoint("api/products", new { controller = "Products", action = "Put", }, "PUT", Products_Put),
                 },
```
jbagga: @rynowak I tried to add endpoints with a different route "api/products/{id}" but that throws an `IndexOutofBoundsException` here https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing/Dispatcher/TreeDispatcher.cs#L170 because it has two `entries` and just one element in `trees`

rynowak: Yeah I just found that bug here too 😆  The correct line should be
```
var tree = trees[entry.Order];
```

rynowak: Go ahead and fix it. This is why we're adding more tests

test/Microsoft.AspNetCore.Dispatcher.FunctionalTest/ApiAppTest.cs
```diff
@@ -30,5 +30,45 @@ public ApiAppTest(ApiAppFixture app)
...
+        {
+            // Arrange
+            var request = new HttpRequestMessage(HttpMethod.Delete, "/api/products");
+
+            // Act
+            var response = await Client.SendAsync(request);
+
+            // Assert
+            Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
+        }
```
rynowak: This is great. I'd also like to see a case where you have a 'fallback'. That would look like
```
new TemplateEndpoint("some/url", new { ..., }, null, SomeApi_Fallback),
new TemplateEndpoint("some/url", new { ..., }, "POST", SomeApi_Post),
```
If you send a GET or any other verb besides POST it should be handled by `SomeApi_Fallback`. This is what all of that fallback stuff is for.

Being specific about the verb you handle makes it a 'better' candidate.

test/Microsoft.AspNetCore.Dispatcher.Test/HttpMethodEndpointSelectorTest.cs
```diff
@@ -0,0 +1,99 @@
...
+        public async Task RequestMethod_MatchesEndpointMethod_IgnoresCase(string httpMethod)
+        {
+            // Arrange
+            var endpoints = new List<Endpoint>()
+            {
+                new TemplateEndpoint("{controller=Home}/{action=Index}/{id?}", new { controller = "Products", action = "Get", }, "GET", Products_Get, "Products:Get()"),
+                new TemplateEndpoint("{controller=Home}/{action=Index}/{id?}", new { controller = "Products", action = "Create", }, "POST", Products_Post, "Products:Post()"),
+            };
+
+            var context = GetEndpointSelectorContext(httpMethod, endpoints, out var selector);
```
rynowak: Use a valuetuple or split this into two calls
```
(var context, var selector) = CreateContextAndSelector(...)
```

test/Microsoft.AspNetCore.Dispatcher.Test/HttpMethodEndpointSelectorTest.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            // Act
+            await selector.SelectAsync(context);
+            var templateEndpoints = context.Endpoints.Select(e => e as TemplateEndpoint);
```
rynowak: Use `.Cast<TemplateEndpont>()` instead of select

test/Microsoft.AspNetCore.Dispatcher.Test/HttpMethodEndpointSelectorTest.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            // Assert
+            Assert.Collection(templateEndpoints,
```
rynowak: newline after paren

jbagga:
@rynowak Updated

test/Microsoft.AspNetCore.Dispatcher.FunctionalTest/ApiAppTest.cs
```diff
@@ -30,5 +30,60 @@ public ApiAppTest(ApiAppFixture app)
...
+
+        [Fact]
+        public async Task ApiApp_NoEndpointWithMatchingHttpMethod_FallbackEndpointSelected()
```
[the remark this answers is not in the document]
jbagga: Added a test that doesn't have a matching HttpMethod or a fallback

PR #465, opened 2017-10-06
Respond to feedback about handler factory

rynowak:
- Handler factories are now associated with the dispatcher entry
- Handler factory is now an interface


src/Microsoft.AspNetCore.Dispatcher/HandlerFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+using System;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    /// <summary>
+    /// A simple implementation of <see cref="IHandlerFactory"/> that adapts a particular endpoint
+    /// type to a handler.
+    /// </summary>
+    public sealed class HandlerFactory<TEndpoint> : IHandlerFactory
```
rynowak: come at me bro

davidfowl: Virus

src/Microsoft.AspNetCore.Dispatcher/DispatcherOptions.cs
```diff
@@ -8,7 +8,5 @@ namespace Microsoft.AspNetCore.Dispatcher
     public class DispatcherOptions
     {
         public MatcherCollection Matchers { get; } = new MatcherCollection();
-
-        public IList<EndpointHandlerFactory> HandlerFactories { get; } = new List<EndpointHandlerFactory>();
```
davidfowl: 👏

src/Microsoft.AspNetCore.Dispatcher/IDefaultMatcherFactory.cs
```diff
@@ -7,6 +7,6 @@ namespace Microsoft.AspNetCore.Dispatcher
 {
     public interface IDefaultMatcherFactory
```
davidfowl: Do we need an interface at all for the default?

rynowak: I introduced it because of a layering issue, but I kinda like it. It won't be necessary when we're done

src/Microsoft.AspNetCore.Dispatcher/HandlerFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public Func<RequestDelegate, RequestDelegate> CreateHandler(Endpoint endpoint)
+        {
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            if (endpoint is TEndpoint myTypeOfEndpoint)
```
jbagga: Do endpoints iterate over all available `IHandlerFactory` implementations to find the matching type? Will there be multiple factories?

rynowak: Yes, exactly - see the composite handler factory

Issue #466, opened 2017-10-06
Port TemplateParser/RouteTemplate to the dispatcher project


rynowak:
Move the implementations of the Template Parser and Template object model to the dispatcher project.

We may want to do some redesign of these things, but for now let's just move/rename them.

rynowak:
The implemenation in routing should be updated to use the code in dispatcher as much as possible.

jbagga:
@rynowak Is moving the `Routing.Dispatcher` namespace part of this work item? We use `Routing` classes like `RouteValueDictionary` and some internal routing stuff there. Should those be moved now?

rynowak:
I would start by moving the `RouteTemplate` and ``TemplateParser` to the dispatcher project.  If these use RVD, update the to use `DispatcherValueCollection`

PR #469, opened 2017-10-13
Port TemplateParser to Dispatcher

jbagga:
Addresses #466


src/Microsoft.AspNetCore.Dispatcher/InlineRouteParameterParser.cs
```diff
@@ -3,9 +3,8 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.AspNetCore.Routing.Template;
 
-namespace Microsoft.AspNetCore.Routing
```
rynowak: This is a breaking change - all of the old APIs need to still exist.

PR #470, opened 2017-10-13
Set RepositoryRoot

natemcmaster:
React to https://github.com/aspnet/BuildTools/pull/419


natemcmaster:
@ryanbrandenburg we need to make this change in our repos to ensure VS builds stay happy. RepositoryRoot is set by KoreBuild during command line build context, but its empty in a VS build context

rynowak:
:shipit:

PR #474, opened 2017-10-17
Avoid allocation of a new char array on every request in RouteCollection class

0xfeeddeadbeef:
This is a **very small micro-optimization**:

When `LowercaseUrls` and/or `AppendTrailingSlash` options are enabled, on every call to `RouteCollection.NormalizeVirtualPath` a `new char[] { '?', '#' }` is being allocated; which happens on every request.

(Does pull request of this _scale_ still require associated issue to exist?)


rynowak:
> (Does pull request of this scale still require associated issue to exist?)

I'm fine with it 👍 Thanks for the PR

Issue #475, opened 2017-10-18
InlineRouteParameterParser shouldn't exist


rynowak:
`InlineRouteParameterParser` is a separate static class from the `TemplateParser` with its own public entry point that is responsible for parsing the contents of a parameter.

Example:
`{controller=Home}/{action=Index}/{id?}`

In this case it would be called once for `controller`, once for `action` and once for `id?`. 

There are some layering problems with this API overall
- It doesn't handle escaping
- It is called with the contents of the parameter, not the whole parameter
- It doesn't need to be public, it's a detail of the template parser
- It can't report errors, it always returns a 'parameter', even if it's in valid

---------

We should obsolete this method in `Microsoft.AspNetCore.Routing` and remove it in 3.0. For the dispatcher we should fix all of the above issues and not make this functionality public.

rynowak:
We should do this after preview 1. It's not a critical issue

rynowak:
I'm putting a few notes in the code related to this. Do a search for `#475` when you fix this.

aspnet-hello:
This issue was moved to aspnet/Home#2658

PR #477, opened 2017-10-18
Redesign public API for templates

rynowak:
-Renamed RouteTemplate -> RoutePattern
-Made immutable
-Added Builder
-Lots of fixes to parser to support new design

There are a few small issues logged for follow-up but this is mostly in
the place I want it design-wise.


rynowak:
@jbagga please have a look

src/Microsoft.AspNetCore.Dispatcher/Patterns/InlineRouteParameterParser.cs
```diff
@@ -3,12 +3,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
-namespace Microsoft.AspNetCore.Dispatcher
+namespace Microsoft.AspNetCore.Dispatcher.Patterns
 {
     public static class InlineRouteParameterParser
```
rynowak: This class is totally broken. Everything here should be a detail of the parser. I logged #475 to follow up on this

src/Microsoft.AspNetCore.Dispatcher/Patterns/ConstraintReference.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Dispatcher.Patterns
+{
+    /// <summary>
+    /// The parsed representation of an constraint in a <see cref="RoutePattern"/> parameter.
```
jbagga: nit: "a constraint"

src/Microsoft.AspNetCore.Dispatcher/Patterns/InlineRouteParameterParser.cs
```diff
@@ -81,19 +75,15 @@ public static TemplatePart ParseRouteParameter(string routeParameter)
...
+            return new RoutePatternParameter(text, parameterName, defaultValue, parameterKind, parseResults.Constraints.ToArray());
         }
 
         private static ConstraintParseResults ParseConstraints(
             string routeParameter,
             int currentIndex,
             int endIndex)
         {
-            var inlineConstraints = new List<InlineConstraint>();
+            var inlineConstraints = new List<ConstraintReference>();
```
jbagga: Should this variable be renamed to something else? Or is `inlineConstraints` still relevant?

src/Microsoft.AspNetCore.Dispatcher/Patterns/InlineRouteParameterParser.cs
```diff
@@ -126,7 +116,7 @@ public static TemplatePart ParseRouteParameter(string routeParameter)
                             case null:
                                 state = ParseState.End;
                                 var constraintText = routeParameter.Substring(startIndex, currentIndex - startIndex);
-                                inlineConstraints.Add(new InlineConstraint(constraintText));
+                                inlineConstraints.Add(ConstraintReference.Create(constraintText));
```
jbagga: What's the benefit of this pattern of .Create vs the ctor/new pattern?

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternPart.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        public static RoutePatternLiteral CreateLiteral(string content)
+        {
+            if (string.IsNullOrEmpty(content))
+            {
+                throw new ArgumentException(Resources.Argument_NullOrEmpty, nameof(content));
+            }
+
+            if (content.IndexOf('?') >= 0)
```
jbagga: Why is this looking for '?' specifically?

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternPart.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        public static RoutePatternParameter CreateParameter(
+            string rawText,
+            string name,
+            object defaultValue,
+            RoutePatternParameterKind parameterKind,
+            params ConstraintReference[] constraints)
+        {
+
```
jbagga: nit: empty line

src/Microsoft.AspNetCore.Dispatcher/Patterns/InlineRouteParameterParser.cs
rynowak: yeah that's a good point, there are probably plenty of variable names to clean up

src/Microsoft.AspNetCore.Dispatcher/Patterns/InlineRouteParameterParser.cs
rynowak: There are two reasons:
- Gives you the flexibility to return a subclass in the future if the design changes
- Can come up with more meaningful names (Look at update) when there are multiple ways to create something
- Consistency with one of the other parts of the system that does similar things, but for a better reason

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternPart.cs
rynowak: `?`  is the delimiter for a query string - you can't put that in a path segment

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternPart.cs
rynowak: 👍

PR #479, opened 2017-10-20
Port IRouteConstraint to the dispatcher project

rynowak:
This is a rough cut of a new abstraction for constraints. We may need to do more here but this is needed to unblock some other work


jbagga:
Looks good :shipit:

PR #482, opened 2017-10-21
Add RoutePatternMatcher to Dispatcher

jbagga:
Add TemplateMatcher logic to the dispatcher

Part of #472


rynowak:
No changes to `TemplateMatcher` in routing? How does it call through to the new one?

jbagga:
@rynowak I was going to that in the next PR. I can work on that if it makes more sense to combine them

rynowak:
No that's fine :shipit: for this one then

PR #483, opened 2017-10-23
Port TemplateBinder to dispatcher

rynowak:


src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -41,320 +33,29 @@ public class TemplateBinder
...
-            _defaults = defaults;
-
-            // Any default that doesn't have a corresponding parameter is a 'filter' and if a value
-            // is provided for that 'filter' it must match the value in defaults.
-            _filters = new RouteValueDictionary(_defaults);
-            foreach (var parameter in _template.Parameters)
-            {
-                _filters.Remove(parameter.Name);
-            }
+            _binder = new RoutePatternBinder(urlEncoder, pool, template.ToRoutePattern(), defaults);
```
rynowak: /cc @eilon - let's talk about this one

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -21,10 +22,6 @@ namespace Microsoft.AspNetCore.Routing.Tree
     /// </summary>
     public class TreeRouter : IRouter
     {
-        // Key used by routing and action selection to match an attribute route entry to a
-        // group of action descriptors.
-        public static readonly string RouteGroupKey = "!__route_group";
-
```
jbagga: Why is this safe to delete?

rynowak: OH! - I deleted it from the wrong spot. Will fix.

jbagga: It wasn't being used in TreeRouter anyway though. Perhaps it can stay deleted

rynowak: it's public in a public class, it needs to stay.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
rynowak: Discussed in person, @Eilon doesn't have concerns about this.

PR #484, opened 2017-10-24
Use RoutePatternMatcher logic in TemplateMatcher

jbagga:
Addresses part of #472


src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternMatcher.cs
```diff
@@ -382,7 +408,12 @@ private bool SavePathSegmentsAsValues(int index, DispatcherValueCollection value
                     // This check is related to the check we do at the very end of this function.
                     if (indexOfLastSegmentUsed == (routeSegment.Parts.Count - 1))
                     {
-                        if ((indexOfLiteral + part.RawText.Length) != requestSegment.Length)
+                        if (part is RoutePatternLiteral literal && ((indexOfLiteral + literal.Content.Length) != requestSegment.Length))
+                        {
+                            return false;
+                        }
+
+                        if (part is RoutePatternSeparator separator && ((indexOfLiteral + separator.Content.Length) != requestSegment.Length))
```
rynowak: This should be an else if - just a clarity thing, it won't change the behaviour

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternMatcher.cs
```diff
@@ -422,7 +453,20 @@ private bool SavePathSegmentsAsValues(int index, DispatcherValueCollection value
                         }
                         else
                         {
-                            parameterStartIndex = newLastIndex + lastLiteral.RawText.Length;
```
rynowak: Can you do me a favour before you check this in? Do a find-all-references for `RawText`. Are there any other uses?

src/Microsoft.AspNetCore.Routing/Template/TemplateMatcher.cs
```diff
@@ -70,387 +68,9 @@ public bool TryMatch(PathString path, RouteValueDictionary values)
...
-                {
-                    values.Add(item.Key, item.Value);
-                }
-
-                return true;
-            }
-
-            return false;
+            var routePattern = Template.ToRoutePattern();
+            var routePatternMatcher = new RoutePatternMatcher(routePattern, Defaults);
```
rynowak: This should be done in the constructor and cached.

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternMatcher.cs
jbagga: I thought I did that. I'll check again

jbagga: Is this ok? Or I shouldn't use it at all? https://github.com/aspnet/Routing/blob/037e771e9ea0adb523cdee703aaa00df22304166/src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternMatcher.cs#L387

jbagga:
@rynowak Updated

PR #485, opened 2017-10-26
Add Template abstraction

rynowak:
This change adds the Template as a top level abstraction. URL templating
is now a two-stage process.

First you use a 'key' to look up a Template, then you use the Template
to create the URL.

This change also has some cleanup of the way RoutePatternBinder gets
instantiated. I added a factory service so that most of the complex
things can be made internal to Dispatcher. Now it's much easier to
constuct and use. These impacts some pubternal APIs that we already
broke, but makes them actually nice :)

Also cleaned up some tests and fixed one that was broken and not
running.


test/Microsoft.AspNetCore.Routing.Tests/Constraints/DateTimeRouteConstraintTests.cs
```diff
@@ -36,9 +23,7 @@ public static IEnumerable<object[]> GetDateTimeObject
         [InlineData("12/25/2009 11:45:00 PM", true)]
         [InlineData("2009-05-12T11:45:00Z", true)]
         [InlineData("not-parseable-as-date", false)]
-        [InlineData(false, false)]
-        [MemberData(nameof(GetDateTimeObject))]
-        public void DateTimeRouteConstraint(object parameterValue, bool expected)
+        public void DateTimeRouteConstraint_ParsesStrings(string parameterValue, bool expected)
```
rynowak: @jbagga - this was the test that always got stuck. I think the problem was mixing `InlineData` and `MemberData`.

PR #487, opened 2017-10-27
Add IConstraintFactory

jbagga:
Add ConstraintResolver infrastructure to Dispatcher to enable porting `TreeMatcher`

Part of #472

PR #488, opened 2017-11-02
Port TreeMatcher

jbagga:
This PR builds on top of PR #487 

Addresses issue #472


src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
```diff
@@ -0,0 +1,547 @@
...
+                var templateEndpoint = endpoint as IRoutePatternEndpoint;
+                if (templateEndpoint == null)
+                {
+                    continue;
+                }
+
+                if (!groups.TryGetValue(new Key(0, templateEndpoint.Pattern), out var group))
+                {
+                    group = new List<Endpoint>();
+                    groups.Add(new Key(0, templateEndpoint.Pattern), group);
```
jbagga: @rynowak This forces the order to 0 so for now none of the tests for `TreeMatcher` check that order is respected.

benchmarks/Microsoft.AspNetCore.Dispatcher.Performance/DispatcherBenchmark.cs
```diff
@@ -19,25 +15,25 @@ public class DispatcherBenchmark
...
+            var dataSource = new DefaultDispatcherDataSource()
+            {
+                Endpoints =
+                {
+                    new RoutePatternEndpoint("api/Widgets", Benchmark_Delegate),
+                    new RoutePatternEndpoint("api/Widgets/{id}", Benchmark_Delegate),
+                    new RoutePatternEndpoint("api/Widgets/search/{term}", Benchmark_Delegate),
+                    new RoutePatternEndpoint("admin/users/{id}", Benchmark_Delegate),
+                    new RoutePatternEndpoint("admin/users/{id}/manage", Benchmark_Delegate),
+                },
```
rynowak: Yay! This is off to an auspicious start.

src/Microsoft.AspNetCore.Dispatcher/Constraints/AlphaDispatcherValueConstraint.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Dispatcher
+{
+    /// <summary>
+    /// Constrains a dispatcher value parameter to contain only lowercase or uppercase letters A through Z in the English alphabet.
+    /// </summary>
+    public class AlphaDispatcherValueConstraint : RegexDispatcherValueConstraint
```
rynowak: ported because it's used in tests? Seems fine for now. We will eventually redesign some of the constraints, but this is a good thing to do if it unblocks you

src/Microsoft.AspNetCore.Dispatcher/Constraints/DefaultConstraintFactory.cs
```diff
@@ -119,16 +118,27 @@ private static IDispatcherValueConstraint CreateConstraint(Type constraintType,
...
+                        {
+                            activationConstructor = constructor;
+                            parameters = ConvertArguments(activationConstructor.GetParameters(), arguments);
+                            return (IDispatcherValueConstraint)activationConstructor.Invoke(parameters);
+                        }
+                        catch
+                        {
+                            continue;
+                        }
+                    }
```
rynowak: This seems suspect. Let's discuss when we find time.

src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
rynowak: hmm. Is the problem here that an endpoint has no way to convey an 'order'?

rynowak: How about adding a metadata that specifies the order?

src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
```diff
@@ -0,0 +1,547 @@
...
+        }
+
+        public int Version { get; private set; }
+
+        public IConstraintFactory ConstraintFactory { get; set; }
+
+        /// <summary>
+        /// Gets the list of <see cref="InboundRouteEntry"/>.
+        /// </summary>
+        public IList<InboundRouteEntry> InboundEntries { get; } = new List<InboundRouteEntry>();
```
rynowak: These public properties are a bit suspect to me

src/Microsoft.AspNetCore.Dispatcher/Constraints/AlphaDispatcherValueConstraint.cs
jbagga: Yeah, I only ported the ones that are used in tests.

src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
jbagga: Will it require an abstraction like `IEndpointOrderMetadata` (or similar) so `GetMetadata` can retrieve it?

jbagga:
@rynowak Updated. I attempted to address the issues, let me know what you think!

src/Microsoft.AspNetCore.Dispatcher.Abstractions/IEndpointOrderMetadata.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Dispatcher
```
rynowak: These both should be in the `Microsoft.AspNetCore.Dispatcher` project

src/Microsoft.AspNetCore.Dispatcher.Abstractions/IEndpointOrderMetadata.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public interface IEndpointOrderMetadata
+    {
+        int Order { get; }
+    }
+
+    public class EndpointOrderMetadata : IEndpointOrderMetadata
```
rynowak: This gets its own file

PR #490, opened 2017-11-08
Add package descriptions

natemcmaster:
Same thing as https://github.com/aspnet/HttpClientFactory/pull/24. Required to unblock moving forward to the 2.2 SDK.

These are definitely placeholder descriptions. Not sure what these packages do.

### DIFF
--- a/src/Microsoft.AspNetCore.Dispatcher.Abstractions/Microsoft.AspNetCore.Dispatcher.Abstractions.csproj
+++ b/src/Microsoft.AspNetCore.Dispatcher.Abstractions/Microsoft.AspNetCore.Dispatcher.Abstractions.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <Description>Abstractions for dispatching.</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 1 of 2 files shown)


Eilon:
@rynowak can you provide some descriptions?

src/Microsoft.AspNetCore.Dispatcher.Abstractions/Microsoft.AspNetCore.Dispatcher.Abstractions.csproj
```diff
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <Description>Abstractions for dispatching.</Description>
```
Eilon: Can we also add some tags while we're at it?

Some examples:
* https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing/Microsoft.AspNetCore.Routing.csproj#L4-L11
* https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing.Abstractions/Microsoft.AspNetCore.Routing.Abstractions.csproj#L4-L11

Eilon: (These are for @rynowak to provide.)

Eilon:
@natemcmaster how about you merge this and I'll file a follow-up bug to get final text.

natemcmaster:
Sounds good to me 👍

PR #492, opened 2017-11-08
[Design] Move UrlMatchingTree to shared source

jbagga:
WIP for refactoring the Tree infrastructure. Instead of #if/#else blocks, we may instead consider using generic types

cc @rynowak


shared/Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources/UrlMatchingNode.cs
```diff
@@ -51,31 +81,26 @@ public UrlMatchingNode(int length)
 
         /// <summary>
         /// Gets or sets the <see cref="UrlMatchingNode"/> representing
-        /// parameter segments with constraints following this segment in the <see cref="TreeRouter"/>.
+        /// parameter segments with constraints following this segment.
```
jbagga: I thought about removing comments but in Routing this API is public.

Routing.sln
```diff
@@ -57,7 +57,15 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DispatcherSample", "samples
...
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Dispatcher.FunctionalTest", "test\Microsoft.AspNetCore.Dispatcher.FunctionalTest\Microsoft.AspNetCore.Dispatcher.FunctionalTest.csproj", "{32107601-C9BE-467B-894C-C9F2E35F03E4}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "shared", "shared", "{2770E573-3E34-4FCE-A3D5-C6CDD7B9A9CC}"
+	ProjectSection(SolutionItems) = preProject
+		shared\Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources\InboundMatch.cs = shared\Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources\InboundMatch.cs
+		shared\Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources\InboundRouteEntry.cs = shared\Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources\InboundRouteEntry.cs
+		shared\Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources\UrlMatchingNode.cs = shared\Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources\UrlMatchingNode.cs
+		shared\Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources\UrlMatchingTree.cs = shared\Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources\UrlMatchingTree.cs
+	EndProjectSection
```
rynowak: I'm really not sure what this is. We don't do it anywhere else.

shared/Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources/InboundMatch.cs
```diff
@@ -2,21 +2,37 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Diagnostics;
+#if UrlMatching_InRouting
```
rynowak: Just simplify these to `#if ROUTING` - `#if DISPATCHER`

shared/Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources/InboundMatch.cs
```diff
@@ -2,21 +2,37 @@
...
 using Microsoft.AspNetCore.Routing.Template;
+#else
+using Microsoft.AspNetCore.Dispatcher;
+#endif
 
+#if UrlMatching_InRouting
 namespace Microsoft.AspNetCore.Routing.Tree
+#elif UrlMatching_InDispatcher
+namespace Microsoft.AspNetCore.Dispatcher.Internal
```
rynowak: Don't use the `.Internal` namespace in dispatcher, we don't need it.

shared/Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources/InboundMatch.cs
```diff
@@ -2,21 +2,37 @@
...
+#else
+#error
+#endif
 {
+#if UrlMatching_InRouting
     /// <summary>
     /// A candidate route to match incoming URLs in a <see cref="TreeRouter"/>.
     /// </summary>
     [DebuggerDisplay("{DebuggerToString(),nq}")]
```
rynowak: Can we get the debuggerdisplay in both?

shared/Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources/InboundRouteEntry.cs
```diff
@@ -44,18 +51,53 @@ public class InboundRouteEntry
         public decimal Precedence { get; set; }
 
         /// <summary>
+        /// Gets or sets an arbitrary value associated with the entry.
+        /// </summary>
+        public object Tag { get; set; }
```
rynowak: This is where generics would help.

shared/Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources/UrlMatchingNode.cs
rynowak: That's fine

Routing.sln
jbagga: https://github.com/aspnet/StaticFiles/blob/dev/StaticFiles.sln#L31
These will show up in VS if added here. Can be removed though

shared/Microsoft.AspNetCore.Routing.UrlMatchingTree.Sources/InboundRouteEntry.cs
rynowak: We should do something better with this.

rynowak: I think this should be array of endpoint for dispatcher

jbagga:
@rynowak Updated

Routing.sln
rynowak: They should show up regardless though right? They show up under each project where they are used

jbagga: Yes they do. This also adds them to the shared directory which shows up in VS. Not really required so I deleted them

PR #497, opened 2017-11-20
Use previous version of Appveyor image to avoid dotnet bug

jbagga:
See https://github.com/aspnet/Mvc/pull/7069

Temporary change until the fix ships in the framework


jbagga:
Work item to undo when the fix is shipped https://github.com/aspnet/Routing/issues/498

Issue #441, opened 2017-09-09
Tolerate a leading / or ~/ on conventional routes


rynowak:
We throw an exception today if you do this, but we allow it in attribute routes. We should allow this.

When we generate a url using a route like `"~/foo"` the resulting url should be "/foo" not `"~/foo"`

jbagga:
This issue was moved to aspnet/Home#2765

PR #499, opened 2017-11-21
Tolerate leading "~/" or "/"

jbagga:
Addresses #441

### DIFF
--- a/src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternBuilder.cs
+++ b/src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternBuilder.cs
@@ -58,8 +58,7 @@ namespace Microsoft.AspNetCore.Dispatcher.Patterns
                 var segment = PathSegments[i];
                 for (var j = 0; j < segment.Parts.Count; j++)
                 {
-                    var parameter = segment.Parts[j] as RoutePatternParameter;
-                    if (parameter != null)
+                    if (segment.Parts[j] is RoutePatternParameter parameter)
                     {
                         parameters.Add(parameter);
                     }
@@ -71,11 +70,6 @@ namespace Microsoft.AspNetCore.Dispatcher.Patterns
 
         public static RoutePatternBuilder Create(string text)
         {
-            if (text.StartsWith("~/"))
-            {
-                return new RoutePatternBuilder() { RawText = text.Substring(1), };
-            }
-
             return new RoutePatternBuilder() { RawText = text, };
         }
     }
--- a/src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternBuilder.cs
+++ b/src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternBuilder.cs
@@ -71,6 +71,11 @@ namespace Microsoft.AspNetCore.Dispatcher.Patterns
 
         public static RoutePatternBuilder Create(string text)
         {
+            if (text.StartsWith("~/"))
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternBuilder.cs
```diff
@@ -71,6 +71,11 @@ public RoutePattern Build()
 
         public static RoutePatternBuilder Create(string text)
         {
+            if (text.StartsWith("~/"))
+            {
+                return new RoutePatternBuilder() { RawText = text.Substring(1), };
+            }
```
rynowak: The `RawText` should still reflect what the user typed in. I don't think we should have this special case.

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternParser.cs
```diff
@@ -26,20 +26,16 @@ internal static class RoutePatternParser
             Asterisk
         };
 
-        public static RoutePattern Parse(string pattern)
+        public static RoutePattern Parse(string originalPattern)
```
rynowak: not digging the rename here

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternParser.cs
```diff
@@ -469,10 +466,21 @@ private static bool IsValidLiteral(TemplateParserContext context, string literal
             return true;
         }
 
-        private static bool IsInvalidPattern(string routeTemplate)
+        private static string PatternStartsWithSlashOrTildeSlash(string routePattern)
```
rynowak: This should be have a smarter name like `TrimPrefix` or `HandleLeadingSlash`. Based on the naming of this I expect this to return a boolean

src/Microsoft.AspNetCore.Dispatcher/Resources.resx
```diff
@@ -164,7 +164,7 @@
     <value>The route parameter name '{0}' is invalid. Route parameter names must be non-empty and cannot contain these characters: '{{', '}}', '/'. The '?' character marks a parameter as optional, and can occur only at the end of the parameter. The '*' character marks a parameter as catch-all, and can occur only at the start of the parameter.</value>
   </data>
   <data name="TemplateRoute_InvalidRouteTemplate" xml:space="preserve">
-    <value>The route template cannot start with a '/' or '~' character.</value>
+    <value>The route template cannot start with a '~' character.</value>
```
rynowak: This is a bit misleading, the route template can start with `~` if its followed by a `/`

jbagga: `The route template cannot start with a '~' character unless followed by a "/".`

Is this any better?

src/Microsoft.AspNetCore.Dispatcher/Patterns/RoutePatternParser.cs
jbagga: Is it okay if I use the `trimmedPattern` after trimming the prefix and call this `pattern` again? I just need to make the distinction between the pattern that's the parameter and the one returned after trimming the leading '/' or "~/".

rynowak: Sure, that sounds good.

jbagga:
@rynowak Updated

PR #501, opened 2017-12-01
Use current Visual Studio image

### LINKED ISSUE #498: Use newest version of image in Appveyor
See https://github.com/aspnet/Mvc/issues/7071

ryanbrandenburg:
Fixes #498. Once these appveyor tests pass we'll know we can merge the PR.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -16,4 +16,4 @@ environment:
 test: off
 deploy: off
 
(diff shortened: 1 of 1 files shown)


ryanbrandenburg:
Considering that both available versions of the image are now broken I suggest that we merge this anyway. It will at least get us to where we'll start passing as soon as they fix it.

PR #503, opened 2017-12-14
Recreate cache of endpoints when data source updated

jbagga:
Addresses https://github.com/aspnet/Home/issues/2663

### DIFF
--- a/src/Microsoft.AspNetCore.Dispatcher/CompositeDispatcherDataSource.cs
+++ b/src/Microsoft.AspNetCore.Dispatcher/CompositeDispatcherDataSource.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.Dispatcher
 
         public override IChangeToken ChangeToken { get; }
 
-        protected override IReadOnlyList<Address> GetAddesses()
+        protected override IReadOnlyList<Address> GetAddresses()
         {
             var addresses = new List<Address>();
             for (var i = 0; i < _dataSources.Length; i++)
--- a/src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
+++ b/src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -30,8 +29,6 @@ namespace Microsoft.AspNetCore.Dispatcher
             _initializer = CreateCache;
         }
 
-        public int Version { get; private set; }
-
         public override async Task MatchAsync(MatcherContext context)
         {
             if (context == null)
@@ -40,6 +37,7 @@ namespace Microsoft.AspNetCore.Dispatcher
             }
 
             EnsureServicesInitialized(context);
+            Extensions.Primitives.ChangeToken.OnChange(() => ChangeToken, () => Volatile.Write(ref _dataInitialized, false));
 
             var cache = LazyInitializer.EnsureInitialized(ref _cache, ref _dataInitialized, ref _lock, _initializer);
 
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
```diff
@@ -40,6 +37,7 @@ public TreeMatcher()
             }
 
             EnsureServicesInitialized(context);
+            Extensions.Primitives.ChangeToken.OnChange(() => ChangeToken, () => Volatile.Write(ref _dataInitialized, false));
```
jbagga: I tried to do this in the ctor but the `DataSource` is set after the ctor in `TreeMatcherFactory` and `ChangeToken` was null as a result

src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
```diff
@@ -136,8 +134,7 @@ internal Cache CreateCache()
             {
                 var endpoint = endpoints[i];
 
-                var templateEndpoint = endpoint as IRoutePatternEndpoint;
-                if (templateEndpoint == null)
+                if (!(endpoint is IRoutePatternEndpoint templateEndpoint))
```
jbagga: We can thank @pranavkm for this

test/Microsoft.AspNetCore.Dispatcher.Test/Tree/TreeMatcherTest.cs
```diff
@@ -776,6 +780,32 @@ public void MatchAsync_DoesNotMatchShorterUrl()
...
+
+            // Arrange 2
+            dataSource.UpdateEndpoints(new List<Endpoint>() { new RoutePatternEndpoint("test2/{p1}", new object(), Test_Delegate, "Test2"), });
+            context = CreateMatcherContext("/test2/parameter2");
+
+            // Act 2
+            await matcher.MatchAsync(context);
+
+            // Assert 2
+            Assert.Equal("Test2", context.Endpoint.DisplayName);
```
jbagga: I wanted to check if context.Values are populated with the key "p1" with value "parameter2" but context.Values is an empty collection. Is that expected?

src/Microsoft.AspNetCore.Dispatcher/CompositeDispatcherDataSource.cs
```diff
@@ -32,7 +32,7 @@ public CompositeDispatcherDataSource(IEnumerable<DispatcherDataSource> dataSourc
 
         public override IChangeToken ChangeToken { get; }
 
-        protected override IReadOnlyList<Address> GetAddesses()
```
rynowak: lol

src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
rynowak: You could do this in `EnsureServicesInitialized`. You can't really do it here, it's going to add another delegate every request.

rynowak: Can you add a using for this?

src/Microsoft.AspNetCore.Dispatcher/Tree/TreeMatcher.cs
rynowak: cool beans

test/Microsoft.AspNetCore.Dispatcher.Test/Tree/TreeMatcherTest.cs
rynowak: No, that doesn't seem right.

test/Microsoft.AspNetCore.Dispatcher.Test/Tree/TreeMatcherTest.cs
```diff
@@ -804,5 +834,65 @@ private static Task Test_Delegate(HttpContext httpContext)
...
+                _changeToken = new TestChangeToken();
+            }
+
+            public void UpdateEndpoints(IList<Endpoint> endpoints)
+            {
+                lock (_lock)
+                {
+                    _endpoints = endpoints;
+                    // Trigger change token here
+                    var previousToken = Interlocked.Exchange(ref _changeToken, new TestChangeToken());
```
rynowak: The interlocked isn't really needed here because it's protected by a lock.

jbagga:
@rynowak  Updated

PR #518, opened 2018-05-06
IUrlHelper and support for RouteOptions.LowercaseUrls

kishanAnem:
IUrlHelper and support for RouteOptions.LowercaseUrls

#issue : https://github.com/aspnet/Mvc/issues/7720

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/RouteCollection.cs
+++ b/src/Microsoft.AspNetCore.Routing/RouteCollection.cs
@@ -153,6 +153,11 @@ namespace Microsoft.AspNetCore.Routing
                 if (_options.LowercaseUrls)
                 {
                     urlWithoutQueryString = urlWithoutQueryString.ToLowerInvariant();
+
+                    if (!string.IsNullOrEmpty(queryString))
+                    {
+                        queryString = queryString.ToLowerInvariant();
+                    }
                 }
 
                 if (_options.AppendTrailingSlash && !urlWithoutQueryString.EndsWith("/"))
--- a/src/Microsoft.AspNetCore.Routing/RouteCollection.cs
+++ b/src/Microsoft.AspNetCore.Routing/RouteCollection.cs
@@ -138,9 +138,9 @@ namespace Microsoft.AspNetCore.Routing
 
             var url = pathData.VirtualPath;
 
-            if (!string.IsNullOrEmpty(url) && (_options.LowercaseUrls || _options.AppendTrailingSlash))
+            if (!string.IsNullOrEmpty(url) && (_options.LowercaseUrls || _options.LowercaseQueryStrings || _options.AppendTrailingSlash))
             {
-                var indexOfSeparator = url.IndexOfAny(new char[] { '?', '#' });
+                var indexOfSeparator = url.IndexOfAny(UrlQueryDelimiters);
                 var urlWithoutQueryString = url;
                 var queryString = string.Empty;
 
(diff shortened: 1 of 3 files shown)


rynowak:
From the issue:

> if you'd like to submit a PR in the Routing repo to handle this with **a separate option** we'd be something we'd consider to take.

This needs to be a separate option or else it's a breaking change.

kishan3034:
@rynowak this is a enhancement. it is not breaking anything.
expected result: query?property=x
actual result: query?propertySomething=x

kishan3034:
hi @rynowak,

please guide me.im interested to work.

rynowak:
Start be adding another property to `RouteOptions` - `bool LowercaseQueryStrings` and then make the new behavior respect that setting. Also you should update the documentation for `LowercaseUrls` to point out that it doesn't include the query string.

kishan3034:
thanks @rynowak.
i’ll do @rynowak.

kishanAnem:
Hi @rynowak,
I made changes. Please review.

src/Microsoft.AspNetCore.Routing/RouteCollection.cs
```diff
@@ -138,9 +138,9 @@ private VirtualPathData NormalizeVirtualPath(VirtualPathData pathData)
 
             var url = pathData.VirtualPath;
 
-            if (!string.IsNullOrEmpty(url) && (_options.LowercaseUrls || _options.AppendTrailingSlash))
+            if (!string.IsNullOrEmpty(url) && (_options.LowercaseUrls || _options.LowercaseQueryStrings || _options.AppendTrailingSlash))
```
rynowak: `LowercaseQueryStrings` shouldn't appear here, that just applies to the query string.

src/Microsoft.AspNetCore.Routing/RouteOptions.cs
```diff
@@ -10,10 +10,15 @@ namespace Microsoft.AspNetCore.Routing
     public class RouteOptions
     {
         /// <summary>
-        /// Gets or sets a value indicating whether all generated URLs are lower-case.
+        /// Gets or sets a value indicating whether all generated URLs are lower-case. it doesn't include the query string.
```
rynowak: Suggestions for these:

`Gets or sets a value indicating whether all generated paths URLs are lower-case. Use <see cref="LowercaseQueryStrings" /> to configure the behavior for query strings.`

src/Microsoft.AspNetCore.Routing/RouteOptions.cs
```diff
@@ -10,10 +10,15 @@ namespace Microsoft.AspNetCore.Routing
...
         /// </summary>
         public bool LowercaseUrls { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether all generated QUERY STRINGS are lower-case. property will when LowercaseUrls true.
+        /// </summary>
+        public bool LowercaseQueryStrings { get; set; }
```
rynowak: Suggesting for this:
`Gets or sets a value indicating whether a generated query strings are lower-case. This property will be unless <see cref="LowercaseUrls" /> is also <c>true</c>`

test/Microsoft.AspNetCore.Routing.Tests/RouteCollectionTest.cs
```diff
@@ -682,12 +701,13 @@ private static INamedRouter CreateNamedRoute(string name, bool accept = false, s
 
         private static Action<RouteOptions> GetRouteOptions(
             bool lowerCaseUrls = false,
-            bool appendTrailingSlash = false)
+            bool appendTrailingSlash = false, bool lowercaseQueryStrings = false)
```
rynowak: please put the new parameter on it's own line. This makes diffs nicer in the future.

kishanAnem:
Hi @rynowa,
Thank you very much for the support. 
You are guiding me step by step.

PR #520, opened 2018-05-31
Code dump of dispatcher prototype code

rynowak:
Here's a code dump of the parts of the Dispatcher prototype codebase
that are needed to get us off the ground.

This first cut attempts to use part of routing where possible, and not
all of those changes will be long-lasting.

I'll leave comments through thoughout the PR for education.


samples/DispatcherSample.Web/Startup.cs
```diff
@@ -0,0 +1,46 @@
...
+                    {
+                        var response = httpContext.Response;
+                        var payloadLength = _helloWorldPayload.Length;
+                        response.StatusCode = 200;
+                        response.ContentType = "text/plain";
+                        response.ContentLength = payloadLength;
+                        return response.Body.WriteAsync(_helloWorldPayload, 0, payloadLength);
+                    },
+                    "/plaintext", new { }, 0, EndpointMetadataCollection.Empty, "Plaintext"),
+                }));
```
rynowak: This is intentionally kinda ugly. 

The plan for 2.2 is that there is no mechanism for **application developers** to add endpoints (other than MVC). We can add endpoints using these kinds of a primitives, but it's not meant to be easier or more fun than it is right here.

This is because we're not trying to make an alternative to MVC or middleware

samples/DispatcherSample.Web/Startup.cs
```diff
@@ -0,0 +1,46 @@
...
+            });
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            app.UseDispatcher();
+
+            // Imagine some more stuff here...
+
+            app.UseEndpoint();
```
rynowak: Let's not get hung up on names at this juncture. All of this is subject to change.

samples/RoutingSample.Web/Program.cs
```diff
@@ -28,23 +20,7 @@ public static IWebHostBuilder GetWebHostBuilder()
...
-
-                    routes.MapGet("api/get/{id}", (request, response, routeData) => response.WriteAsync($"API Get {routeData.Values["id"]}"))
-                          .MapMiddlewareRoute("api/middleware", (appBuilder) => appBuilder.Use((httpContext, next) => httpContext.Response.WriteAsync("Middleware!")))
-                          .MapRoute(
-                            name: "AllVerbs",
-                            template: "api/all/{name}/{lastName?}",
-                            defaults: new { lastName = "Doe" },
-                            constraints: new { lastName = new RegexRouteConstraint(new Regex("[a-zA-Z]{3}", RegexOptions.CultureInvariant, RegexMatchTimeout)) });
-                }));
+                .UseStartup<Startup>();
```
rynowak: I'm not sure why this code wasn't using `Startup`. I fixed it because why not.

samples/RoutingSample.Web/Startup.cs
```diff
@@ -0,0 +1,43 @@
...
+                routes.MapGet("api/get/{id}", (request, response, routeData) => response.WriteAsync($"API Get {routeData.Values["id"]}"))
+                      .MapMiddlewareRoute("api/middleware", (appBuilder) => appBuilder.Use((httpContext, next) => httpContext.Response.WriteAsync("Middleware!")))
+                      .MapRoute(
+                        name: "AllVerbs",
+                        template: "api/all/{name}/{lastName?}",
+                        defaults: new { lastName = "Doe" },
+                        constraints: new { lastName = new RegexRouteConstraint(new Regex("[a-zA-Z]{3}", RegexOptions.CultureInvariant, RegexMatchTimeout)) });
+            });
+        }
+    }
```
rynowak: Just a copy-paste

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            // Both allowed to be null
+            Metadata = metadata ?? EndpointMetadataCollection.Empty;
+            DisplayName = displayName;
+        }
+
+        public string DisplayName { get; }
+
+        public EndpointMetadataCollection Metadata { get; }
+    }
```
rynowak: This class as well as the feature interface will need to go to the HTTP repo once we're slightly more baked. Note that this just has enough info for something like a middleware to introspect metadata. 

The separation between this and `MatcherEndpoint` is intentional and allows middleware to consume this API without a dependency on too many routing concepts.

src/Microsoft.AspNetCore.Routing.Abstractions/IEndpointFeature.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public interface IEndpointFeature
+    {
+        Endpoint Endpoint { get; set; }
+
+        Func<RequestDelegate, RequestDelegate> Invoker { get; set; }
+
+        RouteValueDictionary Values { get; set; }
```
rynowak: @davidfowl do you have thoughts on whether you want `RVD` to move to the HTTP repo or should we change  this to be `IDictionary<string, object>`?

src/Microsoft.AspNetCore.Routing.Abstractions/MetadataCollection.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            for (var i = 0; i < _items.Length; i++)
+            {
+                var item = _items[i] as T;
+                if (item != null)
+                {
+                    yield return item;
+                }
+            }
+        }
```
rynowak: @kichalla - this will look somewhat familiar to you. It's a parallel of https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/FilterContext.cs#L84

rynowak: We can polish this more in the future as we actually us it.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -0,0 +1,70 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class CompositeEndpointDataSource : EndpointDataSource
```
rynowak: This is public so that you can get it from DI. We don't do that it in the implementation of the middleware because the initialization semantics are different.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -0,0 +1,70 @@
...
+        private void Initialize()
+        {
+            lock (_lock)
+            {
+                _changeToken = new CompositeChangeToken(_dataSources.Select(d => d.ChangeToken).ToArray());
+                _endpoints = _dataSources.SelectMany(d => d.Endpoints).ToArray();
+
+                _changeToken.RegisterChangeCallback((state) => Initialize(), null);
+            }
+        }
```
rynowak: Note that there's some subtle and non-trivial things going on with initialization and change tokens here. I tried to add comments where it's needed. let me know where more explanation is needed.

src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+            _logger = logger;
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            var feature = new EndpointFeature();
+            httpContext.Features.Set<IEndpointFeature>(feature);
+
+            // There's an inherit race condition between waiting for init and accessing the matcher
```
rynowak: inherent. DUH

src/Microsoft.AspNetCore.Routing/Internal/DataSourceDependantCache.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Microsoft.AspNetCore.Routing.Internal
+{
+    internal class DataSourceDependantCache<T> where T : class
```
rynowak: I predict that we'll have plenty of cases where we want to use something like this...

Routing data can change, but rarely does, and it's a push model. We want to always be able to synchronously get the current data, but be able to reinitialize in the background when a change occurs.

src/Microsoft.AspNetCore.Routing/Internal/DataSourceDependantCache.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            _dataSource = dataSource;
+            _initializeCore = initialize;
+
+            _initializer = Initialize;
+            _initializerWithState = (state) => Initialize();
+            _lock = new object();
+        }
+
+        public T Value => _value;
```
rynowak: Note that we don't lock here, and think about that in the context of a 'push'. So when data gets 'pushed' we start computing a new state, but we're still able to perform operations on the old state until we've processed the update.

src/Microsoft.AspNetCore.Routing/Matchers/IMatcherFactory.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    internal abstract class MatcherFactory
```
rynowak: Will rename file to match

src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
```diff
@@ -0,0 +1,348 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing.Internal;
+using Microsoft.AspNetCore.Routing.Template;
+using Microsoft.AspNetCore.Routing.Tree;
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    internal class TreeMatcher : Matcher
```
rynowak: This is a temporary implementation based on the tree router. This should be enough to get us started.

src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
```diff
@@ -0,0 +1,348 @@
...
+                    }
+                }
+            }
+
+            return true;
+        }
+
+        private Task SelectEndpointAsync(HttpContext httpContext, IEndpointFeature feature, IReadOnlyList<MatcherEndpoint> endpoints)
+        {
+            switch (endpoints.Count)
```
rynowak: Note that this code doesn't do anything significant now. This will eventually incorporate something like `IActionConstraint`.

src/Microsoft.AspNetCore.Routing/Tree/InboundRouteEntry.cs
```diff
@@ -52,5 +52,7 @@ public class InboundRouteEntry
         /// Gets or sets the <see cref="RouteTemplate"/>.
         /// </summary>
         public RouteTemplate RouteTemplate { get; set; }
+
+        public object Tag { get; set; }
```
rynowak: Hack so we can reuse the tree router

src/Microsoft.AspNetCore.Routing/Tree/TreeRouteBuilder.cs
```diff
@@ -288,167 +288,5 @@ public void Clear()
             InboundEntries.Clear();
             OutboundEntries.Clear();
         }
-
-        private void AddEntryToTree(UrlMatchingTree tree, InboundRouteEntry entry)
```
rynowak: Had to refactor this slightly so we could reuse the tree router

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -233,108 +231,6 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
             }
         }
 
-        private struct TreeEnumerator : IEnumerator<UrlMatchingNode>
```
rynowak: Changed to a top-level class so we could reuse it

src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
```diff
@@ -0,0 +1,348 @@
...
+            }
+
+            public override bool Equals(object obj)
+            {
+                return obj is Key ? Equals((Key)obj) : false;
+            }
+
+            public override int GetHashCode()
+            {
+                var hash = new HashCodeCombiner();
```
JamesNK: First time I've seen someone use this 👍

src/Microsoft.AspNetCore.Routing/Properties/AssemblyInfo.cs
```diff
@@ -4,3 +4,4 @@
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Microsoft.AspNetCore.Routing.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
+[assembly: InternalsVisibleTo("DispatcherSample.Web, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
JamesNK: Hmm?

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
```diff
@@ -179,7 +177,7 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
                 var tokenizer = new PathTokenizer(context.HttpContext.Request.Path);
                 var root = tree.Root;
 
-                var treeEnumerator = new TreeEnumerator(root, tokenizer);
+                var treeEnumerator = new Treenumerator(root, tokenizer);
```
JamesNK: `E` shortage?

JamesNK:
Should this go into dev or are we (dispatching/endpoint) partying in our own branch?

I'll handle any changes identified.

samples/DispatcherSample.Web/Startup.cs
rynowak: I'm open to more discussion about this but what I'm trying to say is that I think it's specifically a non goal. We don't have the other tools that users will need to build real applications in an expressjs style.

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
rynowak: Yes

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
rynowak: I think it should be real public. This is what you'd use if you want to programmatically enumerate all of the endpoints in the application.

src/Microsoft.AspNetCore.Routing/Internal/DataSourceDependantCache.cs
rynowak: It's a new thing.

The route table and set of endpoints are fixed, it can be updated while the application is running. Razor pages is an example of this. You can 'drop in' a page to a running application and it will compile the page and add the appropriate entry to the route table.

src/Microsoft.AspNetCore.Routing/Internal/DataSourceDependantCache.cs
rynowak: Sure

samples/DispatcherSample.Web/Startup.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace DispatcherSample.Web
+{
+    public class Startup
+    {
+        private static readonly byte[] _helloWorldPayload = Encoding.UTF8.GetBytes("Hello, World!");
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddRouting();
```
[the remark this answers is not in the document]
rynowak: It needs to for now since it will use some services that are included by routing. I think whether or not we have `AddRouting` + `AddXyz` will depend on the final naming and how much they share when we're further along. 

I'd lean towards having a single method... which would be `AddRouting` - the cost of 'extra' services you don't use is basically nothing.

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Routing
```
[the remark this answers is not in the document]
rynowak: So, the background on this is that @davidfowl likes the name 'Dispatcher' so much that he wants to take it from me and use it on something deep in the stack that no one will ever understand. 

It seems very likely to me that we'll end up branding all of these new features to be similar to routing in terms of naming, so I've been putting things in the routing namespace.

Routing is already a namespace you don't reach for often, so I think it's fine if it attracts some clutter. We can pivot on this once we know more.

src/Microsoft.AspNetCore.Routing.Abstractions/MetadataCollection.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public Enumerator GetEnumerator() => new Enumerator(this);
+
+        IEnumerator<object> IEnumerable<object>.GetEnumerator() => GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        public struct Enumerator : IEnumerator<object>
+        {
+            private object[] _items;
```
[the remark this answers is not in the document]
rynowak: This is intentionally not-readonly to prevent defensive copies

src/Microsoft.AspNetCore.Routing.Abstractions/MetadataCollection.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        public IEnumerable<T> GetOrderedMetadata<T>() where T : class
+        {
+            for (var i = 0; i < _items.Length; i++)
+            {
+                var item = _items[i] as T;
+                if (item != null)
+                {
+                    yield return item;
```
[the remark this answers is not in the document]
rynowak: I also expect this code to change a bit. The main usage of this boy is for the kinds of things filters do in MVC. You sometimes want to get all policies of a type, and sometimes want to get the 'best' policy.

I think we'll eventually make this type more dictionary-like, which is the reason for the nominal type rather than IReadOnlyList<T>

src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+            if (_initializationTask != null)
+            {
+                return _initializationTask;
+            }
+
+            var initializationTask = new TaskCompletionSource<Matcher>();
+            if (Interlocked.CompareExchange<Task<Matcher>>(
+                ref _initializationTask,
+                initializationTask.Task,
+                null) == null)
```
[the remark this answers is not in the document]
rynowak: You feel smarter doing it this way.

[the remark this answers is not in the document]
rynowak: Serioustalk: I think of these as different patterns, and so the idea of using a lock didn't even enter my brain. 

This is a latch, it's either initialized or not, but once it's initialized, you never need to think about it again. When I see something like this in code I understand that's how it will behave without having to read much more of the context.

When I see a lock, I assume that something more complicated than a contended initialization is happening.

src/Microsoft.AspNetCore.Routing/EndpointMiddleware.cs
```diff
@@ -0,0 +1,77 @@
...
+                "Executed endpoint '{EndpointName}'.");
+
+            public static void ExecutingEndpoint(ILogger logger, Endpoint endpoint)
+            {
+                _executingEndpoint(logger, endpoint.DisplayName, null);
+            }
+
+            public static void ExecutedEndpoint(ILogger logger, Endpoint endpoint)
+            {
+                _executingEndpoint(logger, endpoint.DisplayName, null);
```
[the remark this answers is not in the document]
rynowak: derp.

src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
rynowak: Oh, it's all over MVC and Razor. 

They are adding something like this to the BCL I think???

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
rynowak: I think it's better this way, but we can change if I'm only one who holds that opinion.

src/Microsoft.AspNetCore.Routing/Properties/AssemblyInfo.cs
rynowak: Hmm, I'm not sure if this is needed.

rynowak:
This should go into dev. We don't have anything to hide...

Yeah, feel free to take this over and do the needful if you like.

src/Microsoft.AspNetCore.Routing.Abstractions/MetadataCollection.cs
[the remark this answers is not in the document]
rynowak: I'm not sure that there's actually a problem here, it's just my habit to make private fields mutable on structs to avoid that, I don't think it matters much and if it does we'll find it later.

src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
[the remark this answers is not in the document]
JamesNK: Ok. Compared to locking I guess the disadvantage of this way is allocating additional TCS while waiting for initialization to complete.
Advantage is you don't need a flag to say initialization is complete and a lock object.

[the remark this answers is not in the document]
rynowak: I don't feel strongly about the details, this seemed idiomatic to me.

src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
```diff
@@ -0,0 +1,348 @@
...
+            {
+                if (parameter.DefaultValue != null)
+                {
+                    entry.Defaults.Add(parameter.Name, parameter.DefaultValue);
+                }
+            }
+            return entry;
+        }
+
+        private struct Key : IEquatable<Key>
```
JamesNK: Struct can be readonly

src/Microsoft.AspNetCore.Routing/Properties/AssemblyInfo.cs
JamesNK: MatcherEndpoint is internal

src/Microsoft.AspNetCore.Routing/Tree/TreeRouter.cs
JamesNK: I appreciate what you're doing but I'd prefer the proper name :grimacing:

src/Microsoft.AspNetCore.Routing.Abstractions/MetadataCollection.cs
kichalla: Yup 👍

src/Microsoft.AspNetCore.Routing/EndpointDataSource.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public abstract class EndpointDataSource
+    {
+        public abstract IChangeToken ChangeToken { get; }
```
kichalla: what is the scenario behind having this change token here?

rynowak: Route data and endpoints can change after startup. MVC already has this feature 😆 - if you drop in a Razor Page while the site is running, it will be added to the route table.

src/Microsoft.AspNetCore.Routing/Internal/DataSourceDependantCache.cs
kichalla: Something related to my comment earlier in this PR. Do we have scenarios for dynamic changing of endpoints apart from Razor pages or MVC? Should we do thing iteratively gather perf numbers along the way?

src/Microsoft.AspNetCore.Routing/Matchers/Matcher.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Attempts to asynchronously select an <see cref="Endpoint"/> for the current request.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
+        /// <param name="feature">
+        /// The <see cref="IEndpointFeature"/> associated with the current request. The 
+        /// <see cref="IEndpointFeature"/> will be mutated to contain the result of the operation.</param>
+        /// <returns>A <see cref="Task"/> which represents the asynchronous completion of the operation.</returns>
+        public abstract Task MatchAsync(HttpContext httpContext, IEndpointFeature feature);
```
kichalla: Why this needs to be async? scenarios?

samples/DispatcherSample.Web/Startup.cs
```diff
@@ -0,0 +1,58 @@
...
+            services.AddDispatcher(options =>
+            {
+                options.DataSources.Add(new DefaultEndpointDataSource(new[]
+                {
+                    new MatcherEndpoint((next) => (httpContext) =>
```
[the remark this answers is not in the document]
kichalla: I have created a PR already for this: https://github.com/aspnet/Routing/pull/522

src/Microsoft.AspNetCore.Routing/Internal/DataSourceDependantCache.cs
rynowak: Spoke to Kiran offline about this. I think that this is a critical requirement, and the design here isolates us well from the perf impact of this.

src/Microsoft.AspNetCore.Routing/Matchers/Matcher.cs
rynowak: Spoke offline with @kichalla about this. 

I think the part that will be async is the part where user-defined policies (like versioning) can select between actions. I might be wrong about this, but I'm a bit gunshy saying something will be *never* async.  That later of functionality isn't included in the code dupe because I don't like most of the things I tried in the other iterations of the code. I'd rather us apply some fresh thinking to that area and come up with something great!

rynowak:
@JamesNK - is there anything else to do here or good to go from your point of view?

src/Microsoft.AspNetCore.Routing/Matchers/AmbiguousMatchException.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    /// <summary>
+    /// An exception which indicates multiple matches in endpoint selection.
+    /// </summary>
+    [Serializable]
+    internal class AmbiguousMatchException : Exception
```
JamesNK: Why internal?

rynowak: I can't think of a case where a user would want to handle it. 

But it seems useful for us to have a specific exception type because the name you see in the debugger makes it pretty obvious that it's not a bug in the framework.

src/Microsoft.AspNetCore.Routing.Abstractions/IEndpointFeature.cs
davidfowl: We need to file an issue for this and have that discussion. We should be crisp about what the core exchange types are here. I would be fine moving RVD to HttpAbstractions. We need to rename the other 2 types but those look reasonable as well.

PR #526, opened 2018-06-05
Added a mock Cors scenario

kichalla:


samples/DispatcherSample.Web/CorsApis.cs
```diff
@@ -0,0 +1,666 @@
+﻿using System;
```
kichalla: You can ignore this class as its a copy of some apis from Cors. You can take a look at file `StartupUsingMyCorsMiddleware` to see the usage. Its pretty straightforward

JamesNK: Add a comment for future readers about where this is copied from.
Also add the standard ASP.NET copyright header to all the cs files.

samples/DispatcherSample.Web/MyCorsMiddleware.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace DispatcherSample.Web
+{
+    public class MyCorsMiddleware
```
JamesNK: `SampleCorsMiddleware`?

samples/DispatcherSample.Web/MyCorsMiddleware.cs
```diff
@@ -0,0 +1,106 @@
...
+            var endpointFeature = httpContext.Features.Get<IEndpointFeature>();
+            if (endpointFeature != null)
+            {
+                var metadataCollection = endpointFeature.Endpoint.Metadata;
+                if (metadataCollection.Count > 0)
+                {
+                    for (var i = metadataCollection.Count - 1; i >= 0; i--)
+                    {
+                        var metadata = metadataCollection[i];
+                        if (metadata is IDisableCorsAttribute)
```
JamesNK: Simple yet decoupled. I like it 👍

samples/DispatcherSample.Web/CorsApis.cs
```diff
@@ -0,0 +1,666 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+using Microsoft.Net.Http.Headers;
+
+namespace DispatcherSample.Web
```
rynowak: Since this is just an architectural sample, does it make sense to just do something cheezy rather than copying all of the CORS code into this repo?

To me the value of the sample is that we have an example of the architecture to point to, the fact that it's a real CORS implementation isn't the main idea.

samples/DispatcherSample.Web/MyCorsMiddleware.cs
```diff
@@ -0,0 +1,106 @@
...
+            await _next(httpContext);
+        }
+
+        private bool IsCorsEnabled(HttpContext httpContext)
+        {
+            var endpointFeature = httpContext.Features.Get<IEndpointFeature>();
+            if (endpointFeature != null)
+            {
+                var metadataCollection = endpointFeature.Endpoint.Metadata;
+                if (metadataCollection.Count > 0)
```
rynowak: This code can be simplified a bit:
suggest: `httpContext.Features.Get<IEndpointFeature>()?.Endpoint`
suggest: `remove the `if` at L68, it's redundant.

rynowak: Also https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing.Abstractions/MetadataCollection.cs#L31

samples/DispatcherSample.Web/MyCorsMiddleware.cs
```diff
@@ -0,0 +1,106 @@
...
+                // Since there is a policy which was identified,
+                // always respond to preflight requests.
+                httpContext.Response.StatusCode = StatusCodes.Status204NoContent;
+                return;
+            }
+
+            await _next(httpContext);
+        }
+
+        private bool IsCorsEnabled(HttpContext httpContext)
```
rynowak: You could make this into a `TryGetCorsPolicy` method and simplify things a bit further. The code as it is does the lookup twice.

samples/DispatcherSample.Web/MyCorsMiddleware.cs
kichalla: Yeah, I tried to use the `GetMetadata<>` api initially but the problem was finding the 'nearest' metadata for my endpoint in Cors case. Like I want the nearest `IEnableCorsAttribute` or `IDisableCorsAttribute`. Had there been a single metadata type like `IAuthorizeData`, I would not have a problem using this api.

samples/DispatcherSample.Web/CorsApis.cs
kichalla: Sure, I agree. Actually I didn't think about checking this sample in and just wanted to see how things would work with real Cors implementation. But yeah, we could have this in sample and have couple of tests to verify a integration like scenario.

rynowak: Yeah, I would suggest trying to keep it limited.

rynowak:
What's the plan for this PR - did you accomplish your objective by prototyping this and discussing?

kichalla:
> What's the plan for this PR - did you accomplish your objective by prototyping this and discussing?

Yes @rynowak. As discussed offline, I am closing this PR.

Issue #527, opened 2018-06-06
MVC integration - endpoint matching


JamesNK:
- [x] Update abstractions
- [x] Invoke MVC action via conventional route
- [x] Invoke MVC action via attribute route

JamesNK:
@rynowak How do you imagine conventional routing will work?

Off the top of my head it would be a helper method inside MVC, called at startup like `MapRoute`/`MapAreaRoute`, that a dispatcher data source then uses to resolve the matched template's route values to an action using the equivilent of `ActionSelector.SelectCandidates`, and then filter to a final result using `SelectBestCandidate`.

Warm/cold? Worth discussing in a call?

rynowak:
Yes, that sounds about right. You might be able to reuse a lot of that code. Feel free to setup a call if there's more you want to discuss or confirm, but that's what I had in mind.

JamesNK:
It's almost like what I know what I'm doing 🍰

JamesNK:
Finished with https://github.com/aspnet/Mvc/pull/7928

Will open individual issues for additional fixes/features/work

PR #528, opened 2018-06-06
Fix EndpointMetadataCollection filename

JamesNK:
https://github.com/aspnet/Routing/issues/527


src/Microsoft.AspNetCore.Routing.Abstractions/DispatcherDataSource.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public abstract class DispatcherDataSource : IEndpointCollectionProvider
```
JamesNK: I left out `Address` for now. I need to learn more about the concept of addresses.

JamesNK:
@rynowak Are there abstractions required that I have missed?

rynowak:
what do you think we gain by moving this type to abstractions?

JamesNK:
Ah, I mistakenly thought that MVC referenced only the abstractions so it had to be here.

Hmmm, right now I don't know enough about the use and purpose of the abstractions package to say where it should go.

rynowak:
Yeah I'm honestly not sure either.

rynowak:
> what do you think we gain by moving this type to abstractions?

To help explore this a little more - **if** the data source is part of abstractions - then we're effectively saying "you can provide routing info by just referencing routing.abstractions". I'm not sure if that's going to shake out to be true, or untrue.

----

My perspective (I also want to hear yours): I not sure if it's a priority to enable `you can provide routing info by just referencing routing.abstractions`. Should we let other things happen and see if it comes up? I don't want to make a bunch of 2nd order and 3rd order decisions based on something unimportant.

It's also non-breaking to move a type from routing -> routing.abstractions.

PR #529, opened 2018-06-06
Initial DispatcherMiddleware tests

JamesNK:
A couple of tests written while reviewing the existing routing tests

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
@@ -64,7 +64,11 @@ namespace Microsoft.AspNetCore.Routing
             await matcher.MatchAsync(httpContext, feature);
             if (feature.Endpoint != null)
             {
-                Log.EndpointMatched(_logger, feature);
+                Log.MatchSuccess(_logger, feature);
+            }
+            else
+            {
+                Log.MatchFailure(_logger);
             }
 
             await _next(httpContext);
@@ -101,13 +105,23 @@ namespace Microsoft.AspNetCore.Routing
         private static class Log
         {
             private static readonly Action<ILogger, string, Exception> _matchSuccess = LoggerMessage.Define<string>(
-                LogLevel.Information,
-                new EventId(0, "MatchSuccess"),
+                LogLevel.Debug,
+                new EventId(1, "MatchSuccess"),
                 "Request matched endpoint '{EndpointName}'.");
 
-            public static void EndpointMatched(ILogger logger, EndpointFeature feature)
+            private static readonly Action<ILogger, Exception> _matchFailure = LoggerMessage.Define(
+                LogLevel.Debug,
+                new EventId(2, "MatchFailure"),
+                "Request did not match any endpoints.");
+
+            public static void MatchSuccess(ILogger logger, EndpointFeature feature)
             {
                 _matchSuccess(logger, feature.Endpoint.DisplayName, null);
+            }
+
+            public static void MatchFailure(ILogger logger)
+            {
+                _matchFailure(logger, null);
             }
         }
     }
--- a/test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
@@ -5,7 +5,10 @@ using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing.Matchers;
+using Microsoft.AspNetCore.Routing.Tests.TestObjects;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Extensions.Options;
 using Xunit;
@@ -15,22 +18,15 @@ namespace Microsoft.AspNetCore.Routing
     public class DispatcherMiddlewareTest
     {
         [Fact]
-        public async void Invoke_LogsCorrectValues_WhenNotHandled()
+        public async Task Invoke_OnCall_SetsEndpointFeature()
         {
             // Arrange
-            var expectedMessage = "Request did not match any endpoints.";
-
-            var sink = new TestSink(
-                TestSink.EnableWithTypeName<DispatcherMiddleware>,
-                TestSink.EnableWithTypeName<DispatcherMiddleware>);
-            var loggerFactory = new TestLoggerFactory(sink, enabled: true);
-
             var httpContext = new DefaultHttpContext();
-            httpContext.RequestServices = new ServiceProvider();
+            httpContext.RequestServices = new TestServiceProvider();
 
             RequestDelegate next = (c) => Task.FromResult<object>(null);
 
-            var logger = new Logger<DispatcherMiddleware>(loggerFactory);
+            var logger = new Logger<DispatcherMiddleware>(NullLoggerFactory.Instance);
             var options = Options.Create(new DispatcherOptions());
             var matcherFactory = new TestMatcherFactory(false);
             var middleware = new DispatcherMiddleware(matcherFactory, options, logger, next);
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
```diff
@@ -64,7 +64,11 @@ internal sealed class DispatcherMiddleware
             await matcher.MatchAsync(httpContext, feature);
             if (feature.Endpoint != null)
             {
-                Log.EndpointMatched(_logger, feature);
+                Log.MatchSuccess(_logger, feature);
+            }
+            else
+            {
+                Log.MatchFailure(_logger);
```
JamesNK: RoutingMiddleware only logged on failure. DispatcherMiddleware (previously) only logged on success. What is correct?

src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
```diff
@@ -101,14 +105,24 @@ private Task<Matcher> InitializeAsync()
         private static class Log
         {
             private static readonly Action<ILogger, string, Exception> _matchSuccess = LoggerMessage.Define<string>(
-                LogLevel.Information,
-                new EventId(0, "MatchSuccess"),
+                LogLevel.Debug,
+                new EventId(1, "MatchSuccess"),
```
JamesNK: What number should event ID start at? On SignalR we always started at 1, and I like the idea of leaving zero alone.

test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.Extensions.Logging.Testing;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class DispatcherMiddlewareTest
+    {
+        [Fact]
+        public async void Invoke_LogsCorrectValues_WhenNotHandled()
```
rynowak: async void alert

test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
```diff
@@ -0,0 +1,127 @@
...
+                _isHandled = isHandled;
+            }
+
+            public override Matcher CreateMatcher(EndpointDataSource dataSource)
+            {
+                return new TestMaster(_isHandled);
+            }
+        }
+
+        private class TestMaster : Matcher
```
rynowak: TestMatcher

test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            // Act
+            await middleware.Invoke(httpContext);
+
+            // Assert
+            Assert.Empty(sink.Scopes);
+            var write = Assert.Single(sink.Writes);
+            Assert.Equal(expectedMessage, write.State?.ToString());
+        }
+
```
rynowak: Is there any real functionality we should be testing here? 

I'll be honest, I'm on the record as strongly against adding tests for logging. We did them in a bunch of places in MVC and they have been nothing but a headache. In my experience from MVC it was a really bad tradeoff for us in terms of cost.

test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
```diff
@@ -0,0 +1,127 @@
...
+        private class TestMatcherFactory : MatcherFactory
```
rynowak: I could see these types being used over and over again (in particular test endpoint). Worth just pre-emptively pulling out into top level classes?

src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
JamesNK: I agree on Debug. Can review logging later.

test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
JamesNK: oh dear. This is a copy pasta so I'll update the tests I copied from.

test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
JamesNK: Yeah tbh I'm not crazy about tests for logs except in certain situations. I'll refactor to test that underlying types are called.

test/Microsoft.AspNetCore.Routing.Tests/TestObjects/TestEndpoint.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing.Tests.TestObjects
+{
+    internal class TestEndpoint : Endpoint
```
JamesNK: Thoughts on concrete test objects like this vs Moq? I lean towards test objects

test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            // Arrange
+            var httpContext = new DefaultHttpContext();
+            httpContext.RequestServices = new TestServiceProvider();
+
+            RequestDelegate next = (c) => Task.FromResult<object>(null);
+
+            var logger = new Logger<DispatcherMiddleware>(NullLoggerFactory.Instance);
+            var options = Options.Create(new DispatcherOptions());
+            var matcherFactory = new TestMatcherFactory(false);
```
kichalla: named args

test/Microsoft.AspNetCore.Routing.Tests/TestObjects/TestMatcher.cs
```diff
@@ -0,0 +1,31 @@
...
+        public TestMatcher(bool isHandled)
+        {
+            _isHandled = isHandled;
+        }
+
+        public override Task MatchAsync(HttpContext httpContext, IEndpointFeature feature)
+        {
+            if (_isHandled)
+            {
+                feature.Endpoint = new TestEndpoint(EndpointMetadataCollection.Empty, "Test endpoint");
```
kichalla: Is this temporary? because endpoints are currently exposed through DispatcherOptions, so i think the test setup should reflect that.

test/Microsoft.AspNetCore.Routing.Tests/TestObjects/TestEndpoint.cs
kichalla: I think this is fine.

rynowak: I like concrete test objects for common things 👍

test/Microsoft.AspNetCore.Routing.Tests/DispatcherMiddlewareTest.cs
rynowak: Set your ❤️ free then

test/Microsoft.AspNetCore.Routing.Tests/TestObjects/TestMatcher.cs
rynowak: Do you think it adds value doing that in a testing mock though?

JamesNK: Flowing an endpoint from the DispatcherOptions would be a test for the real matcher. This is a minimum viable product matcher.

But it can evolve in future as we need it to.

JamesNK:
@rynowak Hey what's the deal with namespaces? Should everything sit under `Microsoft.AspNetCore.Routing` and not `Microsoft.AspNetCore.Routing.Tests`?

rynowak:
Yes, the former. If you don't just ever append `.Tests` then your life is much simpler.

PR #542, opened 2018-06-08
Fix Endpoint order in TreeMatcher

JamesNK:


src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
```diff
@@ -194,7 +194,7 @@ private UrlMatchingTree[] CreateTrees(IReadOnlyList<Endpoint> endpoints)
             foreach (var group in groups)
             {
                 var template = TemplateParser.Parse(group.Key.Template);
-                var entryExists = entries.Any(item => item.RouteTemplate.TemplateText == template.TemplateText);
+                var entryExists = entries.Any(item => item.RouteTemplate.TemplateText == template.TemplateText && item.Order == group.Key.Order);
```
rynowak: I'm not totally sure what problem this is solving. Aren't all of the groups already unique by `(template, order)`? Do we need this `if` at all?

JamesNK: I know one thing it is doing. The raw template in the attribute is being parsed and resolved by the `TemplateParser`.

rynowak: Ah yeah, cool and good. I'll not block you then, this is all kinda prototype at this stage

PR #543, opened 2018-06-08
Add sample matcher tests

rynowak:


rynowak:
/cc @JamesNK @kichalla 

I wanted to put this out there for early feedback on the general direction. If y'all find it acceptable/promising then I'll start dumping in test cases from our existing test sets.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/TreeRouterMatcher.cs
```diff
@@ -41,7 +41,6 @@ private TreeRouterMatcher(TreeRouter inner)
 
             if (context.Handler != null)
             {
-                httpContext.Features.Set<IEndpointFeature>(feature);
```
rynowak: This didn't need to be here

test/Microsoft.AspNetCore.Routing.Tests/Matchers/DispatcherAssert.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    internal static class DispatcherAssert
```
rynowak: I tend to build this sort of thing for every significant project I start nowadays. IMO it pays for itself many times.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/MatcherConformanceTest.cs
```diff
@@ -0,0 +1,84 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public abstract class MatcherConformanceTest
```
rynowak: The 'matcher' contract is a little more rich than the routing contracts of yore, so it seems good to express these tests in terms of 'matcher' rather than anything else.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/MatcherConformanceTest.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        internal abstract Matcher CreateMatcher(MatcherEndpoint endpoint);
+
+        [Fact]
+        public virtual async Task Match_SingleLiteralSegment_Success()
```
rynowak: The idea is that each matcher implementation would subclass this class. This class (the base class) contains all of the actual test definitions.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/MatcherConformanceTest.cs
```diff
@@ -0,0 +1,84 @@
...
+            await matcher.MatchAsync(httpContext, feature);
+
+            // Assert
+            DispatcherAssert.AssertMatch(feature, endpoint);
+        }
+
+        [Theory]
+        [InlineData("simple")]
+        [InlineData("/simple")]
+        [InlineData("~/simple")]
```
rynowak: I did this just for you @JamesNK

test/Microsoft.AspNetCore.Routing.Tests/Matchers/TreeRouterMatcherConformanceTest.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.Extensions.ObjectPool;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public class TreeRouterMatcherConformanceTest : MatcherConformanceTest
+    {
+        [Fact(Skip= "OH NOES unsupported")]
+        public override Task Match_SingleLiteralSegment_Success()
```
rynowak: This is an example of how we can handle divergences in behavior. 

You can override and skip if it's a temporary issue, or override and noop if it's a known difference in expectations.

This also allows us to add more targeted tests for specific implementations if we so desire.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/TreeRouterMatcherConformanceTest.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.AspNetCore.Routing.Template;
+using Microsoft.AspNetCore.Routing.Tree;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.ObjectPool;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public class TreeRouterMatcherConformanceTest : MatcherConformanceTest
```
rynowak: I expect we'll have 3 of these:

```
DefaultMatcherConformanceTest 
RouteMatcherConformanceTest 
TreeRouterMatcherConformanceTest 
```

rynowak:
@JamesNK @kichalla - any comments on this idea?

test/Microsoft.AspNetCore.Routing.Tests/Matchers/DispatcherAssert.cs
```diff
@@ -0,0 +1,55 @@
...
+                    $"values: {FormatRouteValues(feature.Values)}.");
+            }
+        }
+
+        public static void AssertNotMatch(IEndpointFeature feature)
+        {
+            if (feature.Endpoint != null)
+            {
+                throw new XunitException(
+                    $"Was expected not to match'{feature.Endpoint.DisplayName}' " +
```
JamesNK: missing space

test/Microsoft.AspNetCore.Routing.Tests/Matchers/DispatcherAssert.cs
```diff
@@ -0,0 +1,55 @@
...
+                throw new XunitException(
+                    $"Was expected to match '{expected.DisplayName}' but matched " +
+                    $"'{feature.Endpoint.DisplayName}' with values: {FormatRouteValues(feature.Values)}.");
+            }
+
+            if (values.Count != feature.Values.Count ||
+                !values.OrderBy(kvp => kvp.Key).SequenceEqual(feature.Values.OrderBy(kvp => kvp.Key)))
+            {
+                throw new XunitException(
+                    $"Was expected to match '{expected.DisplayName}' with values {{}} but matched " +
```
JamesNK: `{{}}` ?

test/Microsoft.AspNetCore.Routing.Tests/Matchers/TreeRouterMatcher.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing.Tree;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    // This is an adapter to use TreeRouter in the conformance tests
+    internal class TreeRouterMatcher : Matcher
+    {
+        private TreeRouter _inner;
```
JamesNK: readonly

test/Microsoft.AspNetCore.Routing.Tests/Matchers/TreeRouterMatcherConformanceTest.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        [Fact(Skip= "OH NOES unsupported")]
```
JamesNK: nit: `Skip =`

test/Microsoft.AspNetCore.Routing.Tests/Matchers/MatcherConformanceTest.cs
```diff
@@ -0,0 +1,84 @@
...
+            var (httpContext, feature) = CreateContext("/simple");
+
+            // Act
+            await matcher.MatchAsync(httpContext, feature);
+
+            // Assert
+            DispatcherAssert.AssertMatch(feature, endpoint);
+        }
+
+        private static (HttpContext httpContext, IEndpointFeature feature) CreateContext(string path)
```
JamesNK: I think we're going to use these helper methods all throughout the unit tests. It would be worth having them public on a test helper class of some kind.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/DispatcherAssert.cs
JamesNK: Ok! Interested to see how it gets used.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/DispatcherAssert.cs
```diff
@@ -0,0 +1,55 @@
...
+                throw new XunitException(
+                    $"Was expected not to match'{feature.Endpoint.DisplayName}' " +
+                    $"but matched with values: {FormatRouteValues(feature.Values)}.");
+            }
+        }
+
+        private static string FormatRouteValues(RouteValueDictionary values)
+        {
+            return "{" + string.Join(", ", values.Select(kvp => $"{kvp.Key} = '{kvp.Value}'")) + "}";
+
```
JamesNK: nit: extra line

test/Microsoft.AspNetCore.Routing.Tests/Matchers/MatcherConformanceTest.cs
rynowak: Well the idea is that all of the tests for matchers would subclass this. So I suppose it wouldn't be private.

rynowak:
OK Cool beans, I'm going to clean this up based on @JamesNK comments and then merge. Then I'll be adding a few more matcher implementations and more tests

test/Microsoft.AspNetCore.Routing.Tests/Matchers/TreeRouterMatcherConformanceTest.cs
rynowak: This is just an example of what we would do if we need to skip a test for a specific matcher - not checking it in 😆

PR #546, opened 2018-06-09
WIP: Dispatcher: Link generation

kichalla:


test/Microsoft.AspNetCore.Routing.Tests/DefaultEndpointFinderTest.cs
```diff
@@ -0,0 +1,224 @@
...
+            // Assert
+            var endpoint = Assert.Single(result);
+            Assert.Same(endpoint2, endpoint);
+        }
+
+        [Fact]
+        public void FindEndpoints_FindsMultipleEndpoints_WithMatchingName()
+        {
+            // Arrange
+            var name = "common-tag-for-all-my-section's-routes";
```
kichalla: One could write a custom finder which might do a wildcard search on the address names "admin*"

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
```diff
@@ -8,15 +8,21 @@ namespace Microsoft.AspNetCore.Routing
...
             Metadata = metadata ?? EndpointMetadataCollection.Empty;
             DisplayName = displayName;
+            Address = address;
         }
 
         public string DisplayName { get; }
 
         public EndpointMetadataCollection Metadata { get; }
+
+        public Address Address { get; }
```
kichalla: I view Address here as kind of a metadata. It can be extended by users who could write a custom finder to look up apart from name/methodinfo

src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -0,0 +1,95 @@
...
+            return false;
+        }
+
+        private string GetLink(
+            RouteTemplate template,
+            IReadOnlyDictionary<string, object> defaultValues,
+            LinkGeneratorContext context)
+        {
+            var defaults = new RouteValueDictionary(defaultValues);
+            var templateBinder = new TemplateBinder(
```
kichalla: Ideally this can be constructed once per matcherendpoint, but wanted to send out this PR to see if the general direction is correct.

samples/DispatcherSample.Web/StartupUsingLinkGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Matchers;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
+
+namespace DispatcherSample.Web
+{
+    public class StartupUsingLinkGenerator
```
rynowak: I think this sample needs a little more explanation before I'm ready to dig into it - I expected to see something a little more bare-bones

src/Microsoft.AspNetCore.Routing.Abstractions/Address.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Reflection;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class Address
```
rynowak: How do you imagine that MVC's URL generation will work (take `Url.Action` as an example)

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
rynowak: Do you think it's fundamental that `Endpoint -> Address` is `1:1?` 

If I wrote a custom finder, how does that address get associated with the endpoint?

src/Microsoft.AspNetCore.Routing.Abstractions/IEndpointFinder.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public interface IEndpointFinder
+    {
+        IEnumerable<Endpoint> FindEndpoints(Address address);
```
rynowak: This seems to imply and than `Address` has multiple endpoints associated - which probably makes sense.

How do you imagine that addresses are compared for equality?

src/Microsoft.AspNetCore.Routing.Abstractions/Address.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public Address(string name)
```
rynowak: Do you think that addresses are polymorphic? (should this class be abstract or sealed)

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGeneratorContext.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class LinkGeneratorContext
+    {
+        public Address Address { get; set; }
+
+        public IDictionary<string,object> AmbientValues { get; set; }
```
rynowak: Formatting YO #DesignPR

src/Microsoft.AspNetCore.Routing/DefaultEndpointFinder.cs
```diff
@@ -0,0 +1,66 @@
...
+                    result.Add(endpoint);
+                }
+
+                if (lookupAddress.MethodInfo != null &&
+                    lookupAddress.MethodInfo.Equals(endpoint.Address.MethodInfo))
+                {
+                    result.Add(endpoint);
+                    break;
+                }
+            }
```
rynowak: I'm a little surprised to see a the `break` there - what are the rules for this algorithm?

src/Microsoft.AspNetCore.Routing/DependencyInjection/DispatcherServiceCollectionExtensions.cs
```diff
@@ -11,6 +11,19 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public static class DispatcherServiceCollectionExtensions
     {
+        public static IServiceCollection AddLinkGeneration(this IServiceCollection services)
```
rynowak: I think this should jsut be included by default with the dsipatcher services.

src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -0,0 +1,95 @@
...
+            foreach (var endpoint in matcherEndpoints)
+            {
+                link = GetLink(endpoint.RouteTemplate, endpoint.Values, context);
+                if (link != null)
+                {
+                    return true;
+                }
+            }
+
+            return false;
```
rynowak: This is exactly the kind of thing I was hoping to see 😻

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
```diff
@@ -30,14 +32,15 @@ public sealed class MatcherEndpoint : Endpoint
...
+            RouteTemplate = TemplateParser.Parse(template);
             Values = new RouteValueDictionary(values);
             Order = order;
         }
 
         public int Order { get; }
         public Func<RequestDelegate, RequestDelegate> Invoker { get; }
         public string Template { get; }
-
+        public RouteTemplate RouteTemplate { get; }
```
rynowak: I have a few reservations about exposing this type prominently because it's kinda gross... I don't think it's blocking for now.

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
JamesNK: If it is metadata should it be something that gets added to the metadata collection?

src/Microsoft.AspNetCore.Routing/DefaultEndpointFinder.cs
```diff
@@ -0,0 +1,66 @@
...
+            var endpointsWithAddress = allEndpoints.Where(ep => ep.Address != null);
+            if (!endpointsWithAddress.Any())
+            {
+                return allEndpoints;
+            }
+
+            var result = new List<Endpoint>();
+            foreach (var endpoint in endpointsWithAddress)
+            {
+                if (!string.IsNullOrEmpty(lookupAddress.Name) &&
```
JamesNK: Should this equality test exist on `Address` via `IEquatable<Address>`?

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
JamesNK: At least add a comment to say this needs to be review

test/Microsoft.AspNetCore.Routing.Tests/DefaultEndpointFinderTest.cs
```diff
@@ -0,0 +1,224 @@
...
+            public void Contact() { }
+        }
+
+        private class AdminController
+        {
+            public void Index() { }
+            public void Contact() { }
+        }
+
+        private class TestEndpoint : Endpoint
```
JamesNK: Reuse `Microsoft.AspNetCore.Routing.TestObjects.TestEndpoint`

src/Microsoft.AspNetCore.Routing.Abstractions/IEndpointFinder.cs
kichalla: > How do you imagine that addresses are compared for equality?

 Please check this if that's what you mean: https://github.com/aspnet/Routing/pull/546/files#r194552515

benchmarkapps/Benchmarks/StartupUsingDispatcher.cs
```diff
@@ -35,7 +35,8 @@ public void ConfigureServices(IServiceCollection services)
                         values: new { },
                         order: 0,
                         metadata: EndpointMetadataCollection.Empty,
-                        displayName: "Plaintext"),
+                        displayName: "Plaintext",
+                        address: null),
```
rynowak: Future design note:

I think this might be too coupled. We might need the ability to pluggably 'scavenge' addresses from endpoints, rather than requiring the addresses to be known when we create them. 

We also might need multiple addresses per endpoint.

I don't think these are blockers right now.

samples/DispatcherSample.Web/StartupUsingLinkGenerator.cs
```diff
@@ -0,0 +1,198 @@
...
+namespace DispatcherSample.Web
+{
+    /// <summary>
+    /// This sets up a sample where a custom EndpointDataSource creates endpoints with MethodInfo based address
+    /// and links are generated based on the lookup of this MethodInfo.
+    /// Flow of a request is: Dispatcher selects an endpoint => Endpoint middleware is invoked => Selected endpoint
+    /// invokes a func which creates an invoker called IHubInvoker which uses the current endpoint's MethodInfo to
+    /// instantiate and execute the target method.
+    /// </summary>
+    public class StartupUsingLinkGenerator
```
rynowak: Discussed in person, and this was just done as the thought experiment, and not to be checked in.

src/Microsoft.AspNetCore.Routing.Abstractions/Address.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class Address
```
rynowak: Future design note:

I think it's fine for the first cut if address is simple and tightly coupled to the idea of a string identifier. We discussed the idea of address being polymorphic since MVC's addresses are a list of key-value-pairs

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
rynowak: This to me seems like something that will grow up as we put more into this.

src/Microsoft.AspNetCore.Routing.Abstractions/ILinkGenerator.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public interface ILinkGenerator
+    {
+        bool TryGetLink(LinkGeneratorContext context, out string link);
+
+        string GetLink(LinkGeneratorContext context);
+    }
```
rynowak: Future design note:

This is pretty bare-bones for now which is fine. We might want this to be an abstract class. It's very likely this surface area will grow.

rynowak: We also don't expect user code to implement this outside of maybe mocks for testing, so there's no real need to make it easy to do so.

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGeneratorContext.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class LinkGeneratorContext
+    {
+        public Address Address { get; set; }
+
+        public IDictionary<string, object> AmbientValues { get; set; }
+
+        public IDictionary<string, object> SuppliedValues { get; set; }
```
rynowak: Future design note:

Having address on here is pretty strange. You'd be reaching into the FINDER to get an address and then calling the link generator. Basically, using a low lever concept to call into a high level concept.

rynowak: These should be RVD I think

src/Microsoft.AspNetCore.Routing/DefaultEndpointFinder.cs
```diff
@@ -0,0 +1,53 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class DefaultEndpointFinder : IEndpointFinder
```
rynowak: Can this be internal?

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
```diff
@@ -36,14 +38,42 @@ public sealed class MatcherEndpoint : Endpoint
...
         }
 
         public int Order { get; }
         public Func<RequestDelegate, RequestDelegate> Invoker { get; }
         public string Template { get; }
-
         public IReadOnlyDictionary<string, object> Values { get; }
+
+        // Todo: needs review
+        public RouteTemplate ParsedTemlate { get; }
```
rynowak: I don't like the idea of putting `RouteTemplate` on any new public APIs because the design of it is kinda wack.

I wrote a **clean** version of `RouteTemplate` and the parsing code that we can use if necessary, but that kinda duplicates the concept so I'm not in love with that idea either.

Could we find a way to make this usage less prominent? (not a public property on a public class).

I'm also OK with doing this temporarily if there's an issue to clean it up.

benchmarkapps/Benchmarks/StartupUsingDispatcher.cs
kichalla: Filed: https://github.com/aspnet/Routing/issues/563

src/Microsoft.AspNetCore.Routing.Abstractions/Address.cs
kichalla: Filed: https://github.com/aspnet/Routing/issues/564

src/Microsoft.AspNetCore.Routing.Abstractions/ILinkGenerator.cs
kichalla: Filed: https://github.com/aspnet/Routing/issues/565

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
kichalla: Filed https://github.com/aspnet/Routing/issues/566

kichalla:
d3ddc1709a8c6d0d9df58fd695cdaa3db890d125

Issue #534, opened 2018-06-06
MVC IActionConstraint like functionality


JamesNK:
Current routing supports selecting the correct action via `IActionSelector` and `IActionConstraint`. Implement similar (or same) ability in dispatching. Design required.

JamesNK:
@rynowak What on your thoughts on where `IActionConstraint` (`IEndpointConstraint`?) like functionality would live?

Will it be embedded inside the `Matcher` and the matcher is responsible for returning a final endpoint?

Alternatively would the matcher return a list of possible endpoints to something like `IEndpointSelector` and it would then select the best match using configured constraints?

rynowak:
I think these are implementation details, so we have the flexibility to do whatever works best for us. If that's inside the matcher, then sure. Remember, it's a explicitly a non-goal to make part of the **dispatcher system** replaceable - so if we do factor it into two concerns then it should be internal 😆 

What I mean by this is that we want users to do things like implement a versioning policy, or implement something like https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/OverloadActionConstraint.cs


Beyond this known and well-defined scenario we're optimizing for perf, scalability, future maintainability, quality of diagnostics, etc over extensibility in this area. Providing flexibility that doesn't align with our core scenarios is a very low priority.

----

One detail that we didn't get right in `IActionConstraint` is that it's a bunch of separate policies that *vote* on whether their action is a match. It's complicated to reason about. I think it would be much much simpler to have a list of ordered policies that act on all actions at a time. Think of it as middleware for action selection.

----

The one non-straightforward thing that I feel obliged to point out is that the result of processing the URL will be (logically) a list of lists of candidates. The route tree has 'groups' of candidates where each 'group' has the same order and precedence. However we want to give any policy (like versioning) the ability to select from **all** groups, or at least to see candidates from all groups. 

I imagine that we could start with a list of lists, and then refine this into something more performant and nicer to code against once we've seen the complete picture. The set of candiates is largely static, so we can do lots of work to optimize the allocs.

---

Things like how the API design impacts allocations are definitely good to consider at this point. It's likely going to lead to us passing in some kind of mutable context.

Imagine for scoping that you'd have 1-8 endpoint candidates and 1-3 selection policies.

---

Another major decision point is whether the HTTP method feature will follow the same design as user-extensibility, will it implement some extensibility API or be totally built in? It's up to us 👍

[...]

PR #548, opened 2018-06-11
Initial endpoint constraints functionality

JamesNK:
https://github.com/aspnet/Routing/issues/534

Rework of MVC's action constraints to dispatching endpoints.


src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointSelector.cs
```diff
@@ -0,0 +1,442 @@
...
+                    return cache;
+                }
+
+                cache = new Cache(_dataSource.Endpoints);
+                Volatile.Write(ref _cache, cache);
+                return cache;
+            }
+        }
+
+        public IReadOnlyList<Endpoint> SelectCandidates(RouteContext context)
```
JamesNK: `SelectCandidates` is not used at the moment. Does it have a place in dispatching?

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointSelector.cs
```diff
@@ -0,0 +1,442 @@
...
+            {
+                Debug.Assert(matchingRouteValues != null);
+                return matchingRouteValues;
+            }
+
+            //_logger.NoEndpointsMatched(context.RouteData.Values);
+            return EmptyEndpoints;
+        }
+
+        public Endpoint SelectBestCandidate(HttpContext context, IReadOnlyList<Endpoint> candidates)
```
JamesNK: Actually used.

src/Microsoft.AspNetCore.Routing/EndpointConstraints/IEndpointConstraint.cs
```diff
@@ -0,0 +1,350 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Routing.EndpointConstraints
+{
+    public class EndpointConstraintContext
```
JamesNK: Dump of constraint related classes.

src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
```diff
@@ -137,6 +141,8 @@ internal class TreeMatcher : Matcher
 
         private Task SelectEndpointAsync(HttpContext httpContext, IEndpointFeature feature, IReadOnlyList<MatcherEndpoint> endpoints)
         {
+            var bestEndpoint = _endpointSelector.SelectBestCandidate(httpContext, endpoints);
```
JamesNK: I haven't added this to any of the other matchers. Do we want to add it?

test/Microsoft.AspNetCore.Routing.Tests/EndpointConstraints/EndpointSelectorTests.cs
```diff
@@ -0,0 +1,458 @@
...
+using Microsoft.Extensions.Logging.Testing;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing.EndpointConstraints
+{
+    public class EndpointSelectorTests
```
JamesNK: Mostly copied from MVC.

test/Microsoft.AspNetCore.Routing.Tests/Internal/HttpMethodEndpointConstraintTest.cs
```diff
@@ -0,0 +1,87 @@
...
+using Microsoft.AspNetCore.Routing.TestObjects;
+using Microsoft.Extensions.Primitives;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing.Internal
+{
+    public class HttpMethodEndpointConstraintTest
```
JamesNK: From MVC

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointSelector.cs
```diff
@@ -0,0 +1,442 @@
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (candidates == null)
+            {
+                throw new ArgumentNullException(nameof(candidates));
+            }
+
+            var finalMatches = EvaluateEndpointConstraints(context, candidates);
```
kichalla: From what I understood in our email discussion, this would be the point where we run a bunch of 'policies' in an order, which essentially act like filters which hopefully would give a single endpoint after their run. The current code could probably be broken into individual policies. Looking at something like `IEndpointConstraintMetadata` on an endpoint would be an implementation detail of our default policy(s).
As a user I could associate any kind of metadata to an endpoint and when my policy gets invoked I look for my own metadata and make decisions.

src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
```diff
@@ -15,14 +15,14 @@ internal sealed class DispatcherMiddleware
     {
         private readonly MatcherFactory _matcherFactory;
         private readonly ILogger _logger;
-        private readonly IOptions<DispatcherOptions> _options;
+        private readonly CompositeEndpointDataSource _endpointDataSource;
```
kichalla: Also we need to have an issue for making `CompositeEndpointDataSource` consider data sources added into DI too
https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing/DependencyInjection/DispatcherServiceCollectionExtensions.cs#L25

src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
```diff
@@ -94,8 +94,7 @@ private Task<Matcher> InitializeAsync()
                 null) == null)
             {
                 // This thread won the race, do the initialization.
-                var dataSource = new CompositeEndpointDataSource(_options.Value.DataSources);
-                var matcher = _matcherFactory.CreateMatcher(dataSource);
```
rynowak: I'm not really sure why I didn't do this from the beginning. I had a some concerns about initialization ordering, but it seems like that's not a problem as long as the composite data source stays lazy.

src/Microsoft.AspNetCore.Routing/DependencyInjection/DispatcherServiceCollectionExtensions.cs
```diff
@@ -33,6 +34,16 @@ public static IServiceCollection AddDispatcher(this IServiceCollection services)
             //
             services.TryAddSingleton<MatcherFactory, TreeMatcherFactory>();
 
+            //
+            // Endpoint Selection
+            //
+            services.TryAddSingleton<EndpointSelector>();
+            services.TryAddSingleton<EndpointConstraintCache>();
```
rynowak: I'm going to restrain myself from commenting too heavily since the plan for now given the plan that you explained here.

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointSelector.cs
rynowak: No, we won't need this. This is part of action selection for convention routing. Dispatcher works like attribute routing

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointSelector.cs
```diff
@@ -0,0 +1,442 @@
...
+            }
+        }
+
+        private IReadOnlyList<Endpoint> EvaluateEndpointConstraints(
+            HttpContext context,
+            IReadOnlyList<Endpoint> endpoints)
+        {
+            var candidates = new List<EndpointSelectorCandidate>();
+
+            // Perf: Avoid allocations
```
rynowak: I'm really satisified that we're finally doing perf work on this code. It's adorable that once it was *worth a comment* to explain that I was using a for-loop on purpose. These were very early days for the product.

src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
rynowak: Let's keep it here for now - I think it would create a lot of noise for the perf work that I'm doing at this stage - which is in a very early stage.

src/Microsoft.AspNetCore.Routing/DispatcherMiddleware.cs
JamesNK: I think that is already done here https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing/DependencyInjection/ConfigureDispatcherOptions.cs

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointSelector.cs
JamesNK: I suspect around ASP.NET 5 ***beta 8*** you were probably all losing your minds 😋

PR #551, opened 2018-06-12
Fix TreeMatcher's use of EndpointSelector

JamesNK:
https://github.com/aspnet/Routing/issues/534

I knew there was something I was forgetting in previous PR.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
@@ -141,37 +141,31 @@ namespace Microsoft.AspNetCore.Routing.Matchers
 
         private Task SelectEndpointAsync(HttpContext httpContext, IEndpointFeature feature, IReadOnlyList<MatcherEndpoint> endpoints)
         {
-            var bestEndpoint = _endpointSelector.SelectBestCandidate(httpContext, endpoints);
+            MatcherEndpoint endpoint;
 
-            // REVIEW: Note that this code doesn't do anything significant now. This will eventually incorporate something like IActionConstraint
-            switch (endpoints.Count)
+            try
             {
-                case 0:
-                    {
-                        Log.MatchFailed(_logger, httpContext);
-                        return Task.CompletedTask;
-                    }
-
-                case 1:
-                    {
-                        var endpoint = endpoints[0];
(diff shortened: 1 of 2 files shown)


JamesNK:
This is blocking me doing MVC stuff so imma merge it. Leave any comments in this PR and I'll clean them up later.

src/Microsoft.AspNetCore.Routing/Matchers/TreeMatcher.cs
```diff
@@ -141,37 +141,31 @@ internal class TreeMatcher : Matcher
...
-                        Log.MatchSuccess(_logger, httpContext, endpoint);
-
-                        feature.Endpoint = endpoint;
-                        feature.Invoker = endpoint.Invoker;
+                endpoint = (MatcherEndpoint)_endpointSelector.SelectBestCandidate(httpContext, endpoints);
+            }
+            catch (Exception ex)
+            {
+                // Handle AmbiguousMatchException from EndpointSelector
+                return Task.FromException(ex);
```
rynowak: uhhhhhh - this shouldn't be super needed.

rynowak: Note that this also defeats teh logging

JamesNK: As far as I know it is best practice to return a task with an exception instead of throwing from methods that return task.

Logging happens in the EndpointSelector before the AmbiguousMatchException is thrown.
The await on MatchAsync will throw the exception from the task.

rynowak: Why though? There are lots of best practices... who's to say which best practices are best?

Much folk wisdom related to `Task` dates back to .NET 4.0 (pre-await) and isn't relevant anymore.

JamesNK: It's to do with getting a `Task` from a method without awaiting it. Since the method isn't public and we're the only callers I guess it doesn't matter.

JamesNK: That whole method doesn't need to be async anyway. I'll refactor it.

PR #555, opened 2018-06-13
Fix endpoint constraint cache

JamesNK:
Fix issue when fetching endpoint constraints from cache

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintCache.cs
+++ b/src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintCache.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Http;
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintCache.cs
```diff
@@ -90,13 +101,17 @@ public IReadOnlyList<IEndpointConstraint> GetEndpointConstraints(HttpContext htt
...
             if (entry.EndpointConstraints != null)
             {
                 return entry.EndpointConstraints;
             }
 
+            if (entry.Items == null)
+            {
+                // Endpoint has no constraints
+                return null;
+            }
```
rynowak: BTW all of this stuff to do with caching is something I hope we can kick to the curb. We did a lot of design work during the early days of ASP.NET while the DI system and our expectations of how users will engage with it were in flux. 

In particular, there was a belief that it was 'important' for services and extensibility points to support *arbitrary* service lifetimes. It leads to the kind of complexity you see here and it has a maintenance and perf cost. In the case of filter and constraint lifetimes it has a pretty significant cost. @kichalla collected data on this 👍

JamesNK: I think there still might be some value here. There is a bad of metadata against each endpoint. A cache that is already resolved will be faster than checking all metadata, every request.

rynowak: Yeah, the part that's complicated here is everything surrounding `ExecuteProviders`. If you assume that the cached data doesn't change and it's **just** data then I agree - that's why I already created https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing/Internal/DataSourceDependantCache.cs

The difference in this (legacy `IActionConstraint`) case is that the items in the cache are themselves the workers, and might be created dynamically.

PR #569, opened 2018-06-16
Add more matcher tests

rynowak:
This is a code dump of existing tests for TemplateMatcher and TreeRouter
converted to the format of matcher conformance tests.

Note that most of the new tests aren't yet supported by our experimental
matchers, which don't support many of these advanced features.

### DIFF
--- a/test/Microsoft.AspNetCore.Routing.Tests/Matchers/FullFeaturedMatcherConformanceTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/Matchers/FullFeaturedMatcherConformanceTest.cs
@@ -1,0 +1,437 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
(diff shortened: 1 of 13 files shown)


test/Microsoft.AspNetCore.Routing.Tests/Matchers/FullFeaturedMatcherConformanceTest.cs
```diff
@@ -0,0 +1,437 @@
...
+        [InlineData("template/5", "template/{*parameter:int}")]
+        [InlineData("template/5", "template/{*parameter}")]
+        [InlineData("template/{parameter}", "template/{parameter:alpha}")] // constraint doesn't match
+        [InlineData("template/{parameter:int}", "template/{parameter}")]
+        [InlineData("template/{parameter:int}", "template/{*parameter:int}")]
+        [InlineData("template/{parameter:int}", "template/{*parameter}")]
+        [InlineData("template/{parameter}", "template/{*parameter:int}")]
+        [InlineData("template/{parameter}", "template/{*parameter}")]
+        [InlineData("template/{*parameter:int}", "template/{*parameter}")]
+        public async Task Match_SelectEndpoint_BasedOnPrecedence(string template1, string template2)
```
JamesNK: virtual? Also applies to the following tests in file.

Or do you only make something virtual when it needs to be overriden?

PR #574, opened 2018-06-21
Add Mvc endpoint finder

kichalla:
Related to PR: https://github.com/aspnet/Mvc/pull/7951


src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -41,13 +41,22 @@ public string GetLink(LinkGeneratorContext context)
...
+                var address = context.Address;
+                endpoints = _endpointFinder.FindEndpoints(address);
+                if (endpoints == null)
+                {
+                    return false;
+                }
+            }
+            else
+            {
+                endpoints = context.Endpoints;
```
rynowak: This really feels like it's two different APIs or abstractions rather than a single method

PR #576, opened 2018-06-21
Fix TreeMatcher with route constraints

JamesNK:


kichalla:
There is a test failure

PR #578, opened 2018-06-26
[WIP] Added a RouteValuesBasedEndpointFinder

kichalla:


src/Microsoft.AspNetCore.Routing.Abstractions/IEndpointFinderOfT.cs
```diff
@@ -5,8 +5,10 @@
 
 namespace Microsoft.AspNetCore.Routing
 {
-    public interface IEndpointFinder
+    public interface IEndpointFinder<T>
     {
-        IEnumerable<Endpoint> FindEndpoints(Address address);
+        IEnumerable<Endpoint> FindEndpoints(T finderContext);
```
rynowak: Suggestion:
```
public interface IEndpointFinder<TAddress>
{
   IEnumerable<Endpoint> FindEndpoints(TAddress address);
}
```

src/Microsoft.AspNetCore.Routing.Abstractions/IRequiredLinkValuesMetadata.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public interface IRequiredLinkValuesMetadata
```
rynowak: As we discussed, this won't be needed.

src/Microsoft.AspNetCore.Routing/NameBasedEndpointFinder.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Routing.Matchers;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class NameBasedEndpointFinder : IEndpointFinder<string>
```
rynowak: internal 😆

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -0,0 +1,170 @@
...
+            CompositeEndpointDataSource endpointDataSource,
+            ObjectPool<UriBuildingContext> objectPool)
+        {
+            // TODO: register for change events?
+            _endpointDataSource = endpointDataSource;
+            _objectPool = objectPool;
+
+            // TODO: Review: IInlineConstraintResolver is registered as Transient, where as we want this finder to be
+            // registered as singleton.
+            _inlineConstraintResolver = new DefaultInlineConstraintResolver(Options.Create(new RouteOptions()));
```
rynowak: As discussed, this can just be returned from DI.

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -0,0 +1,170 @@
...
+        private readonly IInlineConstraintResolver _inlineConstraintResolver;
+        private readonly ObjectPool<UriBuildingContext> _objectPool;
+        private readonly LinkGenerationDecisionTree _linkGenerationTree;
+
+        public RouteValuesBasedEndpointFinder(
+            CompositeEndpointDataSource endpointDataSource,
+            ObjectPool<UriBuildingContext> objectPool)
+        {
+            // TODO: register for change events?
+            _endpointDataSource = endpointDataSource;
```
rynowak: Please log an issue 👍

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -0,0 +1,170 @@
...
+
+                var entry = new OutboundRouteEntryWithEndpoint()
+                {
+                    Handler = NullRouter.Instance,
+                    Order = endpoint.Order,
+                    Precedence = RoutePrecedence.ComputeOutbound(endpoint.ParsedTemlate),
+                    RequiredLinkValues = requiredLinkValues,
+                    // TODO: handle routes with names. What does it mean to have a route name on 
+                    // a parameterized template like {controller}/{action}/{id?}, name="default" ?.
+                    RouteName = null,
```
rynowak: As discussed, we can remove the validation that prevents this from being possible.

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            var endpoints = _endpointDataSource.Endpoints.OfType<MatcherEndpoint>();
+            foreach (var endpoint in endpoints)
+            {
+                var requiredValuesMetadata = endpoint.Metadata.GetMetadata<IRequiredLinkValuesMetadata>();
+                var requiredLinkValues = requiredValuesMetadata?.Values;
+                if (requiredLinkValues == null)
+                {
+                    requiredLinkValues = new RouteValueDictionary();
+                }
```
rynowak: As discussed, let's make sure `MatcherEndpoint` exposes both `defaults` and `values` - where value is always a subset of defaults - defaults includes default values specified in the template and in `MapRoute(...)`

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
```diff
@@ -19,12 +19,12 @@ public sealed class MatcherEndpoint : Endpoint
         public MatcherEndpoint(
             Func<RequestDelegate, RequestDelegate> invoker,
             string template,
-            object values,
+            object defaults,
+            RouteValueDictionary requiredValues,
```
rynowak: These should either be both object or both `RouteValueDictionary`

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
```diff
@@ -75,5 +84,27 @@ private RouteValueDictionary GetDefaults(RouteTemplate parsedTemplate, RouteValu
...
+        // Required values: controller=foo, action=bar
+        // Final constructed template: foo/bar/{category}/{id?}
+        // Final defaults: controller=foo, action=bar, category=products
+        //
+        // Template: {controller=Home}/{action=Index}/{category=products}/{id?}
+        // Defaults: controller=Home, action=Index, category=products
+        // Required values: controller=foo, action=bar
+        // Final constructed template: foo/bar/{category}/{id?}
+        // Final defaults: controller=foo, action=bar, category=products
+        private void EnsureRequiredValuesInDefaults(RouteValueDictionary requiredValues, RouteValueDictionary defaults)
```
rynowak: Would it be better for this to throw?

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -0,0 +1,148 @@
...
+                //        StringComparison.OrdinalIgnoreCase))
+                //{
+                //    throw new ArgumentException(
+                //        Resources.FormatAttributeRoute_DifferentLinkGenerationEntries_SameName(entry.RouteName),
+                //        nameof(linkGenerationEntries));
+                //}
+                //else if (namedMatch == null)
+                //{
+                //    _namedEntries.Add(entry.RouteName, outboundMatch);
+                //}
```
rynowak: This isn't right, it needs to look for endpoints with a matching route name.

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
```diff
@@ -37,24 +37,31 @@ public sealed class MatcherEndpoint : Endpoint
...
         }
 
         public int Order { get; }
         public Func<RequestDelegate, RequestDelegate> Invoker { get; }
         public string Template { get; }
-        public IReadOnlyDictionary<string, object> Values { get; }
+        public IReadOnlyDictionary<string, object> Defaults { get; }
+
+        // Values required by an endpoint for it to be successfully matched on link generation
+        public RouteValueDictionary RequiredValues { get; }
```
rynowak: These should both be the same 😆 Defaults is one type and RequiredValues is another

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -0,0 +1,191 @@
...
+                {
+                    // When the route name is associated with a single endpoint
+                    return new[] { (MatcherEndpoint)namedMatch.Match.Entry.Data };
+                }
+            }
+
+            if (matchResults == null || !matchResults.Any())
+            {
+                return null;
+            }
```
rynowak: return `Array.Empty<>` here. Returning a null IEnumerable is a party foul, no one expect you to do that

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                var namedEntry = new NamedMatch();
+                if (namedOutboundMatch.Value.Count == 1)
+                {
+                    namedEntry.Match = namedOutboundMatch.Value[0];
+                }
+                else
+                {
+                    namedEntry.LinkGenerationTree = new LinkGenerationDecisionTree(namedOutboundMatch.Value.ToArray());
+                }
```
rynowak: I don't think we should treat a single endpoint as different - I can't think of a reason why we want these cases to behave differently

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
kichalla: Oops. Good catch.

PR #584, opened 2018-07-05
Added support for route constraints in Dispatcher world

kichalla:


src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
```diff
@@ -44,6 +46,8 @@ public sealed class MatcherEndpoint : Endpoint
             RequiredValues = requiredValues;
             var mergedDefaults = GetDefaults(ParsedTemplate, defaults);
             Defaults = mergedDefaults;
+
```
JamesNK: nit: remove

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -16,14 +17,17 @@ public class CompositeEndpointDataSource : EndpointDataSource
...
             if (dataSources == null)
             {
                 throw new ArgumentNullException(nameof(dataSources));
             }
 
-            _dataSources = dataSources.ToArray();
+            var allDataSources = new List<EndpointDataSource>();
+            allDataSources.AddRange(dataSources);
+            allDataSources.AddRange(options.Value.DataSources);
+            _dataSources = allDataSources.ToArray();
```
JamesNK: What is the reason behind this change?

src/Microsoft.AspNetCore.Routing/DependencyInjection/DispatcherServiceCollectionExtensions.cs
```diff
@@ -22,14 +22,10 @@ public static IServiceCollection AddDispatcher(this IServiceCollection services)
             }
 
             // Collect all data sources from DI.
-            services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<DispatcherOptions>, ConfigureDispatcherOptions>());
+            //services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<DispatcherOptions>, ConfigureDispatcherOptions>());
 
             // Allow global access to the list of endpoints.
-            services.TryAddSingleton<CompositeEndpointDataSource>(s =>
```
JamesNK: What is the reason behind this change?

src/Microsoft.AspNetCore.Routing/EndpointMatchConstraintBuilder.cs
```diff
@@ -0,0 +1,192 @@
...
+    /// <see cref="EndpointMatchConstraintBuilder"/> allows iterative building a set of route constraints, and will
+    /// merge multiple entries for the same key.
+    /// </remarks>
+    public class EndpointMatchConstraintBuilder
+    {
+        private readonly IInlineEndpointMatchConstraintResolver _inlineConstraintResolver;
+        private readonly string _displayName;
+
+        private readonly Dictionary<string, List<IEndpointMatchConstraint>> _constraints;
+        private readonly HashSet<string> _optionalParameters;
```
JamesNK: nit: trailing newline

src/Microsoft.AspNetCore.Routing.Abstractions/IEndpointMatchConstraint.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.AspNetCore.Routing
+{
+    public interface IEndpointMatchConstraint
+    {
+        void Initialize(string parameter);
+
+        bool Match(string routeKey, RouteValueDictionary values, RouteDirection routeDirection);
```
JamesNK: Thoughts on a context object instead of individual parameters? That way something can be added to the context object without breaking interface signature.

There would be 1 context object per route rather than 1 per evaluated constraint to perf shouldn't be impacted too heavily.

src/Microsoft.AspNetCore.Routing.Abstractions/IEndpointMatchConstraint.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.AspNetCore.Routing
+{
+    public interface IEndpointMatchConstraint
```
JamesNK: `IEndpointMatchConstraint` is perilously close to `IEndpointConstraint`. I think one of the them is going to need a different name, otherwise devs will be very confused.

kichalla:
Hi @JamesNK, Just FYI...I am actually working on a major change to the initial design, so you might see many parts of initial commit not valid anymore, but I will keep a note of your comments and address them.

kichalla:
🆙 📅

samples/DispatcherSample.Web/EndsWithStringMatchProcessor.cs
```diff
@@ -19,9 +19,18 @@ public EndsWithStringMatchProcessor(ILogger<EndsWithStringMatchProcessor> logger
...
+                return false;
+            }
+
             var valueString = Convert.ToString(value, CultureInfo.InvariantCulture);
 
-            var endsWith = valueString.EndsWith(ConstraintArgument);
+            var endsWith = valueString.EndsWith(
+                ConstraintArgument,
+                ignoreCase: true,
+                culture: CultureInfo.InvariantCulture);
```
Eilon: BTW you can use this overload: https://msdn.microsoft.com/en-us/library/ms131430(v=vs.110).aspx

It's a little less verbose.

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
```diff
@@ -44,6 +46,9 @@ public sealed class MatcherEndpoint : Endpoint
...
+            var mergedReferences = MergeMatchProcessorReferences(ParsedTemplate, nonInlineMatchProcessorReferences);
```
JamesNK: Will this change with https://github.com/aspnet/Routing/pull/585?

src/Microsoft.AspNetCore.Routing/Matchers/MatchProcessor.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public abstract class MatchProcessor
```
JamesNK: I'd like to revisit this name in the future.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultMatchProcessorFactory.cs
```diff
@@ -0,0 +1,157 @@
...
+                matchProcessorReference.Optional,
+                constraintType,
+                constraintArgument);
+
+            if (processor != null)
+            {
+                return processor;
+            }
+
+            if (!typeof(IRouteConstraint).GetTypeInfo().IsAssignableFrom(constraintType.GetTypeInfo()))
```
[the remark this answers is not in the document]
kichalla: You are right. They are not required. (Copy pasted code from somewhere else.)

src/Microsoft.AspNetCore.Routing/Matchers/MatchProcessor.cs
kichalla: Yeah right. It would end with "...ConstraintProcesor" I think.

src/Microsoft.AspNetCore.Routing/Matchers/MatchProcessorBase.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public abstract class MatchProcessorBase : MatchProcessor
```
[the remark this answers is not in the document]
kichalla: > It feels weird that MatchProcessorBase inherits from MatchProcessor

Yeah, now that you mention it, yeah it feels weird.

> Is the idea of this class as a helper for the majority of user implementations of MatchProcessor?

Yeah, right.

src/Microsoft.AspNetCore.Routing/Matchers/MatchProcessorBase.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public override bool ProcessOutbound(HttpContext httpContext, RouteValueDictionary values)
+        {
+            return Process(values);
+        }
+
+        private bool Process(RouteValueDictionary values)
+        {
+            if (!values.TryGetValue(ParameterName, out var value) || value == null)
```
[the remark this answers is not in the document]
kichalla: For most of our current `IRouteConstraints`, this is true, but yeah as guessed it might not be valid in all cases. In that case, users should directly inherit from `MatchProcessor` instead. (Example: `OptionalMatchProcessor.cs` file in this PR)

src/Microsoft.AspNetCore.Routing/Matchers/MatchProcessorReference.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        internal MatchProcessor MatchProcessor { get; private set; }
+
+        internal string ConstraintText { get; private set; }
+
+        internal string ParameterName { get; private set; }
+
+        internal bool Optional { get; private set; }
+
+        public static MatchProcessorReference From(string parameterName, string constraintText)
```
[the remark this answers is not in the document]
kichalla: They could be constructors too I think. Even in case of `RouteConstraintMatchProcessorAdapter`, I think its fine to have a constructor.

src/Microsoft.AspNetCore.Routing/Matchers/MatcherEndpoint.cs
kichalla: I kept this here so that we can be consistent with how `Defaults` are handled here. If we change for #585, both `Defaults` and `MatchProcessorReferences` would need to change.

kichalla:
@rynowak Do you have any further feedback on this PR?

samples/DispatcherSample.Web/Startup.cs
```diff
@@ -17,7 +19,13 @@ public class Startup
 
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddRouting();
+            services.AddTransient<EndsWithStringMatchProcessor>();
+
+            services.AddRouting(options =>
+            {
+                options.ConstraintMap.Add("endsWith", typeof(EndsWithStringMatchProcessor));
```
rynowak: How does this work This guy won't be an `IRouteConstraint`?

src/Microsoft.AspNetCore.Routing/Matchers/MatchProcessorReference.cs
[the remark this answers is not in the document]
rynowak: I think I made this suggestion originally. Call it gut feel on my part, it just felt very static-factory-ey to me. 

I don't feel very strongly about this design point - so if there's a good reason to do either one we should do that.

PR #585, opened 2018-07-09
Introduce RoutePattern

rynowak:
Introduces RoutePattern - a new parser and representation for routing
templates, defaults, and constraints.

This is a new representation for all of the 'inputs' to routing that is
immutable and captures 'out of line' information for defaults and
constraints.

This will allow us to unify the handling of constraints and values from
attribute style routes and conventional style routes.


rynowak:
/cc @kichalla and @JamesNK  - please take a look

The topic of how to deal with 'defaults' and constraints that are specified out of line keeps coming up in discussions 😆. It seemed worth it to me to introduce an object model that can represent all of these concepts.

The bulk of the this work was done during 2.1 and I'm just porting it forward because it seems like we'd benefit from it.

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
```diff
@@ -0,0 +1,245 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    internal static class RouteParameterParser
```
rynowak: This is not a great place to start reviewing. This code is a virtual cut/paste from existing code - and some of it is total spaghetti (not my doing lol).

I relayered the existing parser stuff to use the new stuff.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
```diff
@@ -0,0 +1,94 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    [DebuggerDisplay("{DebuggerToString()}")]
+    public sealed class RoutePattern
```
rynowak: This is a good place to start. There's a set of `RoutePattern*` classes introduced by this PR that make up the new object model.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -0,0 +1,698 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Routing.Constraints;
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    public static class RoutePatternFactory
+    {
+        public static RoutePattern Pattern(IEnumerable<RoutePatternPathSegment> segments)
```
JamesNK: Hmm, my first thought is I'd lean towards having `static Create(...)` methods on the various types, e.g. `var myNewRoutePattern = RoutePattern.Create(segements)`, instead of a central static help class

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParameterPart.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public bool IsOptional => ParameterKind == RoutePatternParameterKind.Optional;
+
+        public RoutePatternParameterKind ParameterKind { get; }
+
+        public string Name { get; }
+
+        internal override string DebuggerToString()
+        {
+            return RawText ?? "{" + (IsCatchAll ? "*" : string.Empty) + Name + (IsOptional ? "?" : string.Empty) + "}";
```
JamesNK: Since you're here, you could copy and paste this logic: https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/Internal/RouteTemplateWriter.cs

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternPart.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+
```
JamesNK: nit: too liney

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternPathSegment.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            RawText = rawText;
+            Parts = parts;
+        }
+
+        public bool IsSimple => Parts.Count == 1;
+
+        public IReadOnlyList<RoutePatternPart> Parts { get; }
+
+        public string RawText { get; set; }
```
JamesNK: Should this have a public setter?

src/Microsoft.AspNetCore.Routing/Template/TemplateSegment.cs
```diff
@@ -4,12 +4,22 @@
...
+        }
+
+        public TemplateSegment(RoutePatternPathSegment other)
+        {
+            Parts = new List<TemplatePart>(other.Parts.Select(s => new TemplatePart(s)));
+        }
+
         public bool IsSimple => Parts.Count == 1;
 
         public List<TemplatePart> Parts { get; } = new List<TemplatePart>();
```
JamesNK: Move this to the default `ctor()`? Avoids it being allocated twice in the other ctor.

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
rynowak: I don't really think so since we ultimately need to surface all of the substrings we create as `string`. Also this code runs at startup, if this is on the critical path for anyone then we haven't done our job 😆

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternException.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    public class RoutePatternException : Exception
```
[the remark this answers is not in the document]
rynowak: 👍

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public static RoutePattern Parse(string pattern)
+        {
+            try
+            {
+                return RoutePatternParser.Parse(pattern);
+            }
+            catch (RoutePatternException ex)
+            {
+                throw new ArgumentException(ex.Message, nameof(pattern), ex);
```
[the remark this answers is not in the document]
rynowak: Not sure really, I wrote this a long time ago. I'll undo this and see what happens.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
rynowak: The code we wrote for this in 2.1 was like this and upon review I thought it was horrible because you end up needing to know like 8 different types. If you're using these then you'd want to `static using` this class and it's pretty nice/simple (see the tests).

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParameterPart.cs
rynowak: Sure 👍

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternPathSegment.cs
rynowak: Good catch, this should not be settable.

src/Microsoft.AspNetCore.Routing/Template/TemplateSegment.cs
rynowak: Sure

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternConstraintReference.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    /// <summary>
+    /// The parsed representation of a constraint in a <see cref="RoutePattern"/> parameter.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerToString()}")]
+    public sealed class RoutePatternConstraintReference
+    {
+        internal RoutePatternConstraintReference(string rawText, string name, string content)
```
kichalla: Assign name to Name property. Could you also add a sample in a comment here so that its clear (mainly how `rawText` and `content` differ is not clear for me)

kichalla: Nit: name => parameterName

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
kichalla: I would prefer having having the construction of RoutePattern in one place, in this case RoutePatternFactory where I supply 'rawtext and non-inline defaults and constraints' and removing the creation methods on `RoutePattern` class. Some of these methods can be internal I think.

rynowak: If you guys are really sad about this btw we can make it internal. 

@kichalla  if I understand your suggestion are you asking for me to move the `Parse` methods to this class?

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternConstraintReference.cs
rynowak: I've been using the property/variable name `name` for parameter name. I can change it here but I will also change it in a few other places. @JamesNK @kichalla thoughts?

Also probably `text` vs `rawText`

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
```diff
@@ -0,0 +1,573 @@
...
+
+            var inside = text.Substring(1, text.Length - 2);
+            var decoded = inside.Replace("}}", "}").Replace("{{", "{");
+
+            // At this point, we need to parse the raw name for inline constraint,
+            // default values and optional parameters.
+            var templatePart = RouteParameterParser.ParseRouteParameter(text, decoded);
+
+            // See #475 - this is here because InlineRouteParameterParser can't return errors
+            if (decoded.StartsWith("*") && decoded.EndsWith("?"))
```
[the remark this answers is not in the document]
rynowak: Thanks. Did a search and fixed a few locations.

rynowak:
Updated YO

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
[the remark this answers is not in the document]
JamesNK: `decoded.Length > 0 && decoded[0] == '*' && decoded[decoded.Length - 1] == '?'`?

Will probably be slightly faster.

JamesNK:
Should this be targeting 2.2 branch?

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
```diff
@@ -0,0 +1,246 @@
...
+    {
+        // This code parses the inside of the route parameter 
+        //
+        // Ex: {hello} - this method is responsible for parsing 'hello'
+        // The factoring between this class and RoutePatternParser is due to legacy.
+        public static RoutePatternParameterPart ParseRouteParameter(string text, string parameter)
```
davidfowl: Got benchmarks for this?

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
```diff
@@ -0,0 +1,246 @@
...
+            while (currentIndex <= endIndex)
+            {
+                var currentChar = parameter[currentIndex];
+
+                if ((currentChar == ':' || currentChar == '=') && startIndex != currentIndex)
+                {
+                    // Parameter names are allowed to start with delimiters used to denote constraints or default values.
+                    // i.e. "=foo" or ":bar" would be treated as parameter names rather than default value or constraint
+                    // specifications.
+                    parameterName = parameter.Substring(startIndex, currentIndex - startIndex);
```
davidfowl: Any benefit using StringSegment here?

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
davidfowl: If we're going to make a new thing public, we should make sure it's using all the right types (Span etc).

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternConstraintReference.cs
kichalla: > I've been using the property/variable name name for parameter name. I can change it here but I will also change it in a few other places. @JamesNK @kichalla thoughts?

Sure 👍

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
kichalla: > @kichalla if I understand your suggestion are you asking for me to move the Parse methods to this class?

Yes @rynowak

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
JamesNK: Route patterns (templates) are created once on startup and then live for the lifetime of the app. I don't see Span providing any significant benefit.

RoutePattern should expose whatever is the fastest for matches to be made with. I'm guessing that is `string` but @rynowak has been working on the optimizing matching and would know best.

davidfowl: The fact that it's a public API makes me think about usages outside of ours (once at startup).

rynowak:
The code I'm adding here runs at startup to parse route templates. We've had infrastructure like this for a few years now without it coming up as perf concern, so I'm ignoring all perf-related feedback.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
rynowak: this code plays no role in URL matching

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
rynowak: > if this is on the critical path for anyone then we haven't done our job

davidfowl:
> We've had infrastructure like this for a few years now without it coming up as perf concern, so I'm ignoring all perf-related feedback.

I'm not sure why. Then why make this a new top level public low level route parsing API? Why not internal if we only care about this one use case?

rynowak:
This code parses and validates the template, it plays no role in matching URLs.

rynowak:
I've updated this with @kichalla naming feedback and changed the base branch (thanks @JamesNK).

Regarding the (*other*) feedback. It seems like the main point of discussion here is where the object model (`RoutePattern`, `RoutePatternPathSegment`, `RoutePatternParameterPart`) expose spans or strings. If you're giving this feedback please take a look at these classes and let me know if you think span belongs here, and what benefits that will add. Is span a good fit for these APIs?

If you think that Span does belong here, then let me know if removing 'RawText` from all of the various pieces would change your mind.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternConstraintReference.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    /// <summary>
+    /// The parsed representation of a constraint in a <see cref="RoutePattern"/> parameter.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerToString()}")]
+    public sealed class RoutePatternConstraintReference
+    {
+        internal RoutePatternConstraintReference(string rawText, string parameterName, string content)
```
kichalla: Set the parameterName to the property

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternConstraintReference.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            RawText = rawText;
+            Content = content;
+        }
+
+        internal RoutePatternConstraintReference(string name, IRouteConstraint constraint)
```
kichalla: nit: parameterName

rynowak:
@JamesNK @davidfowl - do you have additional thoughts about [this](https://github.com/aspnet/Routing/pull/585#issuecomment-403899783)?

JamesNK:
My original comment about using Span was about eliminating any intermediary strings. I like exposing String publicly.

rynowak:
> My original comment about using Span was about eliminating any intermediary strings. I like exposing String publicly. 

OK cool. If we find that the perf of the parsing matters, we could always retrofit the parsing code to use spans for fewer substrings.

I think I'll also take the step of removing `RawText` from most of the APIs (I'll leave it on route pattern).  There's no code that actually consumes rawtext other than on RoutePattern and it would make it more beneficial to use span if we need to later.

rynowak:
@JamesNK - any other feedback about this?

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -0,0 +1,517 @@
...
+            {
+                throw new ArgumentNullException(nameof(parts));
+            }
+
+            return SegmentCore(parts);
+        }
+
+        private static RoutePatternPathSegment SegmentCore(IEnumerable<RoutePatternPart> parts)
+        {
+            return new RoutePatternPathSegment(parts.ToArray());
```
JamesNK: `parts` could sometimes already be an array (I don't know how deep you want to go on perf at the moment)

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -0,0 +1,517 @@
...
+        {
+            if (string.IsNullOrEmpty(content))
+            {
+                throw new ArgumentException(Resources.Argument_NullOrEmpty, nameof(content));
+            }
+
+            return SeparatorPartCore(content);
+        }
+
+        private static RoutePatternSeparatorPart SeparatorPartCore( string content)
```
JamesNK: nit: extra space

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -0,0 +1,517 @@
...
+        {
+            return new RoutePatternSeparatorPart( content);
```
JamesNK: nit: extra space

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -0,0 +1,517 @@
...
+        public static RoutePatternParameterPart ParameterPart(string parameterName)
+        {
+            if (string.IsNullOrEmpty(parameterName))
+            {
+                throw new ArgumentException(Resources.Argument_NullOrEmpty, nameof(parameterName));
+            }
+
+            if (parameterName.IndexOfAny(RoutePatternParser.InvalidParameterNameChars) >= 0)
+            {
+                throw new ArgumentException(Resources.FormatTemplateRoute_InvalidParameterName(parameterName));
```
JamesNK: `if (string.IsNullOrEmpty(parameterName))` and `if (parameterName.IndexOfAny(RoutePatternParser.InvalidParameterNameChars) >= 0)` are checked in `ParameterPartCore` and all its `ParameterPart` callers.

Don't need to duplicate check.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternMatcher.cs
```diff
@@ -0,0 +1,511 @@
...
+            RoutePattern pattern,
+            RouteValueDictionary defaults)
+        {
+            if (pattern == null)
+            {
+                throw new ArgumentNullException(nameof(pattern));
+            }
+
+            RoutePattern = pattern;
+            Defaults = defaults ?? new RouteValueDictionary();
```
JamesNK: Allowing a null defaults feels weird.

Maybe a ctor that takes just pattern and another that takes both and checks defaults to not be null?

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
```diff
@@ -0,0 +1,573 @@
...
+                {
+                    throw new RoutePatternException(pattern, context.Error);
+                }
+
+                // A successful parse should always result in us being at the end or at a separator.
+                Debug.Assert(context.AtEnd() || context.Current == Separator);
+
+                if (context.Index <= i)
+                {
+                    throw new InvalidProgramException("Infinite loop in the parser. This is a bug.");
```
JamesNK: I don't like "This is a bug." in error message to user

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
```diff
@@ -0,0 +1,573 @@
...
+
+                if (context.Current == Separator || context.AtEnd())
+                {
+                    // We've reached the end of the segment
+                    break;
+                }
+
+                if (context.Index <= i)
+                {
+                    throw new InvalidProgramException("Infinite loop in the parser. This is a bug.");
```
JamesNK: ^

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
```diff
@@ -0,0 +1,573 @@
...
+        }
+
+        [DebuggerDisplay("{DebuggerToString()}")]
+        private class TemplateParserContext
+        {
+            private readonly string _template;
+            private int _index;
+            private int? _mark;
+
+            private HashSet<string> _parameterNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
JamesNK: readonly

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
```diff
@@ -0,0 +1,573 @@
...
+                }
+            }
+
+            // if a segment has multiple parts, then only the last one parameter can be optional 
+            // if it is following a optional seperator. 
+            for (var i = 0; i < parts.Count; i++)
+            {
+                var part = parts[i];
+
+                if (part.IsParameter && ((RoutePatternParameterPart)part).IsOptional && parts.Count > 1)
```
JamesNK: Part is cast to `RoutePatternParameterPart` many times in the code below. Create a cast variable?

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternPart.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    public abstract class RoutePatternPart
+    {
+        // This class is **not** an extensibility point - every part of the routing system
+        // needs to be aware of what kind of parts we support.
+        //
+        // It is abstract so we can add semantics later inside the library.
+        internal RoutePatternPart(RoutePatternPartKind partKind)
```
JamesNK: Did the private AND internal keyword ever get added to C#? A candidate to use it right here 😝

src/Microsoft.AspNetCore.Routing/Template/TemplatePart.cs
```diff
@@ -5,12 +5,47 @@
...
+            }
+            else if (other.IsSeparator && other is RoutePatternSeparatorPart separator)
+            {
+                Text = separator.Content;
+                IsOptionalSeperator = true;
+            }
+            else
+            {
+                // Unreachable
+                throw new NotSupportedException();
```
JamesNK: Is this an ok thing to do or was it done in a hurry?

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateParserTests.cs
```diff
@@ -528,8 +525,8 @@ public void Parse_RegularExpressions_Unescaped(string template)
 
         [Theory]
         [InlineData("{p1}.{p2?}.{p3}", "p2", ".")]
-        [InlineData("{p1?}{p2}", "p1", "p2")]
-        [InlineData("{p1?}{p2?}", "p1", "p2")]
+        [InlineData("{p1?}{p2}", "p1", "{p2}")]
+        [InlineData("{p1?}{p2?}", "p1", "{p2?}")]
```
JamesNK: How come you changed these?

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
```diff
@@ -0,0 +1,69 @@
...
+            Debug.Assert(pathSegments != null);
+
+            RawText = rawText;
+            Defaults = defaults;
+            Constraints = constraints;
+            Parameters = parameters;
+            PathSegments = pathSegments;
+        }
+
+        public IReadOnlyDictionary<string, object> Defaults { get; }
```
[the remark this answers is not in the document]
JamesNK: And collections on other patterns classes

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
```diff
@@ -0,0 +1,246 @@
...
+            var startIndex = currentIndex;
+            do
+            {
+                var currentChar = currentIndex > endIndex ? null : (char?)text[currentIndex];
+                switch (state)
+                {
+                    case ParseState.Start:
+                        switch (currentChar)
+                        {
+                            case null:
```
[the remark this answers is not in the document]
rynowak: I'm really really unconcerned about perf in this code until we have a scenario that dictates that it matters. Each route template is parsed once at startup and never again.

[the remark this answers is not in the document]
rynowak: Another meta comment about this. All of this code is rename from existing code. I don't mind addressing minor feedback but if we were going to spend time here I would throw it away and start over. I don't think rewriting it is anywhere near a priority.

This code was written in a pretty crazy way by someone who was sulking about being made to redo it after they were told not to use a regex for this. Seriously.

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
```diff
@@ -0,0 +1,246 @@
...
+            ParsingName,
+            InsideParenthesis,
+            End
+        }
+
+        private struct ConstraintParseResults
+        {
+            public int CurrentIndex;
+
+            public IEnumerable<RoutePatternConstraintReference> Constraints;
```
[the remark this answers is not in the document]
rynowak: sure

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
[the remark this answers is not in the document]
rynowak: cool beans 👍

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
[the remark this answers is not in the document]
rynowak: I'm not concerned at all. To me that's equivalent to doing private reflection.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
```diff
@@ -0,0 +1,69 @@
...
+        public IReadOnlyList<RoutePatternParameterPart> Parameters { get; }
+
+        public IReadOnlyList<RoutePatternPathSegment> PathSegments { get; }
+
+        /// <summary>
+        /// Gets the parameter matching the given name.
+        /// </summary>
+        /// <param name="name">The name of the parameter to match.</param>
+        /// <returns>The matching parameter or <c>null</c> if no parameter matches the given name.</returns>
+        public RoutePatternParameterPart GetParameter(string name)
```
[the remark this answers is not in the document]
rynowak: Sure

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternException.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public RoutePatternException(string pattern, string message)
+            : base(message)
+        {
+            if (pattern == null)
+            {
+                throw new ArgumentNullException(nameof(pattern));
+            }
+
+            if (message == null)
```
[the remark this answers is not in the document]
rynowak: Makes sense. Usually if something is optional, then it will tolerate null.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -0,0 +1,517 @@
...
+                }
+            }
+
+            var updatedConstraints = new Dictionary<string, List<RoutePatternConstraintReference>>(StringComparer.OrdinalIgnoreCase);
+            if (constraints != null)
+            {
+                foreach (var kvp in constraints)
+                {
+                    updatedConstraints.Add(kvp.Key, new List<RoutePatternConstraintReference>()
+                {
```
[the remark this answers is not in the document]
rynowak: sure

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -0,0 +1,517 @@
...
+            for (var i = 0; i < updatedSegments.Length; i++)
+            {
+                updatedSegments[i] = VisitSegment(updatedSegments[i]);
+            }
+
+            var parameters = new List<RoutePatternParameterPart>();
+            for (var i = 0; i < updatedSegments.Length; i++)
+            {
+                var segment = updatedSegments[i];
+                for (var j = 0; j < segment.Parts.Count; j++)
```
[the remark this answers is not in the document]
rynowak: seems fine

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -0,0 +1,517 @@
...
+                    {
+                        parameters.Add(parameter);
+                    }
+                }
+            }
+
+            return new RoutePattern(
+                rawText,
+                updatedDefaults,
+                updatedConstraints.ToDictionary(kvp => kvp.Key, kvp => (IReadOnlyList<RoutePatternConstraintReference>)kvp.Value.ToArray()),
```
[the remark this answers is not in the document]
rynowak: I don't have a scenario where this code is perf-sensitive.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
rynowak: I'd still want to make a defensive copy if the user called `Segment(myArray)`

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
rynowak: heh

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
rynowak: wierd

rynowak: I wonder if the editor did that when I removed the `rawText` parameter

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
rynowak: We generally put all argument validation as close as possible to the method the user called (see null checks all over the codebase).

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternMatcher.cs
rynowak: See your comment else where about `Exception(string message)`. It's normal to accept null for a parameter that's optional. 

I think the methods in MVC for `MapRoute` also will tolerate null, so it would require extra work all over the code 😆

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternMatcher.cs
```diff
@@ -0,0 +1,511 @@
...
+        private const string SeparatorString = "/";
+        private const char SeparatorChar = '/';
+
+        // Perf: This is a cache to avoid looking things up in 'Defaults' each request.
+        private readonly bool[] _hasDefaultValue;
+        private readonly object[] _defaultValues;
+
+        private static readonly char[] Delimiters = new char[] { SeparatorChar };
+
+        public RoutePatternMatcher(
```
rynowak: FYI this class is just a rename of existing code with a few details change to use `RoutePattern`.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
```diff
@@ -0,0 +1,573 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    internal static class RoutePatternParser
```
rynowak: This is also another rename of an existing class, the changes here are to support the `RoutePattern` instead of `RouteTemplate`.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
rynowak: What would you suggest? This is not a message a user should be able to see. I want this code to crash hard if there's an infinite loop.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
rynowak: Again, what do you suggest?

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
rynowak: This can probably be simplified now that we have pattern patching.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternPart.cs
rynowak: C# 7.2! DOPE

src/Microsoft.AspNetCore.Routing/Template/TemplatePart.cs
rynowak: It should be unreachable code. What would you suggest?

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateParserTests.cs
rynowak: I changed what gets passed to the error messages. now instead of just including the parameter name or an unrelated thing, it includes the text of the next part..

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
rynowak: fix't

rynowak:
Updated and addressed comments so far

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
JamesNK: In that case you could leave this check (and in other overloads) and remove the one in ParameterPartCore.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
JamesNK: I would have something like "An unexpected infinite loop detected in the parser." and then a comment saying we should never reach here

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParser.cs
JamesNK: ^

src/Microsoft.AspNetCore.Routing/Template/TemplatePart.cs
JamesNK: If you're fine with it, I'm fine with it. Just checking

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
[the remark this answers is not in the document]
rynowak: This is fixed btw, just not showing on in the diff.

PR #591, opened 2018-07-12
[WIP] Productize the DFA matcher

rynowak:
Creating a PR for early feedback. I realize that there's a daunting amount of code and any details to figure out here. Most importantly I need to resolve some of the things I started here that are being worked on by others 😆 

If there are pieces of this that seem good, I can snap those off into their own PRs to make it easier. I figure it's better to show all of the pieces coming together while I do that.

I'll use this top comment to track the major items that need to be resolved before this is ready.
- [x] Support Endpoint Constraints
- [x] Unify with checked in work for route-constraint-likes
- [x] Support ordering entries by arbitrary criteria
- [x] Replace TreeMatcher with this implementation (make this the default)
- [x] Design for CandidateSet

I'll add to this like as y'all find out more stuff I'm missing.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matchers/Candidate.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matchers/Candidate.cs
@@ -34,6 +34,9 @@ namespace Microsoft.AspNetCore.Routing.Matchers
 
         public readonly MatchProcessor[] MatchProcessors;
 
+        // Data for EndpointSelectorPolicy - indexed by policy index in the DFA matcher
+        public readonly object[] PolicyData;
+
         // Used in tests.
         public Candidate(MatcherEndpoint endpoint)
         {
@@ -44,6 +47,7 @@ namespace Microsoft.AspNetCore.Routing.Matchers
             CatchAll = default;
             ComplexSegments = Array.Empty<(RoutePatternPathSegment pathSegment, int segmentIndex)>();
             MatchProcessors = Array.Empty<MatchProcessor>();
+            PolicyData = Array.Empty<object>();
 
             Flags = CandidateFlags.None;
         }
@@ -54,7 +58,8 @@ namespace Microsoft.AspNetCore.Routing.Matchers
             (string parameterName, int segmentIndex, int slotIndex)[] captures,
             (string parameterName, int segmentIndex, int slotIndex) catchAll,
             (RoutePatternPathSegment pathSegment, int segmentIndex)[] complexSegments,
-            MatchProcessor[] matchProcessors)
+            MatchProcessor[] matchProcessors,
+            object[] policyData)
         {
             Endpoint = endpoint;
             Slots = slots;
@@ -62,6 +67,7 @@ namespace Microsoft.AspNetCore.Routing.Matchers
             CatchAll = catchAll;
             ComplexSegments = complexSegments;
             MatchProcessors = matchProcessors;
+            PolicyData = policyData;
 
             Flags = CandidateFlags.None;
             for (var i = 0; i < slots.Length; i++)
--- a/src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
@@ -2,10 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.AspNetCore.Routing.Matchers
 {
-    internal class CandidateSet
+    internal class CandidateSet : IReadOnlyList<MatcherEndpoint>
     {
         public static readonly CandidateSet Empty = new CandidateSet(Array.Empty<Candidate>(), Array.Empty<int>());
 
@@ -35,6 +38,20 @@ namespace Microsoft.AspNetCore.Routing.Matchers
             Groups = groups;
 
             GroupCount = groups.Length == 0 ? 0 : groups.Length - 1;
+        }
+
+        public MatcherEndpoint this[int index] => Candidates[index].Endpoint;
+
+        public int Count => Candidates.Length;
+
+        public IEnumerator<MatcherEndpoint> GetEnumerator()
+        {
+            return Candidates.Select(c => c.Endpoint).GetEnumerator();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
         }
 
         // See description on Groups.
--- a/src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
@@ -8,19 +8,18 @@ using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Routing.EndpointConstraints;
 using Microsoft.AspNetCore.Routing.Patterns;
 
 namespace Microsoft.AspNetCore.Routing.Matchers
 {
     internal sealed class DfaMatcher : Matcher
     {
-        private readonly EndpointSelector _endpointSelector;
+        private readonly EndpointSelectorPolicy[] _policies;
         private readonly DfaState[] _states;
 
-        public DfaMatcher(EndpointSelector endpointSelector, DfaState[] states)
+        public DfaMatcher(EndpointSelectorPolicy[] policies, DfaState[] states)
         {
-            _endpointSelector = endpointSelector;
+            _policies = policies;
             _states = states;
         }
 
@@ -70,6 +69,11 @@ namespace Microsoft.AspNetCore.Routing.Matchers
             var candidatesArray = candidates.Candidates;
             var groups = candidates.Groups;
 
+            // We need to keep track of which policies have rejected endpoints while
+            // we process policies. This allows us to call back into those policies
+            // after processing all groups.
+            var rejectionState = new bool[_policies.Length];
+
             for (var i = 0; i < groups.Length - 1; i++)
             {
                 var start = groups[i];
@@ -93,37 +97,57 @@ namespace Microsoft.AspNetCore.Routing.Matchers
                     groupValues))
                 {
                     // We must have some matches because FilterGroup returned true.
+                    // We need this here in case we don't have any policies to execute.
+                    var hasMatches = true;
 
-                    // So: this code is SUPER SUPER temporary. We don't intent to keep
-                    // EndpointSelector around for very long.
-                    var candidatesForEndpointSelector = new List<Endpoint>();
-                    for (var j = 0; j < group.Length; j++)
+                    // This group has some matches, so call into endpoint selector policies.
+                    for (var j = 0; j < _policies.Length; j++)
                     {
-                        if (members.Get(j))
+                        var policy = _policies[j];
+
+                        // Initialize this on the leading edge of the loop, so we can track per-policy
+                        // whether the policy eliminated candidates.
+                        hasMatches = false;
+
+                        for (var k = 0; k < group.Length; k++)
                         {
-                            candidatesForEndpointSelector.Add(group[j].Endpoint);
+                            var isMatch = members.Get(k) && policy.Match(
+                                httpContext,
+                                group[k].Endpoint,
+                                group[k].PolicyData[j],
+                                groupValues[k]);
+
+                            hasMatches |= isMatch;
+                            members.Set(k, isMatch);
+                        }
+
+                        if (!hasMatches)
+                        {
+                            // This policy has rejected some possible matches. We keep track
+                            // of which policies reject matches so we can ask how to handle
+                            // the failure if no endpoint matches.
+                            rejectionState[i] |= true;
+                            break;
                         }
                     }
 
-                    var result = _endpointSelector.SelectBestCandidate(httpContext, candidatesForEndpointSelector);
-                    if (result != null)
+                    if (hasMatches)
                     {
-                        // Find the route values, based on which endpoint was selected. We have
-                        // to do this because the endpoint selector returns an endpoint
-                        // instead of mutating the feature.
-                        for (var j = 0; j < group.Length; j++)
-                        {
-                            if (ReferenceEquals(result, group[j].Endpoint))
-                            {
-                                feature.Endpoint = result;
-                                feature.Invoker = ((MatcherEndpoint)result).Invoker;
-                                feature.Values = groupValues[j];
-                                return Task.CompletedTask;
-                            }
-                        }
+                        SelectBestCandidate(feature, group, members, groupValues);
+                        return Task.CompletedTask;
                     }
+                }
+            }
 
-                    // End super temporary code
+            // If after all groups we don't yet have any matches, then check if
+            // any matches rejected by policies, and ask those policies how to treat
+            // the failure.
+            for (var i = _policies.Length - 1; i >=0; i--)
+            {
+                _policies[i].Reject(httpContext, feature, candidates);
+                if (feature.Invoker != null)
+                {
+                    break;
                 }
             }
 
@@ -285,6 +309,69 @@ namespace Microsoft.AspNetCore.Routing.Matchers
             }
 
             return true;
+        }
+
+        private void SelectBestCandidate(
+            IEndpointFeature feature,
+            ReadOnlySpan<Candidate> group,
+            BitArray members,
+            RouteValueDictionary[] groupValues)
+        {
+            var found = false;
+            for (var i = 0; i < group.Length; i++)
+            {
+                if (members.Get(i) && found)
+                {
+                    // This is the second match we've found, so there must be an ambiguity.
+                    feature.Endpoint = null;
+                    feature.Invoker = null;
+                    feature.Values = null;
+                    break;
+                }
+                else if (members.Get(i))
+                {
+                    feature.Endpoint = group[i].Endpoint;
+                    feature.Invoker = group[i].Endpoint.Invoker;
+                    feature.Values = groupValues[i];
+                    return;
+                }
+            }
+
+            if (found)
+            {
+                // If we get here it's the result of an ambiguity.
+                var matches = new List<MatcherEndpoint>();
+                for (var i = 0; i < group.Length; i++)
+                {
+                    if (members.Get(i))
+                    {
+                        matches.Add(group[i].Endpoint);
+                    }
+                }
+
+                var message =
+                    "The request matched multiple endpoints. Matches:" + Environment.NewLine +
+                    string.Join(Environment.NewLine, matches.Select(e => e.DisplayName));
+                throw new AmbiguousMatchException(message);
+            }
+        }
+
+        [DebuggerDisplay("{DebuggerToString(),nq}")]
+        public readonly struct State
+        {
+            public readonly CandidateSet Candidates;
+            public readonly JumpTable Transitions;
+
+            public State(CandidateSet candidates, JumpTable transitions)
+            {
+                Candidates = candidates;
+                Transitions = transitions;
+            }
+
+            public string DebuggerToString()
+            {
+                return $"m: {Candidates.Candidates?.Length ?? 0}, j: ({Transitions?.DebuggerToString()})";
+            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
@@ -4,7 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.AspNetCore.Routing.EndpointConstraints;
+using Microsoft.AspNetCore.Routing.Constraints;
 using Microsoft.AspNetCore.Routing.Patterns;
 using Microsoft.AspNetCore.Routing.Template;
 using Microsoft.Extensions.Options;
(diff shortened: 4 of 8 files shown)


benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/JumpTableMultipleEntryBenchmark.cs
```diff
@@ -0,0 +1,454 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
```
rynowak: The benchmarks are not a great place to start reviewing this code. #YOLO

src/Microsoft.AspNetCore.Routing/Matchers/MatcherBase.cs
```diff
@@ -0,0 +1,79 @@
...
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            if (feature == null)
+            {
+                throw new ArgumentNullException(nameof(feature));
+            }
+
+            var path = httpContext.Request.Path.Value;
+            var buffer = stackalloc PathSegment[32];
```
JamesNK: Move array size to const. Add comment explaining why 32

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesAzureBenchmark.cs
```diff
@@ -0,0 +1,97 @@
...
+        [Benchmark(OperationsPerInvoke = SampleCount)]
+        public unsafe void Dfa()
+        {
+            for (var i = 0; i < SampleCount; i++)
+            {
+                var sample = _samples[i];
+                var httpContext = _requests[sample];
+
+                var path = httpContext.Request.Path.Value;
+                var buffer = stackalloc PathSegment[32];
```
JamesNK: Use shared const for array size across benchmarks

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/TrivialMatcher.cs
```diff
@@ -10,16 +10,27 @@ namespace Microsoft.AspNetCore.Routing.Matchers
...
+            {
+                new Candidate(endpoint, Array.Empty<MatchProcessor>()),
+            };
+
+            _candidateIndices = new int[] { 0, };
+            _candidateGroups = new int[] { 1, };
         }
 
-        public override Task MatchAsync(HttpContext httpContext, IEndpointFeature feature)
+        public new Task MatchAsync(HttpContext httpContext, IEndpointFeature feature)
```
JamesNK: Why is `new` needed here?

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        public unsafe override int GetDestination(string path, PathSegment segment)
+        {
+            if (segment.Length == 0)
+            {
+                return _exit;
+            }
+
+            var c = path[segment.Start];
+            if ((c & ~0x7F) != 0)
```
JamesNK: Move value to const variable and/or comment explaining what it is

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                return _default;
+            }
+
+            c = (char)(c & 0x5F);
```
JamesNK: Move value to const variable and/or comment explaining what it is

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
```diff
@@ -0,0 +1,161 @@
...
+                return _default;
+            }
+
+            var offset = (c - 'A') * 2;
+            var start = _indexes[offset];
+            var length = _indexes[offset + 1];
+
+            for (var i = start; i < start + length; i++)
+            {
+                ref var entry = ref _entries[i];
```
JamesNK: Is this ref for performance? Worth commenting?

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
```diff
@@ -0,0 +1,161 @@
...
+                if (segment.Length == entry.text.Length &&
+                    string.Compare(
```
JamesNK: Does `string.Equals` not support these arguments? I would guess an Equals would be slightly faster (if available)

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
JamesNK: Use `ToUpperAscii`?

src/Microsoft.AspNetCore.Routing/Matchers/Candidate.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    internal readonly struct Candidate
+    {
+        public Candidate(Endpoint endpoint, MatchProcessor[] segments)
+        {
+            Endpoint = endpoint;
```
JamesNK: `Debug.Assert` not null these arguments?

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
```diff
@@ -0,0 +1,27 @@
...
+        public string Path;
+        public ReadOnlySpan<PathSegment> Segments;
+        public ReadOnlySpan<Candidate> Candidates;
+        public ReadOnlySpan<int> CandidateIndices;
+        public ReadOnlySpan<int> CandidateGroups;
+        public Span<RouteValueDictionary> Values;
+
+        public CandidateSet(string path, ReadOnlySpan<PathSegment> segments)
+        {
+            Path = path;
```
JamesNK: `Debug.Assert` not null these arguments?

Edit: Now that I've said that in a couple of comments I can see places for `Debug.Assert`s everywhere, which wouldn't be wanted. I guess this is just a prompt to make sure they are where needed.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    internal ref struct CandidateSet
+    {
+        public string Path;
+        public ReadOnlySpan<PathSegment> Segments;
```
JamesNK: Can Path/Segments be readonly?

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
```diff
@@ -0,0 +1,55 @@
...
+            return current;
+        }
+
+        [DebuggerDisplay("{DebuggerToString(),nq}")]
+        public struct State
+        {
+            public Candidate[] Candidates;
+            public int[] CandidateIndices;
+            public int[] CandidateGroups;
+            public JumpTable Transitions;
```
JamesNK: Do these need to be mutable or are they mutable for convenience?

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            public string DebuggerToString()
+            {
+                return $"m: {CandidateIndices?.Length ?? 0}, j: ({Transitions?.DebuggerToString()})";
```
JamesNK: Improve?

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -0,0 +1,436 @@
...
+        private static int[] CreateCandidateGroups(DfaNode node)
+        {
+            if (node.Matches.Count == 0)
+            {
+                return Array.Empty<int>();
+            }
+
+            var groups = new List<int>();
+
+            var order = node.Matches[0].Order;
```
JamesNK: Add a comment that all matches on node have the same order, etc.

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -0,0 +1,436 @@
...
+            for (var i = depth; i < entry.Pattern.Segments.Count; i++)
+            {
+                var segment = entry.Pattern.Segments[i];
+                if (!segment.IsSimple)
+                {
+                    // Complex segments always require more processing
+                    return true;
+                }
+
+                if (segment.Parts[0].IsLiteral)
```
JamesNK: Assign `segment.Parts[0]` to a variable?

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
JamesNK: I guess it is internal so it doesn't really matter

src/Microsoft.AspNetCore.Routing/Matchers/ParameterSegmentMatchProcessor.cs
```diff
@@ -0,0 +1,65 @@
...
+        public override bool Process(
+            HttpContext httpContext,
+            string path,
+            ReadOnlySpan<PathSegment> segments,
+            RouteValueDictionary values)
+        {
+            if (segments.Length > _segment && _isCatchAll)
+            {
+                var segment = segments[_segment];
+                values[_name] = path.Substring(segment.Start, path.Length - segment.Start);
```
JamesNK: Hmmm, has there ever been thought about making this Substring be lazy? The RVD keeps an array segment over the path and only does the Substring when the value is used by something.

I'm guessing not worth the complexity for such a small saving.

src/Microsoft.AspNetCore.Routing/Matchers/RouteConstraintMatchProcessor.cs
```diff
@@ -0,0 +1,45 @@
...
+        public override bool Process(
+            HttpContext httpContext,
+            string path,
+            ReadOnlySpan<PathSegment> segments,
+            RouteValueDictionary values)
+        {
+            return _constraint.Match(httpContext, NullRouter.Instance, _name, values, RouteDirection.IncomingRequest);
+        }
+
+        private class NullRouter : IRouter
```
JamesNK: I think @kichalla has or is in the process or splitting this out into an internal class. Should reuse.

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
```diff
@@ -0,0 +1,161 @@
...
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    // An optimized jump table that trades a small amount of additional memory for
+    // hash-table like performance.
+    //
+    // The optimization here is to use the first character of the known entries
+    // as a 'key' in the hash table in the space of A-Z. This gives us a maximum
+    // of 26 buckets (hence the reduced memory)
+    internal class AsciiKeyedJumpTable : JumpTable
```
rynowak: BTW - I'm not totally confident in this guy yet, I'm not sure that it's completely worth doing vs other approaches. Basically, I think it needs a little more investigation...

rynowak: It's better than a linear search, once you have 3-4 candidates, but worse than approaches based on true hash tables.

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
rynowak: String.Equals does not support ranges. Span does, but based on my testing it's slower.

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        public static bool TryCreate(
+            int @default,
```
JamesNK: These are indexes being passed in? Worth suffixing with Index?

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
rynowak: meta: should we just put a blanket "we did the perf, you should not refactor this code unless you have a reason" on all of the perf sensitive code?

The explanation for most things will be that we benchmarked it and did the fastest thing.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/InstructionMatcherBuilder.cs
```diff
@@ -139,21 +124,31 @@ private OrderNode GetOrCreateRootNode(List<OrderNode> roots, int order)
...
-                Endpoint = endpoint,
-                Parameters = parsed.Segments.Select(s => s.IsSimple && s.Parts[0].IsParameter ? s.Parts[0].Name : null).ToArray(),
-            };
-        }
+                var segment = parsed.Segments[i];
+                if (segment.IsSimple && segment.Parts[0].IsParameter)
+                {
+                    var processor = new ParameterSegmentMatchProcessor(
+                        i,
+                        segment.Parts[0].Name,
```
JamesNK: Assign `segment.Parts[0]` to variable?

src/Microsoft.AspNetCore.Routing/Matchers/ParameterSegmentMatchProcessor.cs
```diff
@@ -0,0 +1,65 @@
...
+            _defaultValue = defaultValue;
+        }
+
+        public override bool Process(
+            HttpContext httpContext,
+            string path,
+            ReadOnlySpan<PathSegment> segments,
+            RouteValueDictionary values)
+        {
+            if (segments.Length > _segment && _isCatchAll)
```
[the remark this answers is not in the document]
JamesNK: https://github.com/aspnet/Routing/pull/591#discussion_r201910412

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
JamesNK: Could make it adaptive? Different strategies depending on the number of candidates

test/Microsoft.AspNetCore.Routing.Tests/Matchers/InstructionMatcher.cs
```diff
@@ -1,16 +1,14 @@
...
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
-using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.Routing.Matchers
 {
-    internal class InstructionMatcher : Matcher
+    internal class InstructionMatcher : MatcherBase
```
JamesNK: What are your plans for when to remove `InstructionMatcher`?

rynowak: It might be a good idea to just remove it now. I've been realizing as I do other thing that I keep finding more places to work on it and keep it building/running/passing etc. I'll create another PR that removes it.

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
rynowak: Yes, it is actually, see `JumpTableMatcherBuilder`. What I mean is that is that the sweet spot for this is really small - it may not be worth the complexity.

Additionally, this looks really good on small artificial benchmarks, but offers much less benefit for something like Azure or github's route table. 

If we ended up with an adaptive builder where:
0: special case
1: special case
2-4: linear search
5-8??: ASCII
9+: Dictionary

Then it seems worth doing. I suspect if we put more time into it, the dictionary/hash based implementation will have it's floor pushed down way further (potentially to 2-3).

src/Microsoft.AspNetCore.Routing/Matchers/MatcherBase.cs
rynowak: Fixed

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesAzureBenchmark.cs
rynowak: Fixed here and a bunch of places 👍

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/TrivialMatcher.cs
rynowak: The base class implementation is `sealed` by design. Adding a comment

rynowak: Actually I'll just remove the inheritance. There's no need for `MatcherBase` to exist once the instruction matcher is gone.

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
rynowak: I'm going to address your feedback on this class even though I'm not sure of it's future. It's easier to keep things clean this way.

rynowak: I've moved all of these out of the main algorithm and into static methods

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
rynowak: Sure

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
rynowak: I've removed this since it's not actually the optimal thing.

src/Microsoft.AspNetCore.Routing/Matchers/Candidate.cs
rynowak: I'm not of fan of using `Debug.Assert` where something trivial is going on. I think there's a really small risk of this regressing.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
rynowak: Agreed, I tend to use Debug.Assert liberally where there's complex conditional logic that I can't otherwise refactor. using it everywhere that's possible is just homework.

src/Microsoft.AspNetCore.Routing/Matchers/ParameterSegmentMatchProcessor.cs
rynowak: This is an interesting idea, and I don't think one we've considered before. What would it look like though?

I'm trying to make this lazy as possible, but I haven't pushed that down to **inside** the RVD yet.

src/Microsoft.AspNetCore.Routing/Matchers/LinearSearchJumpTable.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                return _exit;
+            }
+
+            for (var i = 0; i < _entries.Length; i++)
+            {
+                ref var entry = ref _entries[i];
+                if (segment.Length == entry.text.Length &&
+                    string.Compare(
+                    path,
```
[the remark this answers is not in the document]
rynowak: I've fixed this in #594 - I'll pick it up here once I merge and rebase that bad boy

src/Microsoft.AspNetCore.Routing/Matchers/MatcherBuilderEntry.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            comparison = Precedence.CompareTo(other.Precedence);
+            if (comparison != 0)
+            {
+                return comparison;
+            }
+
+            if (HttpMethod != null && other.HttpMethod == null)
+            {
+                return 0.CompareTo(1);
```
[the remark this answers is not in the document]
rynowak: lol sorry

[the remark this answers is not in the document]
rynowak: @pranavkm - is this better or worse than `patience = 7986`

test/Microsoft.AspNetCore.Routing.Tests/Matchers/InstructionMatcher.cs
rynowak: Rebased on top of `InstructionMatcher`'s deletion.

src/Microsoft.AspNetCore.Routing/Matchers/AsciiKeyedJumpTable.cs
rynowak: This is being cleaned up as part of #597. I'll remove that class from this PR.

src/Microsoft.AspNetCore.Routing/Matchers/RouteConstraintMatchProcessor.cs
rynowak: Sure.

rynowak:
Rebased on top of #594

rynowak:
Closing this PR since it's not really active any more. Most of the meaningful parts of this have been spun off into our changes. I'll send another PR in the near future for the `EndpointSelector` bits.

PR #592, opened 2018-07-12
Add support for httpmethods in swaggatherer

rynowak:
/cc @JamesNK @kichalla 

Split apart from #591 for your convenience


rynowak:
Thanks 😆 - @kichalla if you have any feedback about this I'll address in a follow up.

PR #594, opened 2018-07-12
Productize JumpTable

rynowak:
/cc @JamesNK @kichalla - split from #591 for your convenience


benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/JumpTableMultipleEntryBenchmark.cs
```diff
@@ -0,0 +1,134 @@
...
+        private string[] _strings;
+        private PathSegment[] _segments;
+        private int[] _samples;
+
+        private JumpTable _linearSearch;
+        private JumpTable _dictionary;
+
+        // All factors of 100 to support sampling
+        [Params(2, 5, 10, 25, 50, 100)]
+        public int Count;
```
rynowak: FYI the breakpoint for dictionary == linear search is 10. This is high because of the substring

I'm not including any of the more esoteric implementations in this PR because those will require really detailed tests and review. Everything here is pretty simple.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/JumpTableMultipleEntryBenchmark.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                index = segments[i].Length == 0 ? -1 :
+                    segments[i].Length != strings[i].Length ? 1 :
+                    string.Compare(
+                        strings[i],
+                        segments[i].Start,
+                        strings[i],
+                        0,
+                        strings[i].Length,
+                        StringComparison.OrdinalIgnoreCase);
```
rynowak: This is a little obtuse but it basically simulates the minimum amount of work that an implementation will do.

src/Microsoft.AspNetCore.Routing/Matchers/DictionaryJumpTable.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public override int GetDestination(string path, PathSegment segment)
+        {
+            if (segment.Length == 0)
+            {
+                return _exit;
+            }
+
+            var text = path.Substring(segment.Start, segment.Length);
```
rynowak: Yeah 😢  - this is a longstanding problem in routing. We need case-insensitive hashing of portions of strings.

@GrabYourPitchforks is going to help us with this, and I'd prefer to use whatever he's baking rather than some of the esoteric things I developed in #591

rynowak: Either way I expect this code as-is to be temporary, but at least it's working and simple and temporary.

src/Microsoft.AspNetCore.Routing/Matchers/LinearSearchJumpTable.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            if (segment.Length == 0)
+            {
+                return _exit;
+            }
+
+            var entries = _entries;
+            for (var i = 0; i < entries.Length; i++)
+            {
+                var text = entries[i].text;
```
rynowak: @JamesNK  - FYI we did find that this pattern (not using ref) is the fastest for this loop. #591 isn't consistent. Sometimes less esoteric is better.

src/Microsoft.AspNetCore.Routing/Matchers/LinearSearchJumpTable.cs
```diff
@@ -0,0 +1,48 @@
...
+                if (segment.Length == text.Length &&
+                    string.Compare(
```
rynowak: Will fix indent 😆

src/Microsoft.AspNetCore.Routing/Matchers/SingleEntryJumpTable.cs
```diff
@@ -0,0 +1,50 @@
...
+        public override int GetDestination(string path, PathSegment segment)
+        {
+            if (segment.Length == 0)
+            {
+                return _exit;
+            }
+
+            if (segment.Length == _text.Length &&
+                string.Compare(
+                path,
```
rynowak: Will fix indent 😆

src/Microsoft.AspNetCore.Routing/Matchers/LinearSearchJumpTable.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    internal class LinearSearchJumpTable : JumpTable
+    {
+        private readonly int _default;
+        private readonly int _exit;
+        private readonly (string text, int destination)[] _entries;
+
+        public LinearSearchJumpTable(int @default, int exit, (string text, int destination)[] entries)
```
rynowak: @JamesNK do you still think it's unclear that these all refer to state indices? what about names like `defaultDestination`, `exitDestination`, etc? Better enough to be longer?

src/Microsoft.AspNetCore.Routing/Matchers/DictionaryJumpTable.cs
JamesNK: What about a struct key that is a string+index+length and implements GetHashCode?

JamesNK: Ohh, is the issue is the case-insensitive hash from the segment?

rynowak: Yep.

rynowak: So there are things that are more esoteric we can do to avoid the substring. But help is on the way from ya boy, so I figured we should let them handle that.

src/Microsoft.AspNetCore.Routing/Matchers/LinearSearchJumpTable.cs
JamesNK: I like Destination suffix.

What convinced me was `int segment` in some places and `PathSegment segment` in others.

rynowak: Ok, will do then.

rynowak:
Updated

rynowak:
@JamesNK plz 😁

src/Microsoft.AspNetCore.Routing/Matchers/InvalidOperation.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    [Serializable]
+    internal class InvalidOperation : Exception
```
JamesNK: 1) Exceptions are always suppose to be suffixed with "Exception", and 2) there already exists an `InvalidOperationException` so I'm hella confused 😬 

I'm guessing VS helpfully generated this class from a typo?

src/Microsoft.AspNetCore.Routing/Matchers/JumpTableBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+        public void AddEntry(string text, int destination)
+        {
+            _entries.Add((text, destination));
+        }
+
+        public JumpTable Build()
+        {
+            if (DefaultDestination < 0)
+            {
+                throw new InvalidOperation($"{nameof(DefaultDestination)} is not set. This is a bug.");
```
JamesNK: "This is a bug" probably isn't the right language for productionized code

src/Microsoft.AspNetCore.Routing/Matchers/JumpTableBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            if (ExitDestination < 0)
+            {
+                throw new InvalidOperation($"{nameof(ExitDestination)} is not set. This is a bug.");
+            }
+
+            if (_entries.Count == 0)
```
JamesNK: Add a comment above this section of code that a jumptable implementation is picked on entry count to optimize performance

src/Microsoft.AspNetCore.Routing/Matchers/LinearSearchJumpTable.cs
```diff
@@ -0,0 +1,73 @@
...
+            builder.Append("$+: ");
+            builder.Append(_defaultDestination);
+            builder.Append(", ");
+
+            builder.Append("$0: ");
+            builder.Append(_defaultDestination);
+
+            builder.Append(" }");
+
+
```
JamesNK: nit: too liney

src/Microsoft.AspNetCore.Routing/Matchers/ZeroEntryJumpTable.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public ZeroEntryJumpTable(int defaultDestination, int exitDestination)
+        {
+            _defaultDestination = defaultDestination;
+            _exitDestination = exitDestination;
+        }
+
+        public unsafe override int GetDestination(string path, PathSegment segment)
+        {
+            return segment.Length == 0 ? _exitDestination : _defaultDestination;
```
JamesNK: Maybe add a comment here about why segment length is checked

src/Microsoft.AspNetCore.Routing/Matchers/JumpTableBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    internal class JumpTableBuilder
+    {
+        private readonly List<(string text, int destination)> _entries = new List<(string text, int destination)>();
+
+        // The destination state when none of the text entries match.
+        public int DefaultDestination { get; set; } = -1;
```
JamesNK: I wonder if `-1` should be placed in a const. There are some places in code below where you test if a destination is -1 and a const would help with readability compared to a magic number.

src/Microsoft.AspNetCore.Routing/Matchers/InvalidOperation.cs
rynowak: Yeah, Thanks VS!

src/Microsoft.AspNetCore.Routing/Matchers/JumpTableBuilder.cs
rynowak: What would you suggest? If this were to happen, I would want the user to report a bug. This is an internal framework issues, and not something that the user can take any action to fix.

src/Microsoft.AspNetCore.Routing/Matchers/ZeroEntryJumpTable.cs
rynowak: This is explained on `JumpTableBuilder.ExitDestination`. I don't really want the same comment to appear in 5 other places.

src/Microsoft.AspNetCore.Routing/Matchers/JumpTableBuilder.cs
rynowak: I'm fine with it.

PR #599, opened 2018-07-12
Registering change events in RouteValuesBasedEndpointFinder

kichalla:
Per our offline discussion

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
+++ b/src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Routing
@@ -12,9 +13,9 @@ namespace Microsoft.AspNetCore.Routing
     {
         private readonly EndpointDataSource[] _dataSources;
         private readonly object _lock;
-
-        private IChangeToken _changeToken;
         private IReadOnlyList<Endpoint> _endpoints;
+        private IChangeToken _consumerChangeToken;
+        private CancellationTokenSource _cts;
 
         internal CompositeEndpointDataSource(IEnumerable<EndpointDataSource> dataSources)
         {
@@ -23,6 +24,7 @@ namespace Microsoft.AspNetCore.Routing
                 throw new ArgumentNullException(nameof(dataSources));
             }
 
+            CreateChangeToken();
             _dataSources = dataSources.ToArray();
             _lock = new object();
         }
@@ -32,7 +34,7 @@ namespace Microsoft.AspNetCore.Routing
             get
             {
                 EnsureInitialized();
-                return _changeToken;
+                return _consumerChangeToken;
             }
         }
 
@@ -48,7 +50,7 @@ namespace Microsoft.AspNetCore.Routing
         // Defer initialization to avoid doing lots of reflection on startup.
         private void EnsureInitialized()
         {
-            if (_changeToken == null)
+            if (_endpoints == null)
             {
                 Initialize();
             }
@@ -60,11 +62,49 @@ namespace Microsoft.AspNetCore.Routing
         {
             lock (_lock)
             {
-                _changeToken = new CompositeChangeToken(_dataSources.Select(d => d.ChangeToken).ToArray());
+                if (_endpoints == null)
+                {
+                    _endpoints = _dataSources.SelectMany(d => d.Endpoints).ToArray();
+
+                    foreach (var dataSource in _dataSources)
+                    {
+                        Extensions.Primitives.ChangeToken.OnChange(
+                            () => dataSource.ChangeToken,
+                            () => HandleChange());
+                    }
+                }
+            }
+        }
+
+        private void HandleChange()
+        {
+            lock (_lock)
+            {
+                // Refresh the endpoints from datasource so that callbacks can get the latest endpoints
                 _endpoints = _dataSources.SelectMany(d => d.Endpoints).ToArray();
 
-                _changeToken.RegisterChangeCallback((state) => Initialize(), null);
+                // Prevent consumers from re-registering callback to inflight events as that can 
+                // cause a stackoverflow
+                // Example:
+                // 1. B registers A
+                // 2. A fires event causing B's callback to get called
+                // 3. B executes some code in its callback, but needs to re-register callback 
+                //    in the same callback
+                var oldTokenSource = _cts;
+                var oldToken = _consumerChangeToken;
+
+                CreateChangeToken();
+
+                // Raise consumer callbacks. Any new callback registration would happen on the new token
+                // created in earlier step.
+                oldTokenSource.Cancel();
             }
+        }
+
+        private void CreateChangeToken()
+        {
+            _cts = new CancellationTokenSource();
+            _consumerChangeToken = new CancellationChangeToken(_cts.Token);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
+++ b/src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
@@ -27,7 +27,13 @@ namespace Microsoft.AspNetCore.Routing
             _endpointDataSource = endpointDataSource;
             _objectPool = objectPool;
 
+            // Build initial matches
             BuildOutboundMatches();
+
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -60,14 +65,29 @@ public IEnumerable<Endpoint> FindEndpoints(RouteValuesBasedEndpointFinderContext
...
+        {
+            _changeCallback = _endpointDataSource.ChangeToken.RegisterChangeCallback(
+                (state) =>
+                {
+                    // rebuild the matches
+                    BuildOutboundMatches();
+
+                    // re-register the callback as the change token is one time use only and a new change token
+                    // is produced every time
+                    RegisterCallback();
```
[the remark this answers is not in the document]
rynowak: Yep, I suspect this will be fixed by my suggesiton above. If not lets sit together and figure this out.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -25,14 +28,20 @@ internal CompositeEndpointDataSource(IEnumerable<EndpointDataSource> dataSources
...
+            CreateConsumerChangeToken();
         }
 
         public override IChangeToken ChangeToken
         {
             get
             {
                 EnsureInitialized();
-                return _changeToken;
+                if (_consumerChangeToken.HasChanged)
```
kichalla: This change enables to not register a callback to an inflight event which resulted in stackoverflow

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -60,10 +63,20 @@ private void Initialize()
...
+                _consumerChangeToken = new CancellationChangeToken(_cts.Token);
+                _dataSourcesChangeToken = new CompositeChangeToken(_dataSources.Select(d => d.ChangeToken).ToArray());
                 _endpoints = _dataSources.SelectMany(d => d.Endpoints).ToArray();
 
-                _changeToken.RegisterChangeCallback((state) => Initialize(), null);
+                _dataSourcesChangeToken.RegisterChangeCallback(
+                    (state) =>
+                    {
+                        // raise consumer callbacks
+                        _cts.Cancel();
```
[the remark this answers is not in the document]
kichalla: Yes, you are right. I had to change the order of these lines + few other changes to make it work.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -4,17 +4,20 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Routing
 {
-    public class CompositeEndpointDataSource : EndpointDataSource
+    public class CompositeEndpointDataSource : EndpointDataSource, IDisposable
```
rynowak: What does disposing actually do here? Like, what's the impact if you don't dispose?

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -32,7 +36,7 @@ public override IChangeToken ChangeToken
             get
             {
                 EnsureInitialized();
-                return _changeToken;
+                return _consumerChangeToken;
```
rynowak: should this take the lock?

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -60,11 +67,53 @@ private void Initialize()
...
+                        oldTokenSource.Cancel();
+
+                        oldTokenSource.Dispose();
+                    },
+                    state: null);
+
+                if (oldChangeCallback != null && ReferenceEquals(oldChangeCallback, _dataSourcesChangeCallback))
+                {
+                    oldChangeCallback.Dispose();
+                }
```
rynowak: I don't think this is needed. The Dispose here just tells you to stop listening. We don't need to stop listening because the old callback has already been fired. Did you copy this from somewhere else?

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -45,10 +49,13 @@ public override IReadOnlyList<Endpoint> Endpoints
             }
```
rynowak: Should this take the lock?

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -14,11 +14,12 @@
 
 namespace Microsoft.AspNetCore.Routing
 {
-    internal class RouteValuesBasedEndpointFinder : IEndpointFinder<RouteValuesBasedEndpointFinderContext>
+    internal class RouteValuesBasedEndpointFinder : IEndpointFinder<RouteValuesBasedEndpointFinderContext>, IDisposable
```
rynowak: No. Stop, we should not do this.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -60,11 +62,43 @@ private void Initialize()
...
+                        () => dataSource.ChangeToken,
+                        () => HandleChange());
+                }
             }
         }
+
+        private void HandleChange()
+        {
+            // Refresh the endpoints from datasource so that callbacks can get the latest endpoints
+            Initialize();
```
rynowak: This whole method should also take the lock.

rynowak: Also, add a comment explaining that you are purposely updating the endpoints **before** triggering the public change token. This is important because it avoids the need for a lock to access the endpoints.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -60,11 +62,43 @@ private void Initialize()
...
+
+            // Prevent consumers from re-registering callback to inflight events as that can 
+            // cause a stackoverflow
+            // Example:
+            // 1. B registers A
+            // 2. A fires event causing B's callback to ge called
```
rynowak: get 😆

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
rynowak: I would suggest that you don't call `Initialize` here - that will create redundant change token registrations. Just copy-paste line 65

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -56,14 +62,26 @@ public IEnumerable<Endpoint> FindEndpoints(RouteValuesBasedEndpointFinderContext
...
+        }
+
         private void BuildOutboundMatches()
         {
-            var (allOutboundMatches, namedOutboundMatches) = GetOutboundMatches();
-            _namedMatches = GetNamedMatches(namedOutboundMatches);
-            _allMatchesLinkGenerationTree = new LinkGenerationDecisionTree(allOutboundMatches.ToArray());
+            var (allMatches, namedMatches) = GetOutboundMatches();
+            _namedMatches = GetNamedMatches(namedMatches);
+            _allMatchesLinkGenerationTree = new LinkGenerationDecisionTree(allMatches.ToArray());
```
rynowak: Add a comment here that this is OK because no consumer will read both of these fields. Otherwise you need a lock.

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -27,7 +27,13 @@ internal class RouteValuesBasedEndpointFinder : IEndpointFinder<RouteValuesBased
             _endpointDataSource = endpointDataSource;
             _objectPool = objectPool;
 
+            // Build initial matches
             BuildOutboundMatches();
+
+            // Register for changes in endpoints
+            Extensions.Primitives.ChangeToken.OnChange(
+                        () => _endpointDataSource.ChangeToken,
+                        () => HandleChange());
```
rynowak: nit: too much indent

Issue #588, opened 2018-07-09
UseEndpoint without UseDispatcher


JamesNK:
Today `UseDispatcher` without `UseEndpoint` results in an NRE in `EndpointMiddleware.Invoke`. Not ideal.

What should happen? Throw an error message saying that the user must call `UseDispatcher` on start up or exit the middleware without error?

I think a friendly error message is probably best.

JamesNK:
Look into registering a marker and throwing, similar to what MVC does.

JamesNK:
Added runtime check. Investigate adding startup check as well.

PR #600, opened 2018-07-12
Error message on no dispatcher middleware in endpoint middleware

JamesNK:


src/Microsoft.AspNetCore.Routing/EndpointMiddleware.cs
```diff
@@ -32,6 +32,11 @@ public EndpointMiddleware(ILogger<EndpointMiddleware> logger, RequestDelegate ne
         public async Task Invoke(HttpContext httpContext)
         {
             var feature = httpContext.Features.Get<IEndpointFeature>();
+            if (feature == null)
```
kichalla: We wanted to do at startup time? https://github.com/aspnet/Routing/issues/588#issuecomment-403653357 
You seem to have closed that issue for a different change.

JamesNK: I didn't see a way to do it at startup. The check in UseMvc that I was thinking about is checking DI registration.

kichalla: Based on @rynowak 's idea, `IApplicationBuilder` exposes `Properties` where we could register a flag if UseDispatcher is called and can verify that property in UseEndpoint.

PR #602, opened 2018-07-12
Fix TemplateBinder to consider null and empty string values the same

kichalla:


src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -375,7 +375,14 @@ public static bool RoutePartsEqual(object a, object b)
             var sa = a as string;
             var sb = b as string;
 
-            if (sa != null && sb != null)
+            // In case of strings, consider empty and null the same.
```
kichalla: This change is part of the 4 failing Dispatcher tests in MVC. After this change, I was able to bring down functional test failures from 4 to 2. I want to discuss about the remaining 2 functional tests in person as they ideally do not use the same route setup w.r.t dispatcher

rynowak: Hmm this is interesting. I'm surprised to see that we have a bug here. 

Can you provide a little explanation of what exactly happens?

rynowak: Oh nevermind I found the explanation elsewhere 😆

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -375,7 +375,14 @@ public static bool RoutePartsEqual(object a, object b)
...
+            // Since null cannot tell us the type, consider it to be a string if the other value is a string.
+            if ((sa != null && string.IsNullOrEmpty(sa) && sb == null) ||
+                (sb != null && string.IsNullOrEmpty(sb) && sa == null))
```
rynowak: I think this would be a little clearer if you wrote this as `sa == string.Empty...` and so on. That's a simpler form of `sa != null && string.IsNullOrEmpty(sa)` and it's closer to what you actually want.

JamesNK: `sa != null && string.IsNullOrEmpty(sa)` -> `string.IsNullOrEmpty(sa)`

cause IsNullOrEmpty is already checking for null.

rynowak: Right. and we specifically need to know if it's **exactly** the empty string.

pranavkm: Might be better to write it like so:

```C#
        public static bool RoutePartsEqual(object a, object b)
        {
            if (a is string sa && b is string sb)
            {
                return (string.IsNullOrEmpty(sa) && string.IsNullOrEmpty(sb)) ||
                    string.Equals(sa, sb, StringComparison.OrdinalIgnoreCase);
            }
            else
            {
                if (a != null && b != null)
                {
                    // Explicitly call .Equals() in case it is overridden in the type
                    return a.Equals(b);
                }
                else
                {
                    // At least one of them is null. Return true if they both are
                    return a == b;
                }
            }
        }
```

rynowak: This would have the same bug. `is` will return false when one of them is null.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -375,7 +375,13 @@ public static bool RoutePartsEqual(object a, object b)
             var sa = a as string;
             var sb = b as string;
 
-            if (sa != null && sb != null)
+            // In case of strings, consider empty and null the same.
+            // Since null cannot tell us the type, consider it to be a string if the other value is a string.
+            if ((sa == string.Empty && sb == null) || (sb == string.Empty && sa == null))
```
JamesNK: Does that change need to be migrated to any of the work @rynowak is doing with patterns and/or matches?

rynowak: yes, I'll need to port this 👍 thanks for thinking of it

Issue #608, opened 2018-07-13
Polish startup experience for Routing vNext


rynowak:
We need to do here:
- [x] Update names for middleware and related concepts (`UseGlobalRouting`, `UseEndpoint`)
- [x] Update experience for MVC - "how can MVC users get Routing vNext without making code changes"
- [ ] Template updates (if necessary)

rynowak:
What's the template update? just the compat switch?

JamesNK:
Yeah. Will the 2.2 switch will be done as part of the regular release process and we don't need to worry about?

rynowak:
@JamesNK close? Is this still tracking anything?

PR #610, opened 2018-07-14
Check dispatcher services registered

JamesNK:


JamesNK:
Contributes to https://github.com/aspnet/Routing/issues/608

src/Microsoft.AspNetCore.Routing/Internal/DispatcherMarkerService.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Routing.Internal
+{
+    /// <summary>
+    /// A marker class used to determine if all the dispatcher services were added
+    /// to the <see cref="IServiceCollection"/> before dispatcher is configured.
+    /// </summary>
+    public class DispatcherMarkerService
```
rynowak: Does this have to be pubternal? routing is like, 1 assembly?

src/Microsoft.AspNetCore.Routing/Builder/DispatcherApplicationBuilderExtensions.cs
```diff
@@ -2,19 +2,39 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Internal;
+using Microsoft.Extensions.DependencyInjection;
+using System;
```
rynowak: sort

test/Microsoft.AspNetCore.Routing.Tests/DispatcherApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            // Arrange
+            var app = new ApplicationBuilder(Mock.Of<IServiceProvider>());
+
+            // Act
+            var ex = Assert.Throws<InvalidOperationException>(() => app.UseEndpoint());
+
+            // Assert
+            Assert.Equal(
+                "Unable to find the required services. " +
```
rynowak: suggestion - `Unable to find the required Dispatcher services.`

src/Microsoft.AspNetCore.Routing/Internal/DispatcherMarkerService.cs
JamesNK: I don't think so but I followed the pattern of the existing check for routing and RouteMarkerService was pub internal 🤷‍♂️

rynowak: Yeah, that was written years ago, I'd suggest just making it internal. 

If you that makes you sad that it's inconsistent feel free to change `RouteMarkerService`.

src/Microsoft.AspNetCore.Routing/Builder/DispatcherApplicationBuilderExtensions.cs
JamesNK: Is it not already sorted?

test/Microsoft.AspNetCore.Routing.Tests/DispatcherApplicationBuilderExtensions.cs
JamesNK: Hmm, the rest of the instructions say what you need to do, and it is a consistent message with when routing and MVC is missing.

src/Microsoft.AspNetCore.Routing/Internal/DispatcherMarkerService.cs
JamesNK: Changed both

src/Microsoft.AspNetCore.Routing/Builder/DispatcherApplicationBuilderExtensions.cs
rynowak: System should be first 😆  https://github.com/aspnet/Home/wiki/Engineering-guidelines#coding-style-guidelines--general

JamesNK: oic

rynowak: I could make a job about the sort order being upside-down where you are. If it would put you at ease...

PR #612, opened 2018-07-14
Change benchmarks

rynowak:
/cc @JamesNK @kichalla - separated from #591 to make it easier on all of us :)

Add benchmarks that include some HTTP method matching.

Clean up names and name like-kinded benchmarks alphabetically.

Matcher*Benchmark -> E2E including HTTP method selection
MatcherSelectCandidates*Benchmark -> Focused on just URL path processing


benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/Candidate.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    // This is not yet fleshed out - consider this part of the
+    // work-in-progress definition of CandidateSet.
+    internal readonly struct Candidate
```
rynowak: I needed to define a little more infrastructure to merge the benchmarks

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherAzureBenchmark.cs
```diff
@@ -1,20 +1,18 @@
...
 
-using System;
 using System.Threading.Tasks;
 using BenchmarkDotNet.Attributes;
 
 namespace Microsoft.AspNetCore.Routing.Matchers
 {
     // Generated from https://github.com/Azure/azure-rest-api-specs
-    public partial class AzureMatcherBenchmark : MatcherBenchmarkBase
+    public partial class MatcherAzureBenchmark : MatcherBenchmarkBase
```
rynowak: Renaming and changing these so they can sort alphabetically. It makes the console UI for this much better to use.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherAzureBenchmark.cs
```diff
@@ -1,20 +1,18 @@
...
     {
         private const int SampleCount = 100;
 
         private BarebonesMatcher _baseline;
         private Matcher _dfa;
-        private Matcher _route;
```
rynowak: I'm taking the legacy route implementation out of these big tests, because it's just so slow. It makes the benchmarks really painful to work with and isn't adding much value because it's like 100x slower than the baseline.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherBenchmarkBase.cs
```diff
@@ -40,7 +40,7 @@ internal static MatcherEndpoint CreateEndpoint(string template, string httpMetho
                 new RouteValueDictionary(),
                 new List<MatchProcessorReference>(),
                 0,
-                EndpointMetadataCollection.Empty,
+                new EndpointMetadataCollection(metadata),
```
rynowak: Oopsies

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesAzureBenchmark.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    // Generated from https://github.com/Azure/azure-rest-api-specs
+    public partial class MatcherSelectCandidatesAzureBenchmark : MatcherBenchmarkBase
+    {
+        private const int SampleCount = 100;
+
+        private BarebonesMatcher _baseline;
+        private DfaMatcher _dfa;
```
rynowak: This is the first of a set of 4 small lower level benchmarks for the DFA algorithm. I'm just testing the DFA here because the other ones don't have a similar layering.

src/Microsoft.AspNetCore.Routing/Matchers/FastPathTokenizer.cs
```diff
@@ -9,6 +9,16 @@ namespace Microsoft.AspNetCore.Routing.Matchers
...
     {
+        // The default limit for the number of segments we tokenize.
+        //
+        // Historically the limit on the number of segments routing supports is 28.
+        // RoutePrecedence computes precedence based on a decimal, which supports 28
+        // or 29 digits.
+        //
+        // So setting this limit to 32 should work pretty well. We also expect the tokenizer
+        // to be used with stackalloc, so we want a small number.
+        public const int DefaultSegmentCount = 32;
```
rynowak: This is from PR feedback I already addressed in the other PR 😆

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesAzureBenchmark.cs
```diff
@@ -0,0 +1,72 @@
...
+            _dfa = (DfaMatcher)SetupMatcher(new DfaMatcherBuilder());
+        }
+
+        [Benchmark(Baseline = true, OperationsPerInvoke = SampleCount)]
+        public unsafe void Baseline()
+        {
+            for (var i = 0; i < SampleCount; i++)
+            {
+                var sample = _samples[i];
+                var httpContext = _requests[sample];
```
JamesNK: Can `_requests` and `_endpoints` be protected instead of internal? Also IMO rename to `Requests` and `Endpoints`. I reserve underscores for only private fields.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesSmallEntryCountBenchmark.cs
```diff
@@ -72,43 +68,31 @@ private Matcher SetupMatcher(MatcherBuilder builder)
         }
 
         [Benchmark(Baseline = true)]
-        public async Task Baseline()
+        public unsafe void Baseline()
```
JamesNK: What is unsafe here?

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesSmallEntryCountBenchmark.cs
```diff
@@ -72,43 +68,31 @@ private Matcher SetupMatcher(MatcherBuilder builder)
...
         [Benchmark]
-        public async Task LegacyTreeRouter()
+        public unsafe void Dfa()
         {
-            var feature = _feature;
+            var httpContext = _requests[0];
+            var path = httpContext.Request.Path.Value;
+            var buffer = stackalloc PathSegment[FastPathTokenizer.DefaultSegmentCount];
+            var count = FastPathTokenizer.Tokenize(path, buffer, FastPathTokenizer.DefaultSegmentCount);
+            var segments = new ReadOnlySpan<PathSegment>((void*)buffer, count);
```
JamesNK: Is unsafe code required here? I thought a `stackalloc`ed array implicitly converted to a Span/ReadOnlySpan.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/BarebonesMatcher.cs
```diff
@@ -47,27 +47,32 @@ public sealed class InnerMatcher : Matcher
         {
             private readonly string[] _segments;
             public readonly MatcherEndpoint _endpoint;
```
JamesNK: Change to private or rename to Endpoint

test/Microsoft.AspNetCore.Routing.Tests/Matchers/DfaMatcher.cs
```diff
@@ -65,7 +63,15 @@ public unsafe override Task MatchAsync(HttpContext httpContext, IEndpointFeature
                     }
                 }
 
-                matches.Add((candidates[i].Endpoint, values));
+                // This is some super super temporary code so we can pass the benchmarks
+                // that do HTTP method matching.
+                var httpMethodConstraint = candidate.Endpoint.Metadata.GetMetadata<HttpMethodEndpointConstraint>();
+                if (httpMethodConstraint != null && !MatchHttpMethod(httpContext.Request.Method, httpMethodConstraint))
+                {
+                    goto notmatch;
```
JamesNK: Is this goto really required? Isn't it just skipping one line immediately below it?

test/Microsoft.AspNetCore.Routing.Tests/Matchers/DfaMatcher.cs
```diff
@@ -76,17 +82,39 @@ public unsafe override Task MatchAsync(HttpContext httpContext, IEndpointFeature
...
+
+            for (var i = 0; i < segments.Length; i++)
+            {
+                current = states[current].Transitions.GetDestination(path, segments[i]);
+            }
+
+            return states[current].Matches;
+        }
+
+        public struct State
```
JamesNK: copy pasta the readonly `State` from other PR

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesAzureBenchmark.cs
rynowak: The can be `private protected`. The reason they weren't is that benchmark classes have to be public, and many of these types are internal.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesSmallEntryCountBenchmark.cs
rynowak: Doesn't stackalloc always require unsafe? I'll see what happens if I remove

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesSmallEntryCountBenchmark.cs
rynowak: You might be right. I may be able to get rid of unsafe in a bunch of places.

test/Microsoft.AspNetCore.Routing.Tests/Matchers/DfaMatcher.cs
rynowak: Dang. I try to hide one goto a week in the product and I failed 😢

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesAzureBenchmark.cs
rynowak: BTW it take a really really long time to rename these due to the generated code 🤣

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesSmallEntryCountBenchmark.cs
rynowak: It turns out you're right btw. You don't need unsafe for stackalloc - you need unsafe to dereference a pointer.

rynowak:
@JamesNK updated

I addressed the feedback and did some general cleanup around unsafe. Using `Span<>` instead of a pointer in `FastPathTokenizer` had no negative perf impact so I made that change, and resolved all of the fallout from it. 

I've deliberately left unsafe and pointers in the baselines for the path tokenizer.

PR #614, opened 2018-07-14
Add application startup check that dispatcher is added to request pipeline

JamesNK:
Fixes https://github.com/aspnet/Routing/issues/588

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Builder/DispatcherApplicationBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Routing/Builder/DispatcherApplicationBuilderExtensions.cs
@@ -25,9 +25,9 @@ namespace Microsoft.AspNetCore.Builder
         {
             VerifyDispatcherIsRegistered(builder);
 
-            if (!builder.Properties.TryGetValue(DispatcherRegisteredKey, out var isRegistered))
+            if (!builder.Properties.TryGetValue(DispatcherRegisteredKey, out _))
             {
-                var message = $"{nameof(DispatcherMiddleware)} must be added to the request execution pipeline to use {nameof(EndpointMiddleware)}. " +
+                var message = $"{nameof(DispatcherMiddleware)} must be added to the request execution pipeline before {nameof(EndpointMiddleware)}. " +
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Routing/Builder/DispatcherApplicationBuilderExtensions.cs
```diff
@@ -10,17 +10,29 @@ namespace Microsoft.AspNetCore.Builder
...
+            builder.Properties[DispatcherRegisteredKey] = true;
+
             return builder.UseMiddleware<DispatcherMiddleware>();
         }
 
         public static IApplicationBuilder UseEndpoint(this IApplicationBuilder builder)
         {
             VerifyDispatcherIsRegistered(builder);
 
+            if (!builder.Properties.TryGetValue(DispatcherRegisteredKey, out var isRegistered))
```
rynowak: nit: could also use `_` here since we don't need the value

src/Microsoft.AspNetCore.Routing/Builder/DispatcherApplicationBuilderExtensions.cs
```diff
@@ -10,17 +10,29 @@ namespace Microsoft.AspNetCore.Builder
...
+            {
+                var message = $"{nameof(DispatcherMiddleware)} must be added to the request execution pipeline to use {nameof(EndpointMiddleware)}. " +
+                    $"Please add {nameof(DispatcherMiddleware)} by calling '{nameof(IApplicationBuilder)}.{nameof(UseDispatcher)}' inside the call to 'Configure(...)' in the application startup code.";
```
rynowak: Should we also mention that this needs to come **prior to** `UseEndpoint`?

test/Microsoft.AspNetCore.Routing.Tests/DispatcherApplicationBuilderExtensionsTest.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace Microsoft.AspNetCore.Builder
 {
-    public class DispatcherApplicationBuilderExtensions
```
rynowak: lel

PR #618, opened 2018-07-16
Allow to pass in HttpContext when generating links

kichalla:


src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -2,18 +2,21 @@
...
+using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.Routing
 {
-    public interface ILinkGenerator
+    public abstract class LinkGenerator
     {
-        bool TryGetLink(
+        public abstract bool TryGetLink(
+            HttpContext httpContext,
```
JamesNK: Since this is an optional parameter should this come after other parameter?

kichalla: I am not sure if that would still be clear(if its optional or not) for an end user. Documentation of this API should help.

kichalla: Let's discuss this in during API review though.

kichalla:
https://github.com/aspnet/Routing/commit/aed2e244831f4aeb3ef5b3b5c33f4a053f70a9a2

PR #620, opened 2018-07-17
Hoist path out of loop in BarebonesMatcher

rynowak:
This is an oopsie from a previous change.


rynowak:
FYI only - this is a mistake a made in a previous PR that hurts the perf of some benchmarks in a trivial way 🤣

PR #637, opened 2018-07-19
Added support for configuring route options with LinkGenerator

### LINKED ISSUE #547: Link generation: Expose options to configure link generation at the call site and globally
- fully qualified URLs
- Casing
- trailing slashes

kichalla:
[Fixes #547] Link generation: Expose options to configure link generation at the call site and globally

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
@@ -1,24 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
-using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.Routing
 {
     public abstract class LinkGenerator
     {
-        public abstract bool TryGetLink(
-            HttpContext httpContext,
-            IEnumerable<Endpoint> endpoints,
-            RouteValueDictionary explicitValues,
-            RouteValueDictionary ambientValues,
-            out string link);
+        public abstract bool TryGetLink(LinkGeneratorContext context, out string link);
 
-        public abstract string GetLink(
-            HttpContext httpContext,
-            IEnumerable<Endpoint> endpoints,
-            RouteValueDictionary explicitValues,
-            RouteValueDictionary ambientValues);
+        public abstract string GetLink(LinkGeneratorContext context);
     }
 }
--- a/src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
+++ b/src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
@@ -11,32 +11,38 @@ using Microsoft.AspNetCore.Routing.Matchers;
 using Microsoft.AspNetCore.Routing.Template;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.ObjectPool;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Routing
 {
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -1,24 +1,13 @@
...
 {
     public abstract class LinkGenerator
     {
-        public abstract bool TryGetLink(
-            HttpContext httpContext,
-            IEnumerable<Endpoint> endpoints,
-            RouteValueDictionary explicitValues,
-            RouteValueDictionary ambientValues,
-            out string link);
+        public abstract bool TryGetLink(LinkGeneratorContext context, out string link);
```
JamesNK: 👍 I'm pro a context object. More extensible, and there were getting too many parameters

src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -11,54 +11,60 @@
...
+        private readonly static char[] UrlQueryDelimiters = new char[] { '?', '#' };
         private readonly MatchProcessorFactory _matchProcessorFactory;
         private readonly ObjectPool<UriBuildingContext> _uriBuildingContextPool;
+        private readonly RouteOptions _options;
         private readonly ILogger<DefaultLinkGenerator> _logger;
 
         public DefaultLinkGenerator(
             MatchProcessorFactory matchProcessorFactory,
             ObjectPool<UriBuildingContext> uriBuildingContextPool,
+            IOptions<RouteOptions> options,
```
JamesNK: options -> routeOptions, just to be explicit

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -239,15 +259,259 @@ public void GetLink_Success_AmbientValues()
...
+            var link = linkGenerator.GetLink(
+                new LinkGeneratorContext
+                {
+                    Endpoints = new[] { endpoint },
+                    ExplicitValues = context.ExplicitValues,
+                    AmbientValues = context.AmbientValues
+                });
+
+            // Assert
+            Assert.Equal("/Home/Index?ShowStatus=True&INFO=DETAILED", link);
```
JamesNK: query string is not lowercase?

kichalla: It is lower cased only when LowercaseUrls is also true

JamesNK: Is that existing behavior?

Need to make sure that is documented on property

kichalla: Yeah, its an existing behavior (i know it feels weird) https://github.com/aspnet/Routing/blob/release/2.2/src/Microsoft.AspNetCore.Routing/RouteOptions.cs#L20

JamesNK: Ok! Copy over those properties xml docs

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
rynowak: I'm ok with making some tenative decisions here if we think it's an improvement. We still need to do design work on this area in preview1 - so I would suggest not making any expensive changes until we've had a proper design review.

kichalla:
https://github.com/aspnet/Routing/commit/5f1631ab46e415cadf3a12ceef6c6164a06ac275

PR #639, opened 2018-07-19
[NO MERGE] Add support for CORS preflight requests to HTTP method constraint

JamesNK:
Is this what you meant by making CORS more first class in routing @rynowak?

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/EndpointConstraints/HttpMethodEndpointConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/EndpointConstraints/HttpMethodEndpointConstraint.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Extensions.Primitives;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
@@ -9,6 +10,10 @@ namespace Microsoft.AspNetCore.Routing.EndpointConstraints
 {
     public class HttpMethodEndpointConstraint : IEndpointConstraint
     {
+        private readonly string OriginHeader = "Origin";
+        private readonly string AccessControlRequestMethod = "Access-Control-Request-Method";
+        private readonly string PreflightHttpMethod = "OPTIONS";
+
         public static readonly int HttpMethodConstraintOrder = 100;
 
         private readonly IReadOnlyList<string> _httpMethods;
@@ -55,6 +60,28 @@ namespace Microsoft.AspNetCore.Routing.EndpointConstraints
             var request = context.HttpContext.Request;
             var method = request.Method;
 
+            if (HttpMethodSupported(method))
+            {
+                return true;
+            }
+
+            // Check if request is a CORS OPTIONS request
+            if (string.Equals(request.Method, PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
+                request.Headers.ContainsKey(OriginHeader) &&
+                request.Headers.TryGetValue(AccessControlRequestMethod, out var accessControlRequestMethod) &&
+                !StringValues.IsNullOrEmpty(accessControlRequestMethod))
+            {
+                if (HttpMethodSupported(accessControlRequestMethod))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private bool HttpMethodSupported(string method)
+        {
             for (var i = 0; i < _httpMethods.Count; i++)
             {
                 var supportedMethod = _httpMethods[i];
--- a/src/Microsoft.AspNetCore.Routing/EndpointConstraints/HttpMethodEndpointConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/EndpointConstraints/HttpMethodEndpointConstraint.cs
@@ -19,7 +19,11 @@ namespace Microsoft.AspNetCore.Routing.EndpointConstraints
         private readonly IReadOnlyList<string> _httpMethods;
 
         // Empty collection means any method will be accepted.
-        public HttpMethodEndpointConstraint(IEnumerable<string> httpMethods)
+        public HttpMethodEndpointConstraint(IEnumerable<string> httpMethods) : this(httpMethods, false)
+        {
+        }
+
+        public HttpMethodEndpointConstraint(IEnumerable<string> httpMethods, bool acceptCorsPreflight)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Routing/EndpointConstraints/HttpMethodEndpointConstraint.cs
```diff
@@ -55,6 +60,28 @@ public virtual bool Accept(EndpointConstraintContext context)
             var request = context.HttpContext.Request;
             var method = request.Method;
 
+            if (HttpMethodSupported(method))
+            {
+                return true;
+            }
+
+            // Check if request is a CORS OPTIONS request
+            if (string.Equals(request.Method, PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
```
rynowak: Yes, this is what I had in mind, but it also needs to check whether or not the endpoint supports CORS.

JamesNK: Hmm, that's a challenge because everything CORS references routing, routing doesn't reference CORS.

What about some CORS metadata interfaces to routing abstractions, have CORS attributes implement interfaces, and MvcEndpointDataSource consume it?

rynowak: I don't think CORS references routing though. 

I'd suggest one modification - CORS interfaces in dispatcher, have MVC detect the attributes (like it does today) and plug in it's own class implementiong of the metadata

test/Microsoft.AspNetCore.Routing.Tests/Internal/HttpMethodEndpointConstraintTest.cs
```diff
@@ -27,7 +27,7 @@ public class HttpMethodEndpointConstraintTest
 
         [Theory]
         [MemberData(nameof(AcceptCaseInsensitiveData))]
-        public void HttpMethodEndpointConstraint_IgnoresPreflightRequests(IEnumerable<string> httpMethods, string accessControlMethod)
+        public void HttpMethodEndpointConstraint_AcceptsPreflightRequests(IEnumerable<string> httpMethods, string accessControlMethod)
```
rynowak: We're also going to need the opposite test. An action that doesn't handle CORS requests should not allow a CORS request through

src/Microsoft.AspNetCore.Routing/EndpointConstraints/HttpMethodEndpointConstraint.cs
JamesNK: I checked and it doesn't currently. That's why I thought a routing abstractions reference would be better than a routing reference.

> CORS interfaces in dispatcher

What do you mean by this? Routing package?

rynowak: > What do you mean by this? Routing package?

rynowak: OK I'm dumb https://github.com/aspnet/CORS/blob/master/src/Microsoft.AspNetCore.Cors/Infrastructure/IEnableCorsAttribute.cs

naming isn't great, but CORS already defines something suitable for our purposes. I think it's fine that `Microsoft.AspNetCore.Routing` would reference CORS.

JamesNK: For perf the constraint shouldn't be checking the endpoint metadata inside the constraint when it is evaluated, the `MvcEndpointDataSource` should figure that out at startup and set some config on the constraint to tell it that CORS is enabled.

The issue with figuring it out in the endpoint data source is the duplication of a lot of this code which involves looking at attributes to get `IEnableCorsAttribute`: 
https://github.com/aspnet/Mvc/blob/release/2.2/src/Microsoft.AspNetCore.Mvc.Cors/Internal/CorsApplicationModelProvider.cs

rynowak: Do we need to duplicate the code? what if we inspected the results on `ActionDescriptor.Filters`?

JamesNK:
🆙 📅

src/Microsoft.AspNetCore.Routing/EndpointConstraints/HttpMethodEndpointConstraint.cs
JamesNK: The filters that `CorsApplicationModelProvider.cs` adds are located in `Microsoft.AspNetCore.Mvc.Cors`, plus they are pub internal.

JamesNK:
I've sorted out the merge conflict but I'm not sure what changes are required to `HttpMethodEndpointSelectorPolicy` to support CORS.

Adding AllowCorsPreflightRequest to `IHttpMethodMetadata` is simple enough, but my eyes started going a little cross eyed when adding support to the policy. Currently a `Dictionary<string, int>` is used to figure out where to go which I'm not sure if will still be possible with CORS and an OPTIONS HTTP method is thrown into the mix.

I want to run it by you tomorrow @rynowak

rynowak:
If you like, I cake sort out that part

JamesNK:
If you can do it that would be great. I feel like I know a lot more about policies now but not enough to implement it without a lot of questions.

One thing that I thought odd that I didn't pick up when reviewing the HTTP method policy is it talks about order being important but then uses a dictionary to store values. Are you relying on the dictionary order or is the order important elsewhere?

JamesNK:
PR not necessary. `HttpMethodEndpointConstraint.cs` to be removed and replaced with policy. Leaving PR around for a couple of days for reference.

rynowak:
> One thing that I thought odd that I didn't pick up when reviewing the HTTP method policy is it talks about order being important but then uses a dictionary to store values. Are you relying on the dictionary order or is the order important elsewhere?

The dictionary is effectively used as a Linq groupby. Order is important *within* a set of candidates in the same group.

JamesNK:
https://github.com/aspnet/Mvc/pull/8114 is rebased as a place to test. Tests are failing now because the CORS action constraint is not executed.

You can set the CORS policy, or set the preflight flag on the HTTP method policy here: https://github.com/aspnet/Mvc/pull/8114/files#diff-c935feae6d0877d7f9ae9075ad9306caR346 (I am working to improve the nasty type name check with global routing metadata on app model)

rynowak:
Closing this since we merged the real version 🤣

Issue #625, opened 2018-07-17
Back-compat support for IRoutingFeature on IEndpointFeature


JamesNK:
`RouteDataRequestCultureProvider` is unsurprisingly tied to routing. Do we want to create a version for dispatching?

rynowak:
I think what we might want to do instead is to implement `IRoutingFeature` on the dispatcher feature (lazily). There's a **LOT** of code out there that uses `IRoutingFeature`, or `RouteData`, etc. We can do a pretty reasonable job implementing the route value part of this.

We might also need to do something to support data tokens

JamesNK:
Sounds reasonable

JamesNK:
`MiddlewareFilter_LocalizationMiddlewareRegistration_UsesRouteDataToFindCulture` is passing in MVC

PR #641, opened 2018-07-20
Implement IRoutingFeature on EndpointFeature for back-compat

JamesNK:
https://github.com/aspnet/Routing/issues/625

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
@@ -52,6 +52,20 @@ namespace Microsoft.AspNetCore.Routing
         }
 
         /// <summary>
+        /// Creates a new <see cref="RouteData"/> instance with the specified values.
+        /// </summary>
+        /// <param name="values">The <see cref="RouteValueDictionary"/> copied for values.</param>
+        public RouteData(RouteValueDictionary values)
+        {
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            _values = new RouteValueDictionary(values);
+        }
+
+        /// <summary>
         /// Gets the data tokens produced by routes on the current routing path.
         /// </summary>
         public RouteValueDictionary DataTokens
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
```diff
@@ -51,6 +51,20 @@ public RouteData(RouteData other)
...
+        /// </summary>
+        /// <param name="values">The <see cref="RouteValueDictionary"/> copied for values.</param>
+        public RouteData(RouteValueDictionary values)
+        {
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            _values = new RouteValueDictionary(values);
```
JamesNK: This copies like `RouteData` does. The downside is changes in one RVD won't be automatically reflected in the other.

src/Microsoft.AspNetCore.Routing/EndpointFeature.cs
```diff
@@ -6,12 +6,39 @@
...
-        public RouteValueDictionary Values { get; set; }
+        public RouteValueDictionary Values
+        {
+            get => _values;
+            set
+            {
+                _values = value;
+
+                // RouteData will be created next get with new Values
+                _routeData = null;
```
JamesNK: Best attempt to keep the two RVD in sync I could think of. Do they change during a request much?

src/Microsoft.AspNetCore.Routing.Abstractions/RouteData.cs
rynowak: This seems kinda broken no? What if someone changes the values during the request?

JamesNK: I don't know. Is changing them during a request a thing people do? Do we want changes in the back-compat dictionary effecting endpoints? What do you think should happen here?

rynowak: Yes, people definitely do things to the route value dictionary. I think it will catch some people out if there are two copies.

src/Microsoft.AspNetCore.Routing/EndpointFeature.cs
rynowak: Yeah, this seems fine now that it's not making a copy. It's not common to **set** the values, but it's common to modify the dictionary itself.

PR #645, opened 2018-07-21
Show list of endpoints in CompositeEndpointDataSource's DebuggerDispl…

### LINKED ISSUE #633: Show list of registered endpoints as debugger display string
While debugging through MVC's functional test failures, it helped me to look at how MVC's data source registered endpoints. I used following blob of code to look at them. Would be helpful if we can show this via `DebuggerDisplayString` mechanism

```c#
var sb = new StringBuilder();
            foreach (var endpoint in _endpointDataSource.Endpoints)
            {
                if (endpoint is MatcherEndpoint matcherEndpoint)
                {
                    var template = string.IsNullOrEmpty(matcherEndpoint.Template) ? "\"\"" : matcherEndpoint.Template;
                    var s = template + ", "
                        + "Required Values: new { " + string.Join(", ", matcherEndpoint.RequiredValues.Select(kvp => $"{kvp.Key} = \"{kvp.Value ?? "null"}\"")) + " }, "
                        + "Order:" + matcherEndpoint.Order;

                    sb.Append(s);

                    var endpointConstraintMetadata = matcherEndpoint.Metadata.GetOrderedMetadata<IEndpointConstraintMetadata>();
                    foreach (var constraint in endpointConstraintMetadata)
                    {
                        if (constraint is HttpMethodEndpointConstraint httpMethodEndpointConstraint)
                        {
                            sb.Append(", constriants: " + string.Join(',', httpMethodEndpointConstraint.HttpMethods)
                                + ", constraint order:" + httpMethodEndpointConstraint.Order);
                        }
                    }

                    sb.AppendLine();
                    sb.AppendLine();
                }
                else
                {
                    sb.AppendLine("Non-matcherendpoint. DisplayName:" + endpoint.DisplayName);
                }
            }
            return Content(sb.ToString());
```

kichalla:
…ayString

[Fixes #633] Show list of registered endpoints as debugger display string

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
+++ b/src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
@@ -3,12 +3,17 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
+using System.Text;
 using System.Threading;
+using Microsoft.AspNetCore.Routing.EndpointConstraints;
+using Microsoft.AspNetCore.Routing.Matchers;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Routing
 {
+    [DebuggerDisplay("{DebuggerDisplayString,nq}")]
     public class CompositeEndpointDataSource : EndpointDataSource
     {
         private readonly EndpointDataSource[] _dataSources;
@@ -105,6 +110,52 @@ namespace Microsoft.AspNetCore.Routing
         {
             _cts = new CancellationTokenSource();
             _consumerChangeToken = new CancellationChangeToken(_cts.Token);
+        }
+
+        private string DebuggerDisplayString
+        {
+            get
+            {
+                // Try using private variable '_endpoints' to avoid initialization
+                if (_endpoints == null)
+                {
+                    return "No endpoints";
+                }
+
+                var sb = new StringBuilder();
+                foreach (var endpoint in _endpoints)
+                {
+                    if (endpoint is MatcherEndpoint matcherEndpoint)
+                    {
+                        var template = matcherEndpoint.RoutePattern.RawText;
+                        template = string.IsNullOrEmpty(template) ? "\"\"" : template;
+                        sb.Append(template);
+                        var requiredValues = matcherEndpoint.RequiredValues.Select(kvp => $"{kvp.Key} = \"{kvp.Value ?? "null"}\"");
+                        sb.Append(", Required Values: new { ");
+                        sb.Append(string.Join(", ", requiredValues));
+                        sb.Append(" }");
+                        sb.Append(", Order:");
+                        sb.Append(matcherEndpoint.Order);
+
+                        var httpEndpointConstraints = matcherEndpoint.Metadata.GetOrderedMetadata<IEndpointConstraintMetadata>()
+                            .OfType<HttpMethodEndpointConstraint>();
+                        foreach (var constraint in httpEndpointConstraints)
+                        {
+                            sb.Append(", Http Methods: ");
+                            sb.Append(string.Join(", ", constraint.HttpMethods));
+                            sb.Append(", Constraint Order:");
+                            sb.Append(constraint.Order);
+                        }
+                        sb.AppendLine();
+                    }
+                    else
+                    {
+                        sb.Append("Non-MatcherEndpoint. DisplayName:");
+                        sb.AppendLine(endpoint.DisplayName);
+                    }
+                }
+                return sb.ToString();
+            }
         }
     }
 }


src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -106,5 +111,52 @@ private void CreateChangeToken()
             _cts = new CancellationTokenSource();
             _consumerChangeToken = new CancellationChangeToken(_cts.Token);
         }
+
+        private string DebuggerDisplayString => ToDebuggerDisplayString();
+
+        private string ToDebuggerDisplayString()
```
JamesNK: Rename to DebuggerToString

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -3,12 +3,17 @@
...
 using System.Linq;
+using System.Text;
 using System.Threading;
+using Microsoft.AspNetCore.Routing.EndpointConstraints;
+using Microsoft.AspNetCore.Routing.Matchers;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Routing
 {
+    [DebuggerDisplay("{DebuggerDisplayString,nq}")]
```
JamesNK: [DebuggerDisplay("{DebuggerToString()}")]

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -106,5 +111,52 @@ private void CreateChangeToken()
             _cts = new CancellationTokenSource();
             _consumerChangeToken = new CancellationChangeToken(_cts.Token);
         }
+
+        private string DebuggerDisplayString => ToDebuggerDisplayString();
```
JamesNK: remove

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -106,5 +111,52 @@ private void CreateChangeToken()
...
+
+        private string ToDebuggerDisplayString()
+        {
+            // Try using private variable '_endpoints' to avoid initialization
+            if (_endpoints == null)
+            {
+                return "No endpoints available yet to display.";
```
JamesNK: nit: I think "No endpoints" is fine

Edit: Isn't _endpoints is assigned in the ctor and will never be null? Check for empty instead?

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -106,5 +111,52 @@ private void CreateChangeToken()
...
+                            sb.Append(string.Join(", ", httpMethodEndpointConstraint.HttpMethods));
+                            sb.Append(", Constraint Order:");
+                            sb.Append(httpMethodEndpointConstraint.Order);
+                        }
+                    }
+                    sb.AppendLine();
+                }
+                else
+                {
+                    sb.Append("Non-matcherendpoint. DisplayName:");
```
JamesNK: nit: Non-MatcherEndpoint.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -106,5 +111,52 @@ private void CreateChangeToken()
...
+                    sb.Append(", Required Values: new { ");
+                    sb.Append(string.Join(", ", requiredValues));
+                    sb.Append(" }");
+                    sb.Append(", Order:");
+                    sb.Append(matcherEndpoint.Order);
+
+                    var endpointConstraintMetadata = matcherEndpoint.Metadata.GetOrderedMetadata<IEndpointConstraintMetadata>();
+                    foreach (var constraint in endpointConstraintMetadata)
+                    {
+                        if (constraint is HttpMethodEndpointConstraint httpMethodEndpointConstraint)
```
JamesNK: What about adding internal virtual DebuggerToString to the constraint and then doing a foreach of the constraints. By default it would be type name + order. Override in HttpMethodEndpointConstraint to print methods.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
kichalla: > Edit: Isn't _endpoints is assigned in the ctor and will never be null? Check for empty instead?

No, it isn't assigned in the constructor. Its populated lazily.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
kichalla: I suppose you thought the base was an abstract type?, its actually an interface (IEndpointConstraintMetadata)

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
JamesNK: Remove and just have DebuggerDisplayString

JamesNK (approved):

PR #646, opened 2018-07-22
Implement EndpointSelector and MatcherPolicy

rynowak:
**TODO**
- [x] unit test `DefaultEndpointSelector`
- [x] unit test `HttpMethodMatcherPolicy`
- [x] e2e test matcher + `HttpMethodMatcherPolicy`
- [x] unit test `DfaMatcherBuilder.CreateCandidates` + `Score`
- [x] unit test `CandidateSet`
- [x] unit test `DfaMatcherBuilder.BuildDfaTree` + policy nodes
- [x] unit test `EndpointMetadataComparer`
- [x] unit test `MatcherEndpointComparer`

This change makes the EndpointSelector API more concrete, and is the
beggining of removing EndpointConstraint by making it obsolete.

To that end, I'm introducing MatcherPolicy, which is a
feature-collection API for registering policies that interact with the
DfaMatcher. The two policies that we'll need to start are:
- ability to order endpoints
- ability to append 'policy' nodes to the graph

These two concepts together replace EndpointConstraint. Extending our
graph representation is a really efficient way to processing most common
scenarios.

---

In general this helps with common cases where 4 or so endpoints match
the URL, but with different HTTP methods supported on each. Today we
have to process route values and call into some 'policy' to make a
decision about which one is the winner. This change pushes this
knowledge down into the graph so that it's roughly as cheap as a
dictionary lookup, and can be done allocation-free.

The big savings here is ability to remove more candidates *before*
collecting route data.

---

Along with this change, I also built 'rejection' into the DFA node
model, you can see an example with the HTTP Method handling that I
implemented. I implemented a policy that can treat failure to resolve an
HTTP method as a 405 response by returning a failure endpoint. This is
at the heart of much of the feedback we've gotten in this area around
versioning and http method handling. We also have a version of this
feature in MVC for [Consumes].

## TLDR
image


benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesAzureBenchmark.cs
```diff
@@ -44,7 +44,7 @@ public void Baseline()
 
                 var candidates = _baseline.Matchers[sample].SelectCandidates(path, segments);
 
-                var endpoint = candidates.Candidates[0].Endpoint;
+                var endpoint = candidates[0].Endpoint;
```
rynowak: This is the first of a few changes that spread out all over teh codez. `CandidateSet` is now public since we did away with groups. 

So, a whole bunch of things now work with (immutable) `Candidate[]` where the used to work with `CandidateSet`, because candidate set is now mutable and public.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherBenchmarkBase.cs
```diff
@@ -40,6 +41,7 @@ private protected static MatcherEndpoint CreateEndpoint(string template, string
             if (httpMethod != null)
             {
                 metadata.Add(new HttpMethodEndpointConstraint(new string[] { httpMethod, }));
+                metadata.Add(new HttpMethodMetadata(new string[] { httpMethod, }));
```
rynowak: This isn't needed. I'll remove.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesAzureBenchmark.cs
rynowak: So everwhere you see `.Candidates` deleted from the code, that's the reason.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesSmallEntryCountBenchmark.cs
```diff
@@ -90,9 +90,9 @@ public void Dfa()
             Span<PathSegment> segments = stackalloc PathSegment[FastPathTokenizer.DefaultSegmentCount];
             var count = FastPathTokenizer.Tokenize(path, segments);
 
-            var candidates = _dfa.SelectCandidates(path, segments.Slice(0, count));
+            var candidates = _dfa.FindCandidates(httpContext, path, segments.Slice(0, count));
```
rynowak: I renamed this method because it's not super clear. Lots of things were using the term `Select` for different tasks.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/TrivialMatcher.cs
```diff
@@ -13,17 +13,13 @@ namespace Microsoft.AspNetCore.Routing.Matchers
...
 
         public TrivialMatcher(MatcherEndpoint endpoint)
         {
             _endpoint = endpoint;
 
-            _candidates = new CandidateSet(
-                new Candidate[] { new Candidate(endpoint), },
-
-                // Single candidate group that contains one entry.
-                CandidateSet.MakeGroups(new[] { 1 }));
```
rynowak: See, groups are gone.

src/Microsoft.AspNetCore.Routing/DependencyInjection/DispatcherServiceCollectionExtensions.cs
```diff
@@ -43,11 +42,13 @@ public static IServiceCollection AddDispatcher(this IServiceCollection services)
             services.TryAddSingleton<IEndpointFinder<string>, NameBasedEndpointFinder>();
             services.TryAddSingleton<IEndpointFinder<RouteValuesBasedEndpointFinderContext>, RouteValuesBasedEndpointFinder>();
             services.TryAddSingleton<LinkGenerator, DefaultLinkGenerator>();
+
             //
             // Endpoint Selection
             //
-            services.TryAddSingleton<EndpointSelector>();
+            services.TryAddSingleton<EndpointSelector, EndpointConstraintEndpointSelector>();
```
rynowak: So **right now** we're using an implementation of `EndpointSelector` that supports `EndpointConstraint`. My hope is that we'd take this code to MVC as a back-compat feature, and delete `EndpointConstraint` from routing. 

As outlined in this PR we have options for policies that are an order of magnitude more performant than `EndpointConstraint`/`IActionConstraint`, and I'm aware of 4-5 actual implementations of `IActionConstraint` in the wild.

I had a hard think about what I want to optimize for scenario-wise. I want to make sure that our main path experiences like HTTP methods, consumes and versioning are really efficient. I'm willing to trade complexity for that. 

I think something that will help a lot would be a way to associate an `IRouteConstraint` with an endpoint/action in MVC. I think there's a lot of value in 'out of line' constraints that don't map to a route value for MVC, and we've never had that feature.

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintEndpointSelector.cs
```diff
@@ -1,28 +1,25 @@
...
-using Microsoft.Extensions.Options;
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
-using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing.Matchers;
+using Microsoft.Extensions.Logging;
```
rynowak: sorted 😆

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintEndpointSelector.cs
```diff
@@ -32,37 +29,50 @@ internal class EndpointSelector
             _endpointConstraintCache = endpointConstraintCache;
         }
 
-        public Endpoint SelectBestCandidate(HttpContext context, IReadOnlyList<Endpoint> candidates)
+        public override Task SelectAsync(
+            HttpContext httpContext,
+            IEndpointFeature feature,
+            CandidateSet candidates)
```
rynowak: A quick work about this API definition. I passed in the `IEndpointFeature` for perf. Accessing the feature collection is somewhat expensive in the scale we're working in.

Also, this returns task. You're intended to modify the feature collection. The reason why this is important is that you need to set a few properties on the feature. An API where you just return an endpoint wouldn't be sufficient.

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintEndpointSelector.cs
```diff
@@ -72,31 +82,60 @@ public Endpoint SelectBestCandidate(HttpContext context, IReadOnlyList<Endpoint>
...
+        private IReadOnlyList<EndpointSelectorCandidate> EvaluateEndpointConstraints(
             HttpContext context,
-            IReadOnlyList<Endpoint> endpoints)
+            CandidateSet candidateSet)
         {
             var candidates = new List<EndpointSelectorCandidate>();
 
             // Perf: Avoid allocations
-            for (var i = 0; i < endpoints.Count; i++)
+            for (var i = 0; i < candidateSet.Count; i++)
```
rynowak: Let's all sit together and remark how far we've come from a world where it was considered **worth it** to point out that we were using a `for` loop over `foreach` on purpose. 

This code was written by me in MVC like a zillion years ago.

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintEndpointSelector.cs
```diff
@@ -72,31 +82,60 @@ public Endpoint SelectBestCandidate(HttpContext context, IReadOnlyList<Endpoint>
...
             if (matches != null)
             {
-                results = new List<Endpoint>(matches.Count);
-                // Perf: Avoid allocations
-                for (var i = 0; i < matches.Count; i++)
+                results = new List<EndpointSelectorCandidate>(matches.Count);
+
+                // We need to disambiguate based on 'score' - take the first value of 'score'
+                // and then we only copy matches while they have the same score. This accounts
+                // for a difference in behavior between new routing and old.
```
rynowak: I replaced groups with a somewhat friendlier concept. see the comments on `Candidate.Score`.

src/Microsoft.AspNetCore.Routing/EndpointConstraints/HttpMethodEndpointConstraint.cs
```diff
@@ -40,6 +41,8 @@ public HttpMethodEndpointConstraint(IEnumerable<string> httpMethods)
 
         public int Order => HttpMethodConstraintOrder;
 
+        IReadOnlyList<string> IHttpMethodMetadata.HttpMethods => _httpMethods;
```
rynowak: Added this to ease the transition

src/Microsoft.AspNetCore.Routing/EndpointConstraints/IEndpointConstraint.cs
```diff
@@ -28,21 +27,31 @@ public interface IEndpointConstraintMetadata
     {
     }
 
-    public struct EndpointSelectorCandidate
+    public readonly struct EndpointSelectorCandidate
```
rynowak: @JamesNK

rynowak: #JamesPoints

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
```diff
@@ -2,56 +2,149 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Runtime.CompilerServices;
 
 namespace Microsoft.AspNetCore.Routing.Matchers
 {
-    internal class CandidateSet
+    public sealed class CandidateSet
```
rynowak: Suggestions welcome on making this spicy boy more efficient.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
```diff
@@ -2,56 +2,149 @@
...
-        //    Group1: Candidates[2], Candidates[3], Candidates[4]
-        //
-        // The groups array would look like: { 0, 2, 5, }
-        public readonly int[] Groups;
-
-        public CandidateSet(Candidate[] candidates, int[] groups)
+        private CandidateState _state0;
+        private CandidateState _state1;
+        private CandidateState _state2;
+        private CandidateState _state3;
```
rynowak: 4 is a number I picked.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
```diff
@@ -2,56 +2,149 @@
...
+                    _state0 = new CandidateState(candidates[0].Endpoint, candidates[0].Score);
+                    _state1 = new CandidateState(candidates[1].Endpoint, candidates[1].Score);
+                    _state2 = new CandidateState(candidates[2].Endpoint, candidates[2].Score);
+                    _state3 = new CandidateState(candidates[3].Endpoint, candidates[3].Score);
+
+                    _additionalCandidates = new CandidateState[candidates.Length - 4];
+                    for (var i = 4; i < candidates.Length; i++)
+                    {
+                        _additionalCandidates[i - 4] = new CandidateState(candidates[i].Endpoint, candidates[i].Score);
+                    }
```
rynowak: It appears like there's a lot of memory copying happening here, but my profiling so far shows this cost to be pretty minimal.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
```diff
@@ -0,0 +1,81 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    internal class DefaultEndpointSelector : EndpointSelector
```
rynowak: So my plan is that this is the `EndpointSelector` that we *want* to use. MVC will override it when it needs to.

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
```diff
@@ -48,115 +46,34 @@ public sealed override Task MatchAsync(HttpContext httpContext, IEndpointFeature
...
-            // Now we'll go group by group to capture route values, process constraints,
+            // Now we'll iterate each endpoint to capture route values, process constraints,
             // and process complex segments.
-            //
-            // Perf: using groups.Length - 1 here to elide the bounds check. We're relying
-            // on assumptions of how Groups works.
-            var candidatesArray = candidates.Candidates;
-            var groups = candidates.Groups;
-
-            for (var i = 0; i < groups.Length - 1; i++)
```
rynowak: Getting rid of 'groups' made this really simple and fast. Now if you look at a profile, the main costs are things like string comparisons and dictionary lookups, not our data structures.

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -4,32 +4,37 @@
...
             MatchProcessorFactory matchProcessorFactory,
-            EndpointSelector endpointSelector)
+            EndpointSelector selector,
+            IEnumerable<MatcherPolicy> policies)
         {
-            _matchProcessorFactory = matchProcessorFactory ?? throw new ArgumentNullException(nameof(matchProcessorFactory));
-            _endpointSelector = endpointSelector ?? throw new ArgumentNullException(nameof(endpointSelector));
+            _matchProcessorFactory = matchProcessorFactory;
+            _selector = selector;
+            _policies = policies.OrderBy(p => p.Order).ToArray();
```
rynowak: OK so this is where you see `MatcherPolicy` for the first time. The idea with this is that it's a feature-collection or capability-based thing. Right now matcher policies can:
- order endpoints
- add nodes to the graph
- both
- neither

If we add more capabilites in the future, we don't need to plumb them through everything.

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -4,32 +4,37 @@
...
+
+            // Taking care to use _policies, which has been sorted.
+            _nodeBuilders = _policies.OfType<INodeBuilderPolicy>().ToArray();
+            _comparer = new MatcherEndpointComparer(_policies.OfType<IEndpointComparerPolicy>().ToArray());
         }
 
         public override void AddEndpoint(MatcherEndpoint endpoint)
         {
-            _entries.Add(new MatcherBuilderEntry(endpoint));
```
rynowak: I got rid of `MatcherBuilderEntry`. Endpoint is good enough now on its own.

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -209,46 +218,56 @@ public override Matcher Build()
             var root = BuildDfaTree();
 
             var states = new List<DfaState>();
-            var tables = new List<JumpTableBuilder>();
-            AddNode(root, states, tables);
+            var tableBuilders = new List<(JumpTableBuilder pathBuilder, PolicyJumpTableBuilder policyBuilder)>();
```
rynowak: Ok I'm sure the diff is kinda confusing at this point if you're reading in order. We have two kinds of jump tables now:
- path jump tables
- policy jump tables

I want to make sure theres a lookup from state -> jump table, so I just make each state have a tuple of builders, because that seemed easiest.

src/Microsoft.AspNetCore.Routing/Matchers/EndpointMetadataComparer.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public abstract class EndpointMetadataComparer<TMetadata> : IComparer<Endpoint> where TMetadata : class
```
rynowak: Base class provided for implementing policies that order based on metadata. Look at how it's used in the HTTP method feature.

src/Microsoft.AspNetCore.Routing/Matchers/HttpMethodMatcherPolicy.cs
```diff
@@ -0,0 +1,167 @@
...
+using Microsoft.AspNetCore.Routing.Metadata;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public sealed class HttpMethodEndpointSelectorPolicy : MatcherPolicy, IEndpointComparerPolicy, INodeBuilderPolicy
+    {
+        public IComparer<Endpoint> Comparer => EndpointMetadataComparer<IHttpMethodMetadata>.Default;
+
+        public override int Order => -1000;
```
rynowak: Order chosen to come before naively written policies.

src/Microsoft.AspNetCore.Routing/Matchers/MatcherPolicy.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public abstract class MatcherPolicy
```
rynowak: I considered also making this a marker interface. We need to define some kind of type because DI.

test/Microsoft.AspNetCore.Routing.Tests/EndpointConstraints/EndpointSelectorTests.cs
```diff
@@ -19,37 +20,37 @@ namespace Microsoft.AspNetCore.Routing.EndpointConstraints
     public class EndpointSelectorTests
     {
         [Fact]
-        public void SelectBestCandidate_MultipleEndpoints_BestMatchSelected()
+        public async Task SelectBestCandidate_MultipleEndpoints_BestMatchSelected()
```
rynowak: I did a lot of cleanup here. Mainly I did what I needed to keep them working.

rynowak:
Thanks for digging in here!

I also want to call out that I have not written any unit tests for the new codez.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matchers/MatcherSelectCandidatesSmallEntryCountBenchmark.cs
rynowak: Ok, I'm going to go with `FindCandidateSet` - that will be a little more distinct.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
rynowak: I'm going to document that it's a guess for now. We can collect data on this but I don't think it's entirely pressing.

rynowak: Generally how this will come into play is for a controller that has 4 actions with different verbs and the same route template.

src/Microsoft.AspNetCore.Routing/Matchers/INodeBuilderPolicy.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public interface INodeBuilderPolicy
+    {
+        bool AppliesToNode(IReadOnlyList<Endpoint> endpoints);
+
+        IReadOnlyList<PolicyNodeEdge> GetEdges(IReadOnlyList<Endpoint> endpoints);
+
+        PolicyJumpTable BuildJumpTable(int exitDestination, IReadOnlyList<PolicyJumpTableEdge> edges);
+    }
```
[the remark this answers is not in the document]
rynowak: > choose an endpoint from a list of possibilities using the current request.

`EndpointSelector` is exactly that, so I think we're good 👍

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintEndpointSelector.cs
```diff
@@ -32,37 +29,50 @@ internal class EndpointSelector
...
+                return Task.CompletedTask;
             }
             else if (finalMatches.Count == 1)
             {
-                var selectedEndpoint = finalMatches[0];
+                var endpoint = finalMatches[0].Endpoint;
+                var values = finalMatches[0].Values;
+
+                feature.Endpoint = endpoint;
+                feature.Invoker = (endpoint as MatcherEndpoint)?.Invoker;
```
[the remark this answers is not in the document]
rynowak: Yeah there are two reasons for the separation:
1. `Endpoint` needs to go into HTTP abstractions - but we don't want to move things like `RoutePattern` down to that layer.
2. Its useful to be able to define endpoints that don't participate in matching. For instance if you wanted an endpoint that only participates in URL generation.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
```diff
@@ -2,56 +2,149 @@
...
-        public CandidateSet(Candidate[] candidates, int[] groups)
+        private CandidateState _state0;
+        private CandidateState _state1;
+        private CandidateState _state2;
+        private CandidateState _state3;
+
+        private CandidateState[] _additionalCandidates;
+
+        // Provided to make testing possible/easy.
+        public CandidateSet(MatcherEndpoint[] endpoints)
```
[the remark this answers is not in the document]
rynowak: This is to let someone implementing an `EndpointSelector` write tests

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
```diff
@@ -2,56 +2,149 @@
...
+                    {
+                        _additionalCandidates[i - 4] = new CandidateState(candidates[i].Endpoint, candidates[i].Score);
+                    }
+                    break;
             }
+        }
+
+        public int Count { get; }
+
+        public ref CandidateState this[int index]
```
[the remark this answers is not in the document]
rynowak: Sure I'll add a comment.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateState.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public struct CandidateState
+    {
+        // Provided for testability
+        public CandidateState(MatcherEndpoint endpoint)
```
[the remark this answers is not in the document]
rynowak: This is to let someone implementing an `EndpointSelector` write tests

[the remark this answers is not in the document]
rynowak: Will make this call the other.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            int? foundScore = null;
+            for (var i = 0; i < candidates.Count; i++)
+            {
+                ref var state = ref candidates[i];
+
+                if (state.IsValidCandiate && foundScore == null)
+                {
+                    // This is the first match we've seen - speculatively assign it.
+                    feature.Endpoint = state.Endpoint;
```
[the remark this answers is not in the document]
rynowak: Do you have a specific concern here? This only avoids overhead in the case that there's an ambiguity, which I'm not too worried about.

I'll try rewriting this with locals and see if it's more sensible.

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
```diff
@@ -48,115 +46,34 @@ public sealed override Task MatchAsync(HttpContext httpContext, IEndpointFeature
...
-            ReadOnlySpan<PathSegment> segments,
-            ReadOnlySpan<Candidate> group,
-            BitArray members,
-            RouteValueDictionary[] groupValues)
-        {
-            var hasMatch = false;
-            for (var i = 0; i < group.Length; i++)
-            {
-                // PERF: specifically not copying group[i] into a local. It's a relatively
+                // PERF: specifically not copying candidatesArray[i] into a local. It's a relatively
```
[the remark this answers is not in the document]
rynowak: I'll measure it and see if it helps. I had the same thought, but I don't want to tweak this code without a specific outcome and measurement.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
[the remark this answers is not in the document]
rynowak: I decided it's more sensible 😆

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -38,48 +43,48 @@ public DfaNode BuildDfaTree()
             // because a 'parameter' node can also traverse the same paths that literal nodes
             // traverse. This means that we need to order the entries first, or else we will
             // miss possible edges in the DFA.
-            _entries.Sort();
+            _endpoints.Sort((x, y) => _comparer.Compare(x, y));
```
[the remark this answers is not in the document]
rynowak: fixed

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -209,46 +218,56 @@ public override Matcher Build()
...
         }
 
-        private int AddNode(DfaNode node, List<DfaState> states, List<JumpTableBuilder> tables)
+        private int AddNode(
+            DfaNode node,
+            List<DfaState> states,
+            List<(JumpTableBuilder pathBuilder, PolicyJumpTableBuilder policyBuilder)> tableBuilders)
         {
-            node.Matches.Sort();
+            node.Matches.Sort((x, y) => _comparer.Compare(x, y));
```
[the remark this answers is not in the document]
rynowak: fixed

src/Microsoft.AspNetCore.Routing/Matchers/DfaNode.cs
```diff
@@ -25,15 +26,46 @@ public DfaNode()
...
 
         public DfaNode Parameters { get; set; }
 
         public DfaNode CatchAll { get; set; }
 
+        public INodeBuilderPolicy NodeBuilder { get; set; }
+
+        public Dictionary<object, DfaNode> PolicyEdges { get; }
+
+        public void TraversePostOrder(Action<DfaNode> visitor)
```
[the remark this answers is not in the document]
rynowak: https://en.wikipedia.org/wiki/Tree_traversal

src/Microsoft.AspNetCore.Routing/Matchers/HttpMethodMatcherPolicy.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            var edges = new List<PolicyNodeEdge>();
+            foreach (var kvp in dictionary)
+            {
+                edges.Add(new PolicyNodeEdge(kvp.Key, kvp.Value));
+            }
+
+            return edges;
+
+            IEnumerable<string> CollectHttpMethods()
```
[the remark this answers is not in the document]
rynowak: No guidelines other that be reasonable. Originally when I wrote this code, this was good for readabillity but not really anymore so taking it out.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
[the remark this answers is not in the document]
JamesNK: I don't know if it is a big concern but people could implement their own `IEndpointFeature` and might have logic execute when those properties are set.

src/Microsoft.AspNetCore.Routing/Matchers/DfaNode.cs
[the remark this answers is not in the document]
JamesNK: Add a comment with a link to https://en.wikipedia.org/wiki/Tree_traversal#Post-order_(LRN) ?

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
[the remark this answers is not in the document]
rynowak: Yep. it turns out using ref was an improvment!

rynowak:
I've pushed updates to address most of the feedback so far (thanks!). I'll work on adding unit tests for this tomorrow.

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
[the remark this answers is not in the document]
JamesNK: image

src/Microsoft.AspNetCore.Routing/Matchers/DfaNode.cs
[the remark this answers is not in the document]
rynowak: ehhhhh do you think it makes it clearer if I just rename this to visit? I don't feel like I feel like I must be doing something wrong if I need to document basic concepts like tree traversal

JamesNK:
Ok. Looks awesome so far!

src/Microsoft.AspNetCore.Routing/Matchers/DfaNode.cs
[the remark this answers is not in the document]
JamesNK: The "post order" terminology is what I'm unfamiliar with.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
```diff
@@ -2,56 +2,154 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Runtime.CompilerServices;
 
 namespace Microsoft.AspNetCore.Routing.Matchers
 {
-    internal class CandidateSet
+    public sealed class CandidateSet
```
Tornhoof: Personally, I'd expect a type called CandidateSet to implement ISet.

src/Microsoft.AspNetCore.Routing/Matchers/HttpMethodMatcherPolicy.cs
```diff
@@ -0,0 +1,166 @@
...
+        public PolicyJumpTable BuildJumpTable(int exitDestination, IReadOnlyList<PolicyJumpTableEdge> edges)
+        {
+            var dictionary = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
+            for (var i = 0; i < edges.Count; i++)
+            {
+                // We create this data, so it's safe to cast it to a string.
+                dictionary.Add((string)edges[i].State, edges[i].Destination);
+            }
+
+            if (dictionary.TryGetValue("*", out var matchesAnyVerb))
```
Tornhoof: Nit: AnyMethod from above?

src/Microsoft.AspNetCore.Routing/Matchers/HttpMethodMatcherPolicy.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                // If we have endpoints that match any HTTP method, use that as the exit.
+                exitDestination = matchesAnyVerb;
+                dictionary.Remove("*");
```
Tornhoof: Nit: AnyMethod from above?

Tornhoof:
A more general question:
Matcher.cs and all its implementations are using ``Task MatchAsync``, that DfaMatcher default implementation is synchronous. So why not ValueTask?

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
rynowak: Do you have a suggestion? I have no plans to introduce `ISet` or other high level collection APIs here due to pref cost.

src/Microsoft.AspNetCore.Routing/Matchers/HttpMethodMatcherPolicy.cs
rynowak: thanks!

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
Tornhoof: As far as I understand it, it's a set (duplicate candidatestates wouldn't make sense (to me)), so the name is better than anything I can think of.

src/Microsoft.AspNetCore.Routing/Matchers/CandidateSet.cs
cartermp: `boy` --> `boi`

rynowak: noice

rynowak:
@JamesNK @kichalla - updated and removed WIP

src/Microsoft.AspNetCore.Routing/EndpointConstraints/EndpointConstraintEndpointSelector.cs
```diff
@@ -72,31 +82,60 @@ public Endpoint SelectBestCandidate(HttpContext context, IReadOnlyList<Endpoint>
...
             var candidates = new List<EndpointSelectorCandidate>();
 
             // Perf: Avoid allocations
-            for (var i = 0; i < endpoints.Count; i++)
+            for (var i = 0; i < candidateSet.Count; i++)
             {
-                var endpoint = endpoints[i];
-                var constraints = _endpointConstraintCache.GetEndpointConstraints(context, endpoint);
-                candidates.Add(new EndpointSelectorCandidate(endpoint, constraints));
+                if (candidateSet[i].IsValidCandidate)
```
JamesNK: Assign candidateSet[i] to variable?

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcher.cs
```diff
@@ -46,117 +42,40 @@ public sealed override Task MatchAsync(HttpContext httpContext, IEndpointFeature
...
-                    segments,
-                    group,
-                    members,
-                    groupValues))
-                {
-                    // We must have some matches because FilterGroup returned true.
+                // Reminder!
+                // candidate: readonly data about the endpoint and how to match
+                // state: mutable storarge for our processing
+                ref var candiate = ref candidates[i];
```
JamesNK: candidate

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -267,54 +286,96 @@ private int AddNode(DfaNode node, List<DfaState> states, List<JumpTableBuilder>
...
+        // Builds an array of candidates for a node, assigns a 'score' for each
+        // endpoint.
+        internal Candidate[] CreateCandidates(IReadOnlyList<MatcherEndpoint> endpoints)
+        {
+            if (endpoints.Count == 0)
+            {
+                return Array.Empty<Candidate>();
+            }
+            
+            var candiates = new List<Candidate>();
```
JamesNK: minor optimization: This creates a candidate for each endpoint right? Pre-initialize the list with the right capacity? Or create an array and assign indexes to avoid ToArray call?

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
```diff
@@ -437,5 +499,57 @@ private static bool HasAdditionalRequiredSegments(MatcherBuilderEntry entry, int
...
+
+                    // This node-builder does apply to this node, so we need to create new nodes for each edge,
+                    // and then attach them to the parent.
+                    var edges = nodeBuilder.GetEdges(parent.Matches);
+                    for (var k = 0; k < edges.Count; k++)
+                    {
+                        var edge = edges[k];
+
+                        var next = new DfaNode();
+                        next.Matches.AddRange(edge.Endpoints.Cast<MatcherEndpoint>().ToArray());
```
JamesNK: These endpoints will always be `MatcherEndpoint`? No need to type check and provide a better error message?

src/Microsoft.AspNetCore.Routing/Matchers/EndpointMetadataComparer.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        public static EndpointMetadataComparer<TMetadata> Default = new DefaultComparer<TMetadata>();
```
JamesNK: readonly

src/Microsoft.AspNetCore.Routing/Matchers/HttpMethodMatcherPolicy.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            return new DictionaryPolicyJumpTable(exitDestination, dictionary);
+        }
+
+        private Endpoint CreateRejectionEndpoint(IEnumerable<string> httpMethods)
+        {
+            var allow = string.Join(", ", httpMethods);
+            return new MatcherEndpoint(
```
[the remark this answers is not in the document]
JamesNK: Can you create an issue to track please

src/Microsoft.AspNetCore.Routing/Matchers/PolicyJumpTableEdge.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public readonly struct PolicyJumpTableEdge
+    {
+        public PolicyJumpTableEdge(object state, int destination)
+        {
+            State = state;
```
JamesNK: Can state be null?

src/Microsoft.AspNetCore.Routing/Matchers/PolicyNodeEdge.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public readonly struct PolicyNodeEdge
+    {
+        public PolicyNodeEdge(object state, IReadOnlyList<Endpoint> endpoints)
+        {
+            State = state;
+            Endpoints = endpoints;
```
JamesNK: Can either of these be null?

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
rynowak: sure

src/Microsoft.AspNetCore.Routing/Matchers/DfaMatcherBuilder.cs
rynowak: Related to https://github.com/aspnet/Routing/issues/648 - I'll leave a comment.

src/Microsoft.AspNetCore.Routing/Matchers/HttpMethodMatcherPolicy.cs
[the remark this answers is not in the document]
rynowak: https://github.com/aspnet/Routing/issues/648 already tracks this YO

PR #647, opened 2018-07-22
Change Branding

rynowak:
For preview one the branding is:

new thing = UseGlobalRouting/UseEndpoint
old thing = UseRouter

We're going to drop the name Dispatcher everywhere and make sure that we
position our new work as 'new and improved routing' instead of
introducing a new product/concept name.

We're not totally sure of the term Global yet, but it's what we're doing
for preview 1. Suggestions welcome for dicussion after we do the first
preview :)


benchmarkapps/Benchmarks/Program.cs
```diff
@@ -29,22 +29,22 @@ public static IWebHostBuilder GetWebHostBuilder(string[] args)
                     .UseKestrel();
 
             var scenario = config["scenarios"]?.ToLower();
-            if (scenario == "plaintextdispatcher")
+            if (scenario == "plaintextdispatcher" || scenario == "plaintextglobalrouting")
```
rynowak: I've added new names that reflect the correct branding. I need to talk to some folks about what do to change what the perf server does. I plan to follow up and remove the old names after updating the perf automation.

samples/DispatcherSample.Web/DispatcherSample.Web.csproj
```diff
@@ -1,17 +0,0 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
```
rynowak: It seemed like the simplest thing to do to update the sample to work with multiple startup classes instead of having separate projects.

src/Microsoft.AspNetCore.Routing/EndpointOptions.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.AspNetCore.Routing
 {
-    public class DispatcherOptions
+    public class EndpointOptions
```
rynowak: I'm still not 100% happy with our options for what to do with this. It doesn't seem to me like adding the data sources to `RoutingOptions` is a great thing to do... So I did something simple for now. If anyone has an amazing suggestion I'm looking for ideas.

src/Microsoft.AspNetCore.Routing/DependencyInjection/DispatcherServiceCollectionExtensions.cs
```diff
@@ -2,73 +2,36 @@
...
-using Microsoft.AspNetCore.Routing.Matchers;
-using Microsoft.Extensions.DependencyInjection.Extensions;
-using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
+    // This whole class is temporary until we can remove its usage from MVC
     public static class DispatcherServiceCollectionExtensions
     {
+        // This whole class is temporary until we can remove its usage from MVC
```
rynowak: xD

My plan is to react to verify that MVC still passes with these changes, and then:
1. Merge this PR
2. Update branding in MVC
3. Remove these stub `Dispatcher` things from here

rynowak:
So the good news is that this doesn't require an immediate reaction in MVC.

samples/DispatcherSample.Web/DispatcherSample.Web.csproj
JamesNK: 👍 

I think I did this before I understood how a single site could be initialized with different startup classes.

JamesNK:
> We're not totally sure of the term Global yet, but it's what we're doing
> for preview 1. Suggestions welcome for dicussion after we do the first
> preview :)

What about endpoint routing?

`UseGlobalRouting()` -> `UseEndpointRouting()`

`UseEndpoint()`

rynowak:
I suggested `UseEndpointRouting` - I think it's much better than `UseGlobalRouting` but it's not up to me 👼 - I got agreement with the powers that be `UseGlobalRouting` is "good enough for the first preview". I'm sure we'll discuss more.

PR #651, opened 2018-07-23
Remove AddDispatcher/UseDispatcher

rynowak:


rynowak:
FYI only - just following up on previous cleanup

PR #654, opened 2018-07-23
Use TemplateBinder instead of decision tree for named routes.

kichalla:


src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -45,11 +45,9 @@ public IEnumerable<Endpoint> FindEndpoints(RouteValuesBasedEndpointFinderContext
                     context.ExplicitValues,
                     context.AmbientValues);
             }
-            else if (_namedMatches.TryGetValue(context.RouteName, out var linkGenerationTree))
+            else if (_namedMatchResults.TryGetValue(context.RouteName, out var namedMatchResults))
             {
-                matchResults = linkGenerationTree.GetMatches(
-                    context.ExplicitValues,
-                    context.AmbientValues);
+                matchResults = namedMatchResults;
```
kichalla: I wanted to write a useful comment here as to why it is this way but felt it's going to need a good big explanation (which probably we want to avoid in a comment?)

test/Microsoft.AspNetCore.Routing.Tests/RouteValueBasedEndpointFinderTest.cs
```diff
@@ -169,6 +170,31 @@ public void EndpointDataSource_ChangeCallback_Refreshes_OutboundMatches()
                 });
         }
 
+        [Fact]
+        public void FindEndpoints_ReturnsEndpoint_WhenLookedUpByRouteName()
```
kichalla: This test failed before my product change

kichalla: Also the earlier failing MVC functional tests related to this pass fine now.

test/Microsoft.AspNetCore.Routing.Tests/RouteValueBasedEndpointFinderTest.cs
```diff
@@ -169,6 +170,31 @@ public void EndpointDataSource_ChangeCallback_Refreshes_OutboundMatches()
...
+                {
+                    ExplicitValues = new RouteValueDictionary(new { id = 10 }),
+                    AmbientValues = new RouteValueDictionary(new { controller = "Home", action = "Index" }),
+                    RouteName = "OrdersApi"
+                });
+
+            // Assert
+            var actual = Assert.Single(foundEndpoints);
+            Assert.Same(expected, actual);
+        }
```
rynowak: Do we have a test for the failure case? Like this, but without values for controller/action

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
kichalla: Explanation for the change:
Decision tree is built using the 'required values' of actions.
- When generating a url using route values, we check the explicitly supplied route values + ambient values to see if they match the required values based tree.
- When generating a url using route name, route values for controller, action etc. might not be provided (this is expected because as a user i want to avoid writing all those and instead chose to use a route name which is quick). So since these values are not provided and might not be even in ambient values, we decision tree would fail to find a match.

test/Microsoft.AspNetCore.Routing.Tests/RouteValueBasedEndpointFinderTest.cs
kichalla: I am a bit unclear @rynowak . There are no explicit values for controller/action in the above test actually.

kichalla: I am planning to add more tests to GlobalRoutingUrlHelperTest in MVC

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
rynowak: I think this is a good explanation, and importantly it's not too big to leave in a comment. I'd suggest a big comment at the beginning of the method body that explains what cases matter and why they are different. examples go a looooong way.

src/Microsoft.AspNetCore.Routing/RouteValuesBasedEndpointFinder.cs
```diff
@@ -78,15 +76,16 @@ private void BuildOutboundMatches()
...
             // as refresh of new endpoints happens within a lock and also these fields are not publicly accessible.
-            var (allMatches, namedMatches) = GetOutboundMatches();
-            _namedMatches = GetNamedMatches(namedMatches);
+            var (allMatches, namedMatchResults) = GetOutboundMatches();
+            _namedMatchResults = namedMatchResults;
             _allMatchesLinkGenerationTree = new LinkGenerationDecisionTree(allMatches.ToArray());
         }
 
-        protected virtual (IEnumerable<OutboundMatch>, IDictionary<string, List<OutboundMatch>>) GetOutboundMatches()
+        protected virtual (IEnumerable<OutboundMatch>, IDictionary<string, List<OutboundMatchResult>>) GetOutboundMatches()
```
rynowak: Is there a good reason for this to be protected? used in tests? a comment would help

kichalla:
https://github.com/aspnet/Routing/commit/7da1baf9d844472a421133e18a7b0ee03cadefed

PR #668, opened 2018-07-27
Remove EndpointConstraints

rynowak:
Adds IEndpointSelectorPolicy so that MVC can plug in to the
EndpointSelector to run action constraints.


src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
```diff
@@ -11,17 +11,34 @@ namespace Microsoft.AspNetCore.Routing.Matchers
...
         public override Task SelectAsync(
             HttpContext httpContext,
             IEndpointFeature feature,
-            CandidateSet candidates)
+            CandidateSet candidateSet)
         {
+            for (var i = 0; i < _policies.Length; i++)
+            {
+                _policies[i].Apply(httpContext, candidateSet);
+            }
```
rynowak: @JamesNK  - I think you're going to be happy about this.

This is a simple way to plug in a DI service that does a pass over the candidates and gets to remove some from consideration.

So, I spent a few hours this morning trying to wire it up so that MVC would replace the `EndpointSelector` wholesale when it detects some `IActionConstraint` implementations, and it was kinda a mess.

The main issues were:
1. Code duplication - I needed to write another thing does what `EndpointSelector` already does
2. Our DI doesn't support decorators

So I think it's probably warranted to step back from that cliff. 

This is going to be very reasonable perf-wise as well since it's using data structures we already have.  We still want to do the graph node thing for perf-sensitive features.

I still want to make MVC able to run `IActionConstraints` for compat. Reaction PR to follow (obviously) since this breaks lots of things. I expect I won't be able to merge this until your CORS work is done.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
```diff
@@ -11,17 +11,34 @@ namespace Microsoft.AspNetCore.Routing.Matchers
...
+        private readonly IEndpointSelectorPolicy[] _policies;
+        
+        public DefaultEndpointSelector(IEnumerable<MatcherPolicy> policies)
+        {
+            if (policies == null)
+            {
+                throw new ArgumentNullException(nameof(policies));
+            }
+
+            _policies = policies.OrderBy(p => p.Order).OfType<IEndpointSelectorPolicy>().ToArray();
```
JamesNK: My first impression is it feels a little weird that something has to inherit from `MatcherPolicy` and implement `IEndpointSelectorPolicy`.

JamesNK: Them both ending in Policy is also a little confusing.

JamesNK: And `policies` + `_policies`. Do you think Policy is being overloaded?

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
JamesNK: If someone wants per request DI do they access `HttpContext.RequestServices`?

src/Microsoft.AspNetCore.Routing/Matchers/IEndpointSelectorPolicy.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Matchers
+{
+    public interface IEndpointSelectorPolicy
+    {
+        void Apply(HttpContext httpContext, CandidateSet candidates);
```
JamesNK: Do you think async policies are a thing people would want?

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
rynowak: > (maybe all together in one back-compat policy that then executes the current action constraint selector)?

Yes, exactly.

> If someone wants per request DI do they access HttpContext.RequestServices?

Yes.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
rynowak: Well, the reason why it's separated this way is that there's a bunch of policies that all play together. Maybe it would make sense as a marker interface that they all inherit from as well....

I'll log an issue for this.

src/Microsoft.AspNetCore.Routing/Matchers/IEndpointSelectorPolicy.cs
rynowak: Yeah, I was little unsure about that. We've traditionally not offered async at this level and it's been fine so far. You can always make an argument that things should be async, so I'm not really swayed by speculation.

rynowak: I've tagged this for discussion as part of API review.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
rynowak: > If someone wants per request DI do they access HttpContext.RequestServices?

BTW

The thing about this (pre-rant in case it comes up)... If we did add support for handling per-request instances of things, we could do it by having a factory that calls into the service provider to register it per-request. For all intents and purposes it's equivalent to the user doing it themselves. 

The difference is twofold:
- The people that know they want per-request feel emotionally validated because we thought of them. 
- Another set of people will be very confused because they don't know what our guidance is. They see multiple ways of doing things, and they are unsure of what they are actually supposed to do. 

An example of the latter is filters. We have many ways of plugging in filters. Some of the most relevant ones to this discussion are `[TypeFilter]` and `[ServiceFilter]` both of these allow you to plug in a filter per request, but they do it in different ways.

As a result we get lots of questions about which one to use in which case. It's clear that what we build provides flexibility that folk don't need. You might go so far as to call it *vexibility*, because novices are vexxed.

src/Microsoft.AspNetCore.Routing/Matchers/DefaultEndpointSelector.cs
rynowak: I'll rename these variables for now. If we want to have a discussion about the naming of extensibility points lets do that during API review.

Issue #544, opened 2018-06-08
Link generation: Discard ambient values unless routing to the same address


kichalla:
Today ambient values are handled inconsistently across Conventional and Attribute routing.
**Conventional Routing**: Ambient values are ignored when trying to link to a different action
**Attribute Routing**: Ambient values are honored in this case.

We want ambient values to be removed when generating links to addresses other than own.

kichalla:
More detail: https://gist.github.com/rynowak/174c46b31f52d9674b55bf27dbf8b19d

PR #669, opened 2018-07-27
Discard ambient values when not generating links to same endpoint

kichalla:
Fixes https://github.com/aspnet/Routing/issues/544

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
+++ b/src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
@@ -92,7 +92,9 @@ namespace Microsoft.AspNetCore.Routing
 
             var templateValuesResult = templateBinder.GetValues(
                 ambientValues: context.AmbientValues,
-                values: context.ExplicitValues);
+                explicitValues: context.ExplicitValues,
+                endpoint.RequiredValues.Keys);
+
             if (templateValuesResult == null)
             {
                 // We're missing one of the required values for this route.
--- a/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Text.Encodings.Web;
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -65,156 +66,57 @@ public class TemplateBinder
         // Step 1: Get the list of values we're going to try to use to match and generate this URI
         public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteValueDictionary values)
         {
```
kichalla: I refactored this method as it was huge.

src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -92,7 +92,9 @@ private string GetLink(MatcherEndpoint endpoint, LinkGeneratorContext context)
 
             var templateValuesResult = templateBinder.GetValues(
                 ambientValues: context.AmbientValues,
-                values: context.ExplicitValues);
+                explicitValues: context.ExplicitValues,
+                endpoint.RequiredValues.Keys);
```
rynowak: How does this compile? It's a positional argument following named args. Did that make that allowed?

rynowak: Oh DOPE: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-2#non-trailing-named-arguments

they made that allowed

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -65,156 +66,57 @@ public class TemplateBinder
...
-                    }
-                }
-                else
-                {
-                    if (hasCurrentParameterValue)
-                    {
-                        context.Accept(parameterName, currentParameterValue);
-                    }
-                }
+            // In case a route template is not parameterized, we do not what the required parameters are, so look
```
rynowak: > we do not what

I do not what either.

Should this be `we do not know`?

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -65,156 +66,57 @@ public class TemplateBinder
...
+            // at required values of an endpoint to get the key names and then use them to decided if ambient values
```
rynowak: `decided` -> `decide`

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -65,156 +66,57 @@ public class TemplateBinder
...
+            // should be used.
```
rynowak: It would really help to include an example in the comments of what cases like this would look like. Feel free to add comments explaining how MVC uses it.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -414,6 +316,197 @@ private static bool IsRoutePartNonEmpty(object routePart)
             }
         }
 
+        private bool CanCopyParameterAmbientValues(
+            RouteValueDictionary ambientValues,
```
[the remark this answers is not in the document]
rynowak: It's not really nit. Especially in this code where the parameters are of similar types.

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -67,7 +67,7 @@ public void GetLink_Fail_ThrowsException()
         public void TryGetLink_Fail()
         {
             // Arrange
-            var endpoint = CreateEndpoint("{controller}/{action}");
+            var endpoint = EndpointHelper.CreateMatcherEndpoint("{controller}/{action}");
```
rynowak: Suggestion. Make this a `using static`. Seeing `Helper` all over the code just makes my eyes bleed.

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -1276,28 +1276,194 @@ public void GetLink_TwoOptionalParametersAfterDefault_LastValueFromAmbientValues
...
+                    {
+                        new { area = "Admin", controller = "Products", action = "Edit", id = 10 },
+                        new { area = "Admin", controller = "Products", action = "Edit" },
+                        new { area = "Admin", controller = "Products", action = "Edit" }
+                    },
+                    {
+                        new { controller = "Products", action = "Edit", id = 10 },
+                        new { controller = "Products", action = "Edit" },
+                        new { area = (string)null, controller = "Products", action = "Edit" }
+                    },
```
rynowak: It would be good to include a case with pages too, for instance linking from a page to a controller and the reverse.

test/Microsoft.AspNetCore.Routing.Tests/EndpointHelper.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Routing.Matchers;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    internal static class EndpointHelper
```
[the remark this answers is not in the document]
rynowak: Or just `EndpointFactory` 😆

test/Microsoft.AspNetCore.Routing.Tests/EndpointHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        public static MatcherEndpoint CreateMatcherEndpoint(
+            string template,
+            object defaults = null,
+            object requiredValues = null,
+            object constraints = null,
+            int order = 0,
+            string routeName = null,
+            EndpointMetadataCollection metadataCollection = null)
```
rynowak: Suggestions here to make this a little more general purpose.

Make the last parameter of this be `params object[] metadata`.
Also, suggestion, don't include route name here. It's pretty specific to one kind of test, and we can just inline `new RouteNameMetadata` there.

test/Microsoft.AspNetCore.Routing.Tests/EndpointHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+                    metadataCollection = new EndpointMetadataCollection(new[] { new RouteNameMetadata(routeName) });
+                }
+            }
+
+            // Typically this is how MvcEndpointDataSource does, i.e it creates an endpoint only if all the required
+            // values of an action can satisfy a route template.
+            if (requiredValues != null && defaults == null)
+            {
+                defaults = new RouteValueDictionary(requiredValues);
+            }
```
rynowak: I don't think this should be here. The caller should pass in the data they want.

kichalla (commented):
🆙 📅

test/Microsoft.AspNetCore.Routing.Tests/EndpointFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.AspNetCore.Routing
+{
+    internal static class EndpointFactory
+    {
+        public static MatcherEndpoint CreateMatcherEndpoint(
+            string template,
+            object defaults = null,
+            object requiredValues = null,
+            object nonInlineConstraints = null,
```
rynowak: can you move this above required values, and just call it constraints?

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -1276,28 +1276,208 @@ public void GetLink_TwoOptionalParametersAfterDefault_LastValueFromAmbientValues
...
+                        new { area = "Admin", controller = "Products", action = "Edit", id = 10 },
+                        new { area = "Admin", controller = "Products", action = "Edit" },
+                        new { area = "Admin", controller = "Products", action = "Edit" },
+                        new { area = "Admin", controller = "Products", action = "Edit" }
+                    },
+                    {
+                        new { controller = "Products", action = "Edit", id = 10 },
+                        new { controller = "Products", action = "Edit" },
+                        new { area = (string)null, controller = "Products", action = "Edit" },
+                        new { controller = "Products", action = "Edit" }
```
rynowak: Could you please add a test related to Razor Pages?

kichalla (commented):
🆙 📅

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -1276,28 +1276,292 @@ public void GetLink_TwoOptionalParametersAfterDefault_LastValueFromAmbientValues
...
+                        new { controller = "Products", action = "Edit", id = 10 },
+                        new { page = "Products/Edit" },
+                        new { page = "Products/Edit"},
+                        new { page = "Products/Edit" }
+                    },
+
+                    // link from a page to controller-action
+                    {
+                        new { page = "Products/Edit", id = 10 },
+                        new { controller = "Products", action = "Edit" },
```
rynowak: Note that in a case like this the required values for a controller would be `new { controller = "Products", action = "Edit", page = (string) null },`

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -342,7 +342,14 @@ private static bool IsRoutePartNonEmpty(object routePart)
                 var hasAmbientValue = ambientValues != null &&
                     ambientValues.TryGetValue(keyName, out ambientValue);
 
-                if (hasAmbientValue && !RoutePartsEqual(ambientValue, explicitValue))
+                // This indicates an explicit value was provided whose key does not exist in ambient values
+                // Example: Link from controller-action to a page or vice versa.
+                if (!hasAmbientValue)
+                {
+                    return false;
+                }
```
rynowak: I'm not sure that this is right. MVC should handle this case by making sure that all of the endpoints have appropriate `null` values. Basically, I would expect the set of required keys to be `controller, action, area, page` for all actions, for an application that uses areas, controllers, and pages.

rynowak: Basically, I think your test setup is wrong

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
kichalla: True, I agree and `area = (string) null`

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -414,6 +326,196 @@ private static bool IsRoutePartNonEmpty(object routePart)
...
+            foreach (var keyName in requiredKeys)
+            {
+                if (!explicitValues.TryGetValue(keyName, out var explicitValue))
+                {
+                    continue;
+                }
+
+                object ambientValue = null;
+                var hasAmbientValue = ambientValues != null &&
+                    ambientValues.TryGetValue(keyName, out ambientValue);
```
pranavkm: `out var`?

pranavkm (commented):
Looks good to me

rynowak (approved):

PR #673, opened 2018-07-30
Rename global routing to endpoint routing

JamesNK:


src/Microsoft.AspNetCore.Routing/Builder/EndpointRoutingApplicationBuilderExtensions.cs
```diff
@@ -8,27 +8,33 @@
 
 namespace Microsoft.AspNetCore.Builder
 {
-    public static class GlobalRoutingApplicationBuilderExtensions
+    public static class EndpointRoutingApplicationBuilderExtensions
     {
-        private const string GlobalRoutingRegisteredKey = "__GlobalRoutingMiddlewareRegistered";
+        private const string EndpointRoutingRegisteredKey = "__EndpointRoutingMiddlewareRegistered";
 
+        [Obsolete("Replaced with UseEndpointRouting.")]
```
JamesNK: Will other projects build when an `ObsoleteAttribute` is encountered?

benchmarkapps/Benchmarks/StartupUsingEndpointRouting.cs
```diff
@@ -43,7 +43,7 @@ public void ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app)
         {
-            app.UseGlobalRouting();
+            app.UseEndpointRouting();
```
JamesNK: Did we settle on `UseRouting`?

src/Microsoft.AspNetCore.Routing/Builder/EndpointRoutingApplicationBuilderExtensions.cs
rynowak: No, we have warnings as errors.

benchmarkapps/Benchmarks/StartupUsingEndpointRouting.cs
rynowak: I think we settled at defining  no user-facing API for registering new middleware.

I think a sensible place to land would be to define the middleware as pubternal and register it from MVC.

benchmarkapps/Benchmarks/Program.cs
```diff
@@ -29,11 +29,11 @@ public static IWebHostBuilder GetWebHostBuilder(string[] args)
                     .UseKestrel();
 
             var scenario = config["scenarios"]?.ToLower();
-            if (scenario == "plaintextdispatcher" || scenario == "plaintextglobalrouting")
+            if (scenario == "plaintextdispatcher" || scenario == "plaintextglobalrouting" || scenario == "plaintextendpointrouting")
```
rynowak: You can nuke `plaintextglobalrouting`. It's not used anywhere yet. I was waiting for the dust to settle on this before making changes in the DB.

benchmarkapps/Benchmarks/StartupUsingEndpointRouting.cs
rynowak: I think we should make sure that we don't leak these `UseXyz` methods into public API in 2.2.0 - could you just do `.UseMiddleware<>` here instead?

Basically we don't want to put the middleware in a public/supported context because then it lives forever. I think a reasonable thing to do might be to pubternal the middleware and then reference them directly from MVC

rynowak: Or you could move these use methods to `.Internal` namespace. I'm open to other ideas, but I want to make sure we don't accidentlly put something in public API by accident.

JamesNK:
🆙 📅

PR #689, opened 2018-08-02
Creating link generation template

kichalla:


src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -153,5 +153,7 @@ public string GetLink<TAddress>(HttpContext httpContext, TAddress address, objec
             object values,
             LinkOptions options,
             out string link);
+
+        public abstract LinkGenerationTemplate GetTemplate(HttpContext httpContext, object values, LinkOptions options);
```
kichalla: A quick prototype based on my earlier api change. Here I put the `GetTemplate` api within LinkGenerator itself. I feel this is fine compared to having a separate factory. what do you guys think?

rynowak: This seems fine. This is a good thing for users, how it's implemented is up to us 👍

src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -234,7 +247,41 @@ private RouteValueDictionary GetAmbientValues(HttpContext httpContext)
                     return feature.Values;
                 }
             }
-            return null;
+            return new RouteValueDictionary();
+        }
+
+        public override LinkGenerationTemplate GetTemplate(HttpContext httpContext, object values, LinkOptions options)
```
rynowak: This should probably start with a `TAddress` version, and then specialize from there

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public abstract class LinkGenerationTemplate
```
rynowak: I think this is a good start.

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        public abstract string MakeUrl(object values);
```
JamesNK: Does this need to be abstract? IMO just have a non-virtual method call MakeUrl with null LinkOptions.

src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -234,7 +305,27 @@ private RouteValueDictionary GetAmbientValues(HttpContext httpContext)
...
+            var matcherEndpoints = endpoints.OfType<MatcherEndpoint>();
+            if (!matcherEndpoints.Any())
+            {
+                return null;
+            }
+
+            return matcherEndpoints;
         }
+
+
```
JamesNK: nit: too liney

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
kichalla: I agree. Changed it now.

kichalla (commented):
🆙 📅

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.AspNetCore.Routing
+{
+    /// <summary>
+    /// Defines a contract to generate a URL from a template.
+    /// </summary>
+    public abstract class LinkGenerationTemplate
+    {
+        /// <summary>
+        /// Generates a URL with an absolute path.
```
JamesNK: Include details about the parameters in the summary when there are overloads. For these two overloads something like:

1. Generates a URL with an absolute path from route values.
2. Generates a URL with an absolute path from route values and the specified link options.

e.g. https://github.com/aspnet/Routing/blob/9f3edca05bb736a15b9e401121779b0e43e388f2/src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs#L579-L582

Do this everywhere in the PR's docs

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -6,78 +6,184 @@
...
+        /// <returns><c>true</c> if a URL was generated successfully, <c>false</c> otherwise.</returns>
         public bool TryGetLink(object values, LinkOptions options, out string link)
         {
             return TryGetLink(httpContext: null, routeName: null, values, options, out link);
         }
 
+        /// <summary>
+        /// Generates a URL with an absolute path.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with current request.</param>
```
JamesNK: Does the `HttpContext` have to be the one for the current request?

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -6,78 +6,184 @@
...
+        /// <param name="values">An object that contains route values.</param>
+        /// <returns>The generated absolute URL.</returns>
         public string GetLink(HttpContext httpContext, object values)
         {
             return GetLink(httpContext, routeName: null, values, options: null);
         }
 
+        /// <summary>
+        /// Tries generating a URL with an absolute path.
```
JamesNK: I like the summary int.TryParse has for the TryParse pattern - https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=netframework-4.7.2

Just have the standard summary and add "A return value indicates whether the operation succeeded." to the end.

When it comes to XML docs I tend to copy what other people have already done 😄

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -6,78 +6,184 @@
...
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with current request.</param>
+        /// <param name="values">An object that contains route values.</param>
+        /// <param name="link">The generated absolute URL.</param>
```
JamesNK: Follow what int.TryParse does for the return value docs.

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
```diff
@@ -0,0 +1,29 @@
...
+        public string MakeUrl(object values)
+        {
+            return MakeUrl(values, options: null);
+        }
+
+        /// <summary>
+        /// Generates a URL with an absolute path.
+        /// </summary>
+        /// <param name="values">An object that contains route values.</param>
+        /// <param name="options">The <see cref="LinkOptions"/>.</param>
```
JamesNK: Do these options override the default LinkOptions when specified? Worth mentioning?

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -88,28 +194,65 @@ public string GetLink(HttpContext httpContext, string routeName, object values,
...
             LinkOptions options,
             out string link);
 
+        /// <summary>
+        /// Generates a URL with an absolute path.
+        /// </summary>
+        /// <param name="address">The information used to look up endpoints for generating a URL.</param>
+        /// <param name="values">An object that contains route values.</param>
+        /// <returns>The generated absolute URL.</returns>
         public string GetLinkByAddress<TAddress>(TAddress address, object values)
```
JamesNK: +docs for typeparam. Copy to all the overloads.

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -88,28 +194,65 @@ public string GetLink(HttpContext httpContext, string routeName, object values,
...
+        /// <returns><c>true</c> if a URL was generated successfully, <c>false</c> otherwise.</returns>
         public abstract bool TryGetLink(
             HttpContext httpContext,
             string routeName,
             object values,
             LinkOptions options,
             out string link);
 
+        /// <summary>
+        /// Generates a URL with an absolute path.
```
JamesNK: Summary needs more info about what is special about GetLinkByAddress.

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -147,11 +313,107 @@ public string GetLinkByAddress<TAddress>(TAddress address, HttpContext httpConte
...
             out string link);
+
+        /// <summary>
+        /// Creates a template object to generate a URL.
+        /// </summary>
+        /// <param name="values">
+        /// An object that contains route values. These values are used to lookup endpoint(s).
+        /// </param>
+        /// <returns>
+        /// If an endpoint(s) was found succesffully, then this returns a template object representing that,
```
JamesNK: succesffully

image

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -147,11 +313,107 @@ public string GetLinkByAddress<TAddress>(TAddress address, HttpContext httpConte
...
+        /// <returns><c>true</c> if a URL was generated successfully, <c>false</c> otherwise.</returns>
         public abstract bool TryGetLinkByAddress<TAddress>(
             TAddress address,
             HttpContext httpContext,
             object values,
             LinkOptions options,
             out string link);
+
+        /// <summary>
+        /// Creates a template object to generate a URL.
```
JamesNK: nit: Summary says create, method is called Get. I view them as different things.

Change this to "Gets a `LinkGenerationTemplate` for an endpoint matching the specified route values."?

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -1282,6 +1282,169 @@ public void TryGetLink_WithCustomAddress_CanGenerateLink_RespectsLinkOptions_Sup
...
+
+            // Act3
+            link = template.MakeUrl(new { id = 25 });
+
+            // Assert3
+            Assert.Equal("/Product/Edit/25", link);
+        }
+
+        [Fact]
+        public void GetTemplate_ByCustomAddress_ReturnsTemplate()
```
JamesNK: GetTemplateByAddress_ByCustomAddress_ReturnsTemplate

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -1365,5 +1528,14 @@ public NameMetadata(string name)
             }
             public string Name { get; }
         }
+
+        private class RouteNameMetadata : IRouteNameMetadata
```
JamesNK: FYI this is now RouteValuesAddress and there is a public implementation to use

JamesNK (commented):
Code changes are good. Get @rynowak's feedback on the docs

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
rynowak: That's something we generally say to make it clear that you're not supposed to just new one up. It's only meaningful when you have a request.

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -6,78 +6,184 @@
...
         {
             return TryGetLink(httpContext: null, routeName: null, values, options, out link);
         }
 
+        /// <summary>
+        /// Generates a URL with an absolute path.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with current request.</param>
+        /// <param name="values">An object that contains route values.</param>
+        /// <returns>The generated absolute URL.</returns>
```
rynowak: Absolute path right?

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -6,78 +6,184 @@
...
+    /// <summary>
+    /// Defines a contract to generate URLs to endpoints.
+    /// </summary>
     public abstract class LinkGenerator
     {
+        /// <summary>
+        /// Generates a URL with an absolute path.
+        /// </summary>
+        /// <param name="values">An object that contains route values.</param>
+        /// <returns>The generated absolute URL.</returns>
```
rynowak: These are all absolute paths though right?

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -147,11 +313,107 @@ public string GetLinkByAddress<TAddress>(TAddress address, HttpContext httpConte
...
+        /// </summary>
+        /// <param name="address">The information used to look up endpoints for creating a template.</param>
+        /// <param name="httpContext">The <see cref="HttpContext"/> associated with current request.</param>
+        /// <returns>
+        /// If an endpoint(s) was found succesffully, then this returns a template object representing that,
+        /// <c>null</c> otherwise.
+        /// </returns>
+        public abstract LinkGenerationTemplate GetTemplateByAddress<TAddress>(
+            TAddress address,
+            HttpContext httpContext);
```
rynowak: HttpContext should be the first parameter

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -1282,6 +1282,169 @@ public void TryGetLink_WithCustomAddress_CanGenerateLink_RespectsLinkOptions_Sup
...
+                requiredValues: new { controller = "Customers", action = "Details" },
+                defaults: new { controller = "Customers", action = "Details" },
+                metadata: new NameMetadata("CustomerDetails"));
+            var linkGenerator = CreateLinkGenerator(new[] { endpoint1, endpoint2 }, new RouteOptions(), services);
+
+            // Act1
+            var template = linkGenerator.GetTemplateByAddress<INameMetadata>(new NameMetadata("CustomerDetails"));
+
+            // Assert1
+            Assert.NotNull(template);
```
rynowak: It would be much better for these kinds of tests if the test was focused on the fact that a template object was returned, ie: don't generate a URL, just inspect the details of the template.

kichalla: Sure, I agree.

kichalla (commented):
🆙 📅

rynowak (approved):

kichalla:
c8946a40e4311b535d50a79c35279f1a4ca27255

PR #692, opened 2018-08-03
Add XML docs to data sources and metadata

JamesNK:


src/Microsoft.AspNetCore.Routing/Metadata/HttpMethodMetadata.cs
```diff
@@ -8,14 +8,32 @@
...
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HttpMethodMetadata" /> class.
+        /// </summary>
+        /// <param name="httpMethods">
+        /// The HTTP methods used during routing.
+        /// An empty collection means any HTTP method will be accepted.
+        /// </param>
+        /// <param name="acceptCorsPreflight">A value indicating whether routing accepts CORS preflight requests.</param>
```
rynowak: Suggestion:

> >A value indicating whether the associated endpoint should accept CORS preflight requests

Issue #686, opened 2018-08-02
XML docs for 2.2.0 additions


rynowak:
# List of new things to add docs:

#### Routing
- [x] Endpoint
- [x] IEndpointFeature
- [x] EndpointFeature
- [x] EndpointMetadataCollection
- [x] LinkGenerator
- [x] LinkGeneratorContext
- [x] RouteData.RouteData(RouteValueDictionary values)
#### MVC
- [x] MvcOptions.EnableEndpointRouting

#### Routing
- [x] Everything new in `.Patterns`
- [x] EndpointDataSource
- [x] CompositeEndpointDataSource
- [x] DefaultEndpointDataSource
- [x] `IEndpointFinder<TAddress>`
- [x] INameMetadata
- [x] IRouteNameMetadata
- [x] IRequiredValuesMetadata
- [x] ISuppressLinkGenerationMetadata
- [x] RouteOptions.LowercaseQueryStrings
- [x] RouteValuesAddress
- [x] CandidateSet
- [x] CandidateState
- [x] EndpointSelector
- [x] MatcherEndpoint
- [x] HttpMethodMetadata
- [x] IHttpMethodMetadata
- [x] MatcherPolicy
- [x] IEndpointComparerPolicy
- [x] IEndpointSelectorPolicy
- [x] HttpMethodMatcherPolicy
- [x] `EndpointMetadataComparer<TMetadata>`

## P2 (cut)
- [ ] LinkGenerationDecisionTree.GetMatches
- [ ] INodeBuilderPolicy
- [ ] PolicyJumpTable
- [ ] PolicyJumpTableEdge
- [ ] PolicyNodeEdge
- [ ] InlineConstraint.InlineConstraint(RoutePatternConstraintReference other)
- [ ] RouteTemplate.RouteTemplate(RoutePattern other)
- [ ] RouteTemplate.ToRoutePattern
- [ ] TemplatePart.TemplatePart(RoutePatternPart other)
- [ ] TemplatePart.ToRoutePatternPart
- [ ] TemplateSegment.TemplateSegment(RoutePatternPathSegment other)
- [ ] TemplateSegment.ToRoutePatternPathSegment
- [ ] OutboundRouteEntry.Data

## SUPER CUTSVILLE (not long for this world)
- [ ] MatchProcessor
- [ ] MatchProcessorBase

rynowak:
@JamesNK - if you have the list from our API review could you help fill this out?

rynowak:
[...]

PR #695, opened 2018-08-06
Add docs to RouteValuesAddress

JamesNK:
Last docs for https://github.com/aspnet/Routing/issues/686 🎂


src/Microsoft.AspNetCore.Routing/RouteValuesAddress.cs
```diff
@@ -3,12 +3,24 @@
...
+        /// </summary>
         public string RouteName { get; set; }
 
+        /// <summary>
+        /// Gets or sets the route values that are explicitly specified.
+        /// </summary>
         public RouteValueDictionary ExplicitValues { get; set; }
 
+        /// <summary>
+        /// Gets or sets ambient routes values from the current HTTP request.
```
kichalla: route values

PR #698, opened 2018-08-07
Add support for GraphViz

rynowak:
Adds **internal** support for dumping a route table to GraphViz DOT
notation. This allows us to dump the DFA graph for a route table and
visualize it.

Example:
https://gist.github.com/rynowak/2b24e4a6a602ca6f9c4de3ec227d621b

### DIFF
--- a/samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
+++ b/samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
@@ -2,10 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Text;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Internal;
 using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Internal;
 using Microsoft.AspNetCore.Routing.Matching;
 using Microsoft.AspNetCore.Routing.Patterns;
 using Microsoft.Extensions.DependencyInjection;
@@ -77,6 +80,21 @@ namespace RoutingSample.Web
                         0,
                         EndpointMetadataCollection.Empty,
                         "withoptionalconstraints"),
+                    new MatcherEndpoint((next) => (httpContext) =>
+                    {
+                        using (var writer = new StreamWriter(httpContext.Response.Body, Encoding.UTF8, 1024, leaveOpen: true))
+                        {
+                            var graphWriter = httpContext.RequestServices.GetRequiredService<DfaGraphWriter>();
+                            var dataSource = httpContext.RequestServices.GetRequiredService<CompositeEndpointDataSource>();
+                            graphWriter.Write(dataSource, writer);
+                        }
+
+                        return Task.CompletedTask;
+                    },
+                    RoutePatternFactory.Parse("/graph"),
+                    0,
+                    new EndpointMetadataCollection(new HttpMethodMetadata(new[]{ "GET", })),
+                    "DFA Graph"),
                 });
 
             services.TryAddEnumerable(ServiceDescriptor.Singleton<EndpointDataSource>(endpointDataSource));
--- a/src/Microsoft.AspNetCore.Routing/Internal/DfaGraphWriter.cs
+++ b/src/Microsoft.AspNetCore.Routing/Internal/DfaGraphWriter.cs
@@ -1,0 +1,92 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.AspNetCore.Routing.Matching;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Routing.Internal
+{
+    /// <summary>
+    /// <para>
+    /// A singleton service that can be used to write the route table as a state machine
+    /// in GraphViz DOT language https://www.graphviz.org/doc/info/lang.html
+    /// </para>
+    /// <para>
+    /// You can use http://www.webgraphviz.com/ to visualize the results.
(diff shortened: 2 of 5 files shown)


rynowak:
image

rynowak:
I was working on this as part of a blog post, and this seems generally useful enough to have.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
```diff
@@ -229,5 +230,22 @@ public sealed override Task MatchAsync(HttpContext httpContext, IEndpointFeature
 
             return true;
         }
+
+        internal void ToGraph(TextWriter writer)
```
JamesNK: What do you think of wrapping the graphvis stuff in `#if DEBUG`?

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -303,6 +304,21 @@ public override int GetDestination(HttpContext httpContext)
...
+                foreach (var kvp in _destinations)
+                {
+                    writer.WriteLine($"{label} -> {kvp.Value} [label=\"HTTP: {kvp.Key}\"]");
+                }
+
+                foreach (var kvp in _corsPreflightDestinations)
+                {
+                    writer.WriteLine($"{label} -> {kvp.Value} [label=\"CORS Preflight: {kvp.Key}\"]");
+                }
+            }
```
JamesNK: What will be displayed for custom policies?

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
JamesNK: Would it be possible to make this an external app or does it require internal access to the DfaMatcher data?

JamesNK:
Cool idea!

I used graphviz about 15 years ago to do something similar for a custom workflow engine. I haven't thought about it in years :)

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
rynowak: Nothing right now, we can't deeply introspect custom jump tables. That would require a public API for this. If we're convinced its valuable, I'm not opposed to a pubternal API for this.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
rynowak: I don't like the put things like behind debug, because the debug builds only happen in this repo. As a user you have no power to 'undebug' the bits we ship.

Do you have any cool ideas about how we could make this accessible to application authors? Ultimately I want us to ship some kind of experience for users to be able to dump the graph. I didn't try to solve all of the worlds problems in this PR.

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
rynowak: Another idea that I have here is to make this part of the DFA builder, which might not require that.

kichalla (approved):
This is awesome!

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
rynowak: Yeah I coded this up last night and I've been stewing over it for about 18 hours.

rynowak: So I think what I want to do for now is turn this into a pubternal service that you can call given a data source. Then I'll add a sample to the routing sample that uses it in a static HTML page.

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
rynowak: I'm going to try doing this in the DFA matcher builder so we can capture custom policies as well.

rynowak:
Updated, feel free to take a look. This is a somewhat cleaner approach and will work just find with user-defined policies.

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
```diff
@@ -77,6 +80,21 @@ public void ConfigureServices(IServiceCollection services)
...
+                    new MatcherEndpoint((next) => (httpContext) =>
+                    {
+                        using (var writer = new StreamWriter(httpContext.Response.Body, Encoding.UTF8, 1024, leaveOpen: true))
+                        {
+                            var graphWriter = httpContext.RequestServices.GetRequiredService<DfaGraphWriter>();
+                            var dataSource = httpContext.RequestServices.GetRequiredService<CompositeEndpointDataSource>();
+                            graphWriter.Write(dataSource, writer);
+                        }
+
+                        return Task.CompletedTask;
```
rynowak: I started trying to add a page that would render the graph and quickly got lost in the morass of NPM+dotnet. I don't want to spend too much time on that now.

JamesNK (approved):
I like it :shipit: 

Add ToString info to MVC policies?

src/Microsoft.AspNetCore.Routing/Internal/DfaGraphWriter.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// <para>
+    /// This type has no support contract, and may be removed or changed at any time in
+    /// a future release.
+    /// </para>
+    /// </summary>
+    public class DfaGraphWriter
+    {
+        private readonly IServiceProvider _services;
+
+        public DfaGraphWriter(IServiceProvider services)
```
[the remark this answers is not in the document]
rynowak: DfaMatcherBuilder is a stateful transient. If you resolve this service and use it multiple times that would be broken.

rynowak:
> Add ToString info to MVC policies?

YEAH BOI

PR #701, opened 2018-08-08
Add EndpointBuilder

JamesNK:
Note that this is against master.

### DIFF
--- a/samples/RoutingSample.Web/HelloExtension/HelloAppBuilderExtensions.cs
+++ b/samples/RoutingSample.Web/HelloExtension/HelloAppBuilderExtensions.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Options;
+using RoutingSample.Web.HelloExtension;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class HelloAppBuilderExtensions
+    {
+        public static IApplicationBuilder UseHello(this IApplicationBuilder app, string greeter)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            return app.UseMiddleware<HelloMiddleware>(Options.Create(new HelloOptions
+            {
+                Greeter = greeter
+            }));
+        }
+    }
+}
--- a/samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
+++ b/samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
@@ -3,20 +3,16 @@
 
 using System;
 using System.Text;
+using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Internal;
-using Microsoft.AspNetCore.Routing;
-using Microsoft.AspNetCore.Routing.Matching;
-using Microsoft.AspNetCore.Routing.Patterns;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.DependencyInjection.Extensions;
 
 namespace RoutingSample.Web
 {
     public class UseEndpointRoutingStartup
     {
         private static readonly byte[] _homePayload = Encoding.UTF8.GetBytes("Endpoint Routing sample endpoints:" + Environment.NewLine + "/plaintext");
-        private static readonly byte[] _helloWorldPayload = Encoding.UTF8.GetBytes("Hello, World!");
+        private static readonly byte[] _plainTextPayload = Encoding.UTF8.GetBytes("Plain text!");
 
         public void ConfigureServices(IServiceCollection services)
         {
@@ -26,65 +22,59 @@ namespace RoutingSample.Web
             {
                 options.ConstraintMap.Add("endsWith", typeof(EndsWithStringMatchProcessor));
             });
-
-            var endpointDataSource = new DefaultEndpointDataSource(new[]
-                {
-                    new MatcherEndpoint((next) => (httpContext) =>
-                        {
-                            var response = httpContext.Response;
-                            var payloadLength = _homePayload.Length;
-                            response.StatusCode = 200;
-                            response.ContentType = "text/plain";
-                            response.ContentLength = payloadLength;
-                            return response.Body.WriteAsync(_homePayload, 0, payloadLength);
-                        },
-                        RoutePatternFactory.Parse("/"),
-                        0,
-                        EndpointMetadataCollection.Empty,
-                        "Home"),
-                    new MatcherEndpoint((next) => (httpContext) =>
-                        {
-                            var response = httpContext.Response;
-                            var payloadLength = _helloWorldPayload.Length;
-                            response.StatusCode = 200;
-                            response.ContentType = "text/plain";
-                            response.ContentLength = payloadLength;
-                            return response.Body.WriteAsync(_helloWorldPayload, 0, payloadLength);
-                        },
-                         RoutePatternFactory.Parse("/plaintext"),
-                        0,
-                        EndpointMetadataCollection.Empty,
-                        "Plaintext"),
-                    new MatcherEndpoint((next) => (httpContext) =>
-                        {
-                            var response = httpContext.Response;
-                            response.StatusCode = 200;
-                            response.ContentType = "text/plain";
-                            return response.WriteAsync("WithConstraints");
-                        },
-                        RoutePatternFactory.Parse("/withconstraints/{id:endsWith(_001)}"),
-                        0,
-                        EndpointMetadataCollection.Empty,
-                        "withconstraints"),
-                    new MatcherEndpoint((next) => (httpContext) =>
-                        {
-                            var response = httpContext.Response;
-                            response.StatusCode = 200;
-                            response.ContentType = "text/plain";
-                            return response.WriteAsync("withoptionalconstraints");
-                        },
-                        RoutePatternFactory.Parse("/withoptionalconstraints/{id:endsWith(_001)?}"),
-                        0,
-                        EndpointMetadataCollection.Empty,
-                        "withoptionalconstraints"),
-                });
-
-            services.TryAddEnumerable(ServiceDescriptor.Singleton<EndpointDataSource>(endpointDataSource));
         }
 
         public void Configure(Microsoft.AspNetCore.Builder.IApplicationBuilder app)
         {
-            app.UseEndpointRouting();
+            app.UseEndpointRouting(builder =>
+            {
+                builder.MapHello("/helloworld", "World");
+
+                builder.MapEndpoint(
+                    (next) => (httpContext) =>
+                    {
+                        var response = httpContext.Response;
+                        var payloadLength = _homePayload.Length;
+                        response.StatusCode = 200;
+                        response.ContentType = "text/plain";
+                        response.ContentLength = payloadLength;
+                        return response.Body.WriteAsync(_homePayload, 0, payloadLength);
+                    },
+                    "Home",
+                    "/");
+                builder.MapEndpoint(
+                    (next) => (httpContext) =>
+                    {
+                        var response = httpContext.Response;
+                        var payloadLength = _plainTextPayload.Length;
+                        response.StatusCode = 200;
+                        response.ContentType = "text/plain";
+                        response.ContentLength = payloadLength;
+                        return response.Body.WriteAsync(_plainTextPayload, 0, payloadLength);
+                    },
+                    "Plaintext",
+                    "/plaintext");
+                builder.MapEndpoint(
+                    (next) => (httpContext) =>
+                    {
+                        var response = httpContext.Response;
+                        response.StatusCode = 200;
+                        response.ContentType = "text/plain";
+                        return response.WriteAsync("WithConstraints");
+                    },
+                    "withconstraints",
+                    "/withconstraints/{id:endsWith(_001)}");
+                builder.MapEndpoint(
+                    (next) => (httpContext) =>
+                    {
+                        var response = httpContext.Response;
+                        response.StatusCode = 200;
+                        response.ContentType = "text/plain";
+                        return response.WriteAsync("withoptionalconstraints");
+                    },
+                    "withoptionalconstraints",
+                    "/withoptionalconstraints/{id:endsWith(_001)?}");
+            });
 
             // Imagine some more stuff here...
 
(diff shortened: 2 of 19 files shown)


samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
```diff
@@ -26,65 +22,59 @@ public void ConfigureServices(IServiceCollection services)
...
+                    {
+                        var response = httpContext.Response;
+                        var payloadLength = _homePayload.Length;
+                        response.StatusCode = 200;
+                        response.ContentType = "text/plain";
+                        response.ContentLength = payloadLength;
+                        return response.Body.WriteAsync(_homePayload, 0, payloadLength);
+                    },
+                    "Home",
+                    "/");
```
JamesNK: My thinking is MapEndpoint is for mapping a raw invoker.

MVC could add overloads for MapMvcEndpoint which would have MVC specific params. Need to think about how to get it to perform the logic that MvcDataSourceEndpoint runs for resolving MVC params to MVC actions and then create endpoints.

src/Microsoft.AspNetCore.Routing/EndpointDataSourceBuilder.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Builder;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public abstract class EndpointDataSourceBuilder
+    {
+        public abstract ICollection<EndpointBuilder> Endpoints { get; }
+
+        public abstract IApplicationBuilder New();
```
JamesNK: Is `New()` a common thing? I was wondering if it should be NewApplicationBuilder, or NewAppBuilder.

samples/RoutingSample.Web/HelloExtension/HelloAppBuilderExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System;
+using Microsoft.Extensions.Options;
+using RoutingSample.Web.HelloExtension;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class HelloAppBuilderExtensions
+    {
+        public static IApplicationBuilder UseHello(this IApplicationBuilder app, string greeter)
```
rynowak: Note that there's no real reason for this level of indirection - unless you're doing it for sampley reasons 😆

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
rynowak: What's the signature here, basically the same as `Use(...)` - seems pretty right for a first go.

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
```diff
@@ -26,65 +22,59 @@ public void ConfigureServices(IServiceCollection services)
...
-                        RoutePatternFactory.Parse("/withoptionalconstraints/{id:endsWith(_001)?}"),
-                        0,
-                        EndpointMetadataCollection.Empty,
-                        "withoptionalconstraints"),
-                });
-
-            services.TryAddEnumerable(ServiceDescriptor.Singleton<EndpointDataSource>(endpointDataSource));
         }
 
         public void Configure(Microsoft.AspNetCore.Builder.IApplicationBuilder app)
```
rynowak: Another idea that we've been discussing is making the `EndpointDataSourceBuilder` a parameter here. So we would register this as a DI service, and you would resolve it into configuration.

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class MapEndpointEndpointDataSourceBuilderExtensions
+    {
+        public static EndpointBuilder MapEndpoint(
+            this EndpointDataSourceBuilder builder,
+            Func<RequestDelegate, RequestDelegate> invoker,
+            string name,
```
rynowak: this should be `displayName`, and come after route pattern (like matcher endpoint). It would be good to have these args follow the same order as `MatcherEndpoint` if possible.

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            return MapEndpoint(builder, invoker, name, pattern, null);
+        }
+
+        public static EndpointBuilder MapEndpoint(
+            this EndpointDataSourceBuilder builder,
+            Func<RequestDelegate, RequestDelegate> invoker,
+            string name,
+            string template,
+            IList<object> metadata)
```
rynowak: `params object[]` ?

samples/RoutingSample.Web/HelloExtension/HelloAppBuilderExtensions.cs
JamesNK: I figured you did it this way because health check already had middleware, but I left it like this to functional test this approach worked.

src/Microsoft.AspNetCore.Routing/DefaultEndpointDataSourceBuilder.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Builder;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    internal class DefaultEndpointDataSourceBuilder : EndpointDataSourceBuilder
+    {
+        public IApplicationBuilder ApplicationBuilder { get; set; }
```
rynowak: I think we should follow a similar approach to what we did in MVC to get the app builder... OR.... use https://github.com/aspnet/Hosting/blob/master/src/Microsoft.AspNetCore.Hosting/Builder/ApplicationBuilderFactory.cs#L11 - that's what it's there for.

src/Microsoft.AspNetCore.Routing/EndpointDataSourceBuilder.cs
rynowak: or maybe `CreateApplicationBuilder()`.

src/Microsoft.AspNetCore.Routing/Matching/MatcherEndpointBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Routing.Matching
+{
+    public class MatcherEndpointBuilder : EndpointBuilder
```
rynowak: sealed?

src/Microsoft.AspNetCore.Routing/Matching/MatcherEndpointBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        private readonly Func<RequestDelegate, RequestDelegate> _invoker;
+        private readonly RoutePattern _routePattern;
+        private readonly RouteValueDictionary _requiredValues;
+        private readonly int _order;
```
rynowak: I think to take this to non-demo zone this should all be publicly mutable.

src/Microsoft.AspNetCore.Routing/Matching/MatcherEndpointBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public MatcherEndpointBuilder(
+           Func<RequestDelegate, RequestDelegate> invoker,
+           RoutePattern routePattern,
+           RouteValueDictionary requiredValues,
```
rynowak: Required values is based on older design, don't need it here.

src/Microsoft.AspNetCore.Routing/Matching/MatcherEndpointBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+            _invoker = invoker;
+            _routePattern = routePattern;
+            _requiredValues = requiredValues;
+            _order = order;
+        }
+
+        public override Endpoint Build()
+        {
+            var resolvedMetadata = new List<object>(Metadata);
+            resolvedMetadata.Add(new RouteValuesAddressMetadata(null, _requiredValues));
```
rynowak: We shouldn't put this here. `RVAM` is kinda an MVC thing. We don't want to spread the complexity if we can avoid it.

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
JamesNK: For these methods I followed `MapRouteRouteBuilderExtensions.MapRoute` which has name before template.

Why is it important to follow `MatcherEndpoint`?

samples/RoutingSample.Web/HelloExtension/HelloAppBuilderExtensions.cs
rynowak: alright dope

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
JamesNK: Yeah could do. I worry a little about disambiguating overloads that involve `params object[]` though.

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
JamesNK: I quite like what we have with UseEndpointRouting providing the router. It provides an obvious space to place any routes.

JamesNK: But something to play with in the future...

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
rynowak: Oh is this route name or display name? I'd really, really like to keep old concepts like RVA and route name out of these apis

JamesNK: It is the display name. I'll rename it.

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            return MapEndpoint(builder, invoker, name, RoutePatternFactory.Parse(template), metadata);
+        }
+
+        public static EndpointBuilder MapEndpoint(
```
JamesNK: Should this return EndpointBuilder or MatcherEndpointBuilder?

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Matching;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class MapEndpointEndpointDataSourceBuilderExtensions
+    {
+        public static EndpointBuilder MapEndpoint(
```
JamesNK: Should this return `EndpointBuilder` or `MatcherEndpointBuilder`? There are public properties on MatcherEndpointBuilder now but no one will likely use them because the base type is returned.

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
rynowak: This should be `MatcherEndpointBuilder`.

src/Microsoft.AspNetCore.Routing/DefaultEndpointDataSourceBuilder.cs
rynowak: OK, I'm fine with leaving this as-is right now. We'll sort this out when we've got more clarity.

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
rynowak: Yeah the disagreement about this is coming from folks that don't think its obvious how endpoints interact with ordering.

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+            this EndpointDataSourceBuilder builder,
+            Func<RequestDelegate, RequestDelegate> invoker,
+            string pattern,
+            string name)
```
kichalla: Should this be `displayName` as its not clear if a user might use it for a `route name`

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
```diff
@@ -26,65 +22,59 @@ public void ConfigureServices(IServiceCollection services)
...
         {
-            app.UseEndpointRouting();
+            app.UseEndpointRouting(builder =>
+            {
+                builder.MapHello("/helloworld", "World");
+
+                builder.MapEndpoint(
```
kichalla: Remove the `builder` here as we are chaining?

JamesNK: The return type is the builder for that specific endpoint.

src/Microsoft.AspNetCore.Routing/Builder/MapEndpointEndpointDataSourceBuilderExtensions.cs
JamesNK: I will leave it returning `EndpointBuilder` unless someone says otherwise 👼

rynowak: This should be `MatcherEndpointBuilder`. Sorry, I thought I commented to taht effect. Maybe I did it in the wrong spot.

PR #706, opened 2018-08-10
 Optimize RoutePattern allocations

JamesNK:
Avoid copying collections when there is already a safe non-parameter copy.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
+++ b/src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
@@ -21,10 +21,10 @@ namespace Microsoft.AspNetCore.Routing.Patterns
 
         internal RoutePattern(
             string rawText,
-            Dictionary<string, object> defaults,
-            Dictionary<string, IReadOnlyList<RoutePatternConstraintReference>> constraints,
-            RoutePatternParameterPart[] parameters,
-            RoutePatternPathSegment[] pathSegments)
+            IReadOnlyDictionary<string, object> defaults,
+            IReadOnlyDictionary<string, IReadOnlyList<RoutePatternConstraintReference>> constraints,
+            IReadOnlyList<RoutePatternParameterPart> parameters,
+            IReadOnlyList<RoutePatternPathSegment> pathSegments)
         {
             Debug.Assert(defaults != null);
             Debug.Assert(constraints != null);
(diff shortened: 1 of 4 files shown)


JamesNK:
https://github.com/aspnet/Mvc/pull/8249#issuecomment-411960843

rynowak:
Do you have a sense for the impact of this on MVC?

JamesNK:
Not yet but it is a feature branch so I'll profile MVC endpoint datasource before merging.

JamesNK:
A couple more changes to reduce copies. Results: https://github.com/aspnet/Mvc/pull/8249#issuecomment-411971139

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -357,6 +373,14 @@ RoutePatternPart VisitPart(RoutePatternPart part)
                     parameterConstraints.AddRange(parameter.Constraints);
                 }
 
+                if (Equals(parameter.Default, @default)
+                    && parameter.Constraints.Count == 0
+                    && (parameterConstraints?.Count ?? 0) == 0)
```
JamesNK: Is this ok? I choose to be conservative and force a copy if there are any constraints.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
```diff
@@ -21,10 +21,10 @@ public sealed class RoutePattern
 
         internal RoutePattern(
             string rawText,
-            Dictionary<string, object> defaults,
-            Dictionary<string, IReadOnlyList<RoutePatternConstraintReference>> constraints,
-            RoutePatternParameterPart[] parameters,
-            RoutePatternPathSegment[] pathSegments)
+            IReadOnlyDictionary<string, object> defaults,
```
davidfowl: This probably makes the performance worse.

JamesNK: The property it is assigned to is `IReadOnlyDictionary<string, object>` so it makes no difference.

davidfowl: Ah, it was already bad. Maybe make all Dictionary

JamesNK: It's a read only dictionary because RoutePattern and friends are immutable.

The interface overhead doesn't effect routing performance because everything is loaded into runtime optimized data structures (DFA matcher state machine).

davidfowl: So this overall optimization isn't that impactful is what you're saying?

JamesNK: It's cutting allocations at startup.

Issue #677, opened 2018-08-01
EndpointFeature.Invoker - RequestDelegate or Func<RequestDelegate, RequestDelegate>?


JamesNK:
Investigate whether Invoker should be `RequestDelegate` or `Func<RequestDelegate, RequestDelegate>`

JamesNK:
Done https://github.com/aspnet/Routing/pull/712

PR #707, opened 2018-08-12
Change endpoint invoker to RequestDelegate

JamesNK:
Fixes https://github.com/aspnet/Routing/issues/677

Yay or nay? Can anyone think of a real-world use for the next delegate with an endpoint?


src/Microsoft.AspNetCore.Routing.Abstractions/IEndpointFeature.cs
```diff
@@ -22,7 +22,7 @@ public interface IEndpointFeature
         /// Gets or sets a delegate that can be used to invoke the current
         /// <see cref="Routing.Endpoint"/>.
         /// </summary>
-        Func<RequestDelegate, RequestDelegate> Invoker { get; set; }
+        RequestDelegate Invoker { get; set; }
```
davidfowl: rename to RequestDelegate

JamesNK:
Replaced by https://github.com/aspnet/Routing/pull/712

PR #708, opened 2018-08-13
Minor RVD and EndpointMetadataCollection refactors

JamesNK:
From https://github.com/aspnet/HttpAbstractions/pull/1030

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
@@ -222,13 +222,7 @@ namespace Microsoft.AspNetCore.Routing
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -606,9 +594,5 @@ private static void ValidatePropertyNames(Type type, PropertyHelper[] properties
                 }
             }
         }
-
-        private class PropertyCache : ConcurrentDictionary<Type, PropertyHelper[]>
```
JamesNK: @rynowak I looked in history and you added this type. Was there a reason?

rynowak: Just because I like private classes

PR #712, opened 2018-08-14
API refactor: MatcherEndpoint to RouteEndpoint, Endpoint namespace

JamesNK:
* Rename MatcherEndpoint to RouteEndpoint
* Move Invoker to Endpoint

~Note that this on top of https://github.com/aspnet/Routing/pull/707~


src/Microsoft.AspNetCore.Routing/RouteEndpoint.cs
```diff
@@ -26,13 +26,13 @@ public sealed class MatcherEndpoint : Endpoint
...
+        public RouteEndpoint(
             RequestDelegate invoker,
             RoutePattern routePattern,
             int order,
             EndpointMetadataCollection metadata,
             string displayName)
-            : base(metadata, displayName)
+            : base(invoker, metadata, displayName)
         {
             if (invoker == null)
```
JamesNK: Endpoint allows for a null invoker, RouteEndpoint does not.

src/Microsoft.AspNetCore.Routing/RouteEndpoint.cs
```diff
@@ -26,13 +26,13 @@ public sealed class MatcherEndpoint : Endpoint
         /// The <see cref="EndpointMetadataCollection"/> or metadata associated with the endpoint.
         /// </param>
         /// <param name="displayName">The informational display name of the endpoint.</param>
-        public MatcherEndpoint(
+        public RouteEndpoint(
             RequestDelegate invoker,
```
JamesNK: Is there any desire to change the order of RouteEndpoint ctor parameters now that invoker is used on the base ctor?

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
```diff
@@ -1,27 +1,35 @@
...
         /// </summary>
+        /// <param name="invoker">The delegate used to processes requests for the endpoint.</param>
         /// <param name="metadata">
         /// The endpoint <see cref="EndpointMetadataCollection"/>. May be null.
         /// </param>
         /// <param name="displayName">
         /// The informational display name of the endpoint. May be null.
         /// </param>
         protected Endpoint(
+            RequestDelegate invoker,
```
rynowak: I think @davidfowl wanted this renamed to `requestDelegate`. I'm kinda neutral on it, as long as you guys agree I'm happy 👍

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
```diff
@@ -36,6 +44,11 @@ public abstract class Endpoint
         /// </summary>
         public EndpointMetadataCollection Metadata { get; }
 
+        /// <summary>
+        /// Gets the delegate used to processes requests for the endpoint.
+        /// </summary>
+        public RequestDelegate Invoker { get; }
```
rynowak: Same here re: `RequestDelegate` vs `Invoker`

JamesNK:
Renamed.

Also removed `Invoker` from `IEndpointFeature` to match master.

src/Microsoft.AspNetCore.Routing/RouteEndpoint.cs
JamesNK: I don't know. Earlier you mentioned you wanted the ctor order to match the base `Endpoint`.
Currently request delegate is first and the other two parameters are last.

src/Microsoft.AspNetCore.Routing.Abstractions/Endpoint.cs
```diff
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Http;
+using System.Threading.Tasks;
```
davidfowl: nit: Wrong usings order

src/Microsoft.AspNetCore.Routing/RouteEndpoint.cs
```diff
@@ -7,56 +7,47 @@
...
     /// Represents an <see cref="Endpoint"/> that can be used in URL matching or URL generation.
     /// </summary>
-    public sealed class MatcherEndpoint : Endpoint
+    public sealed class RouteEndpoint : Endpoint
     {
-        internal static readonly Func<RequestDelegate, RequestDelegate> EmptyInvoker = (next) =>
-        {
-            return (context) => Task.CompletedTask;
-        };
+        internal static readonly RequestDelegate EmptyRequestDelegate = (context) => Task.CompletedTask;
```
davidfowl: This should be in the base

rynowak: Then it moves to the HttpAbstractions repo, and we replace it with something else.

JamesNK: It's just used by tests. I'm going to shift it to a helper class in a test project.

JamesNK:
🆙 📅 
Changed IEndpointFeature usages to EndpointFeature so as to access RouteValues.

benchmarks/Swaggatherer/Template.cs
```diff
@@ -94,7 +94,7 @@ private Matcher SetupMatcher(MatcherBuilder builder)
             return builder.Build();
         }}
 
-        private MatcherEndpoint CreateEndpoint(string template, string controllerName, string actionName, string httpMethod)
+        private RouteEndpoint CreateEndpoint(string template, string controllerName, string actionName, string httpMethod)
```
rynowak: Find + Replace for the win!

src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
```diff
@@ -77,6 +77,18 @@ public async Task Invoke(HttpContext httpContext)
             await _next(httpContext);
         }
 
+        private static void SetEndpointFeature(HttpContext httpContext, EndpointFeature feature)
+        {
+            // An IRouteValuesFeature might have already been set
+            // Copy its RouteValues collection if present
+            var currentRouteValuesFeature = httpContext.Features.Get<IRouteValuesFeature>();
+            feature.RouteValues = currentRouteValuesFeature?.RouteValues;
```
rynowak: Is there a scenario you had in mind for this? Since we're trouncing an existing endpoint feature, I think it's better that we also trounce the route values.

src/Microsoft.AspNetCore.Routing/Matching/EndpointSelector.cs
```diff
@@ -21,13 +22,12 @@ public abstract class EndpointSelector
         /// <param name="candidates">The <see cref="CandidateSet"/>.</param>
         /// <returns>A <see cref="Task"/> that completes asynchronously once endpoint selection is complete.</returns>
         /// <remarks>
-        /// An <see cref="EndpointSelector"/> should assign the <see cref="IEndpointFeature.Endpoint"/>,
-        /// <see cref="IEndpointFeature.Invoker"/>, and <see cref="IEndpointFeature.Values"/> properties
-        /// once an endpoint is selected.
+        /// An <see cref="EndpointSelector"/> should assign the <see cref="EndpointFeature.Endpoint"/>
+        /// and <see cref="EndpointFeature.RouteValues"/> properties once an endpoint is selected.
```
rynowak: 👍

src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
JamesNK: It is unlikely anyone would have registered an `IEndpointFeature`. Meanwhile `IRouteValuesFeature` could have been modified in V3 through `HttpRequest.RouteValues`.

rynowak: Right, but who would do that and what would their expectations be? I think it's really wierd to go our of our way to address an unknown (at least to me) scenario.

JamesNK: I don't know. It just feels like being a good citizen of the eco-system that if we are replacing the IRouteValuesFeature then we should preserve the underlying values, and I don't see a downside.

rynowak: The downside is that we've shipped a behavior that we have to preserve without knowing why it's valuable -- instead of doing the simplest thing. 

I have a counter proposal that I think you'll like. I'll send a PR

PR #713, opened 2018-08-14
Tokenize Less

rynowak:
This change reduces the amount of buffer space we have to stackalloc, by basing the size on the complexity of the DFA graph. This means for a techempower plaintext scenario, we need space for 2 segments, instead of 32. 

This should dramatically decrease the overhead associated with the zero-intialization of this buffer in all cases, but will have the most effect on low overhead tests like techempower plaintext.

**Before**

|           Method |       Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|----------------- |-----------:|----------:|----------:|------------:|-------:|---------:|-------:|----------:|
|         Baseline |   139.9 ns |  2.728 ns |  2.801 ns | 7,149,030.5 |   1.00 |     0.00 | 0.0005 |      40 B |
|              Dfa |   255.4 ns |  4.063 ns |  3.393 ns | 3,914,758.0 |   1.83 |     0.04 | 0.0014 |     168 B |
| LegacyTreeRouter | 1,622.2 ns | 24.161 ns | 21.418 ns |   616,453.8 |  11.60 |     0.27 | 0.0038 |     480 B |
|     LegacyRouter | 1,215.8 ns | 13.214 ns | 11.714 ns |   822,470.5 |   8.70 |     0.19 | 0.0038 |     344 B |

**After**

|           Method |       Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|----------------- |-----------:|----------:|----------:|------------:|-------:|---------:|-------:|----------:|
|         Baseline |   143.7 ns |  1.317 ns |  1.100 ns | 6,958,574.2 |   1.00 |     0.00 | 0.0002 |      40 B |
|              Dfa |   237.0 ns |  3.874 ns |  3.435 ns | 4,218,553.8 |   1.65 |     0.03 | 0.0014 |     168 B |
| LegacyTreeRouter | 1,708.0 ns | 15.888 ns | 14.862 ns |   585,475.2 |  11.89 |     0.13 | 0.0057 |     480 B |
|     LegacyRouter | 1,056.5 ns | 20.943 ns | 17.489 ns |   946,490.6 |   7.35 |     0.13 | 0.0038 |     344 B |

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/FastPathTokenizer.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/FastPathTokenizer.cs
@@ -9,16 +9,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
     // to PathTokenizer.
     internal static class FastPathTokenizer
     {
-        // The default limit for the number of segments we tokenize.
-        //
-        // Historically the limit on the number of segments routing supports is 28.
-        // RoutePrecedence computes precedence based on a decimal, which supports 28
-        // or 29 digits.
-        //
-        // So setting this limit to 32 should work pretty well. We also expect the tokenizer
-        // to be used with stackalloc, so we want a small number.
-        public const int DefaultSegmentCount = 32;
-
         // This section tokenizes the path by marking the sequence of slashes, and their
         // and the length of the text between them.
         //
--- a/benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/MatcherFindCandidateSetAzureBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/MatcherFindCandidateSetAzureBenchmark.cs
@@ -9,6 +9,10 @@ namespace Microsoft.AspNetCore.Routing.Matching
     // Generated from https://github.com/Azure/azure-rest-api-specs
     public partial class MatcherFindCandidateSetAzureBenchmark : MatcherBenchmarkBase
     {
+        // SegmentCount should be max-segments + 1, but we don't have a good way to compute
+        // it here, so using 16 as a safe guess.
+        private const int SegmentCount = 16;
+
         private const int SampleCount = 100;
 
         private BarebonesMatcher _baseline;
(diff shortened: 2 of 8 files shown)


benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/MatcheFindCandidateSetSingleEntryBenchmark.cs
```diff
@@ -9,9 +9,12 @@ namespace Microsoft.AspNetCore.Routing.Matching
 {
     public class MatcheFindCandidateSetSingleEntryBenchmark : MatcherBenchmarkBase
     {
+        // SegmentCount should be max-segments + 1
+        private const int SegmentCount = 2;
+
         private TrivialMatcher _baseline;
         private DfaMatcher _dfa;
-
+
```
JamesNK: nit

src/Microsoft.AspNetCore.Routing/Matching/FastPathTokenizer.cs
```diff
@@ -49,7 +39,7 @@ public static int Tokenize(string path, Span<PathSegment> segments)
             {
                 segments[count++] = new PathSegment(start, length);
             }
-
+
```
JamesNK: nit

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/MatcheFindCandidateSetSingleEntryBenchmark.cs
rynowak: huh. I really thought I told VS to do the automatic code cleanup then and fix this. I really dream of a day when I don't have to deal with this kind of thing anymore. Anyone have gofmt for c#?

davidfowl:
@rynowak We should have a hard max right?

rynowak:
We do. The route precedence computation limits you to 28 (sometimes 29) segments.  This has been in place since MVC5 and one guy has hit it, who was trying to see how many we allow.

davidfowl:
I was more asking because we stack alloc. It's not obvious from the code that there's a hard limit (or I missed it).

PR #714, opened 2018-08-14
Do not encode slashes in catch all parameter value

kichalla:


rynowak:
This is different from the fix that's described in the issue: https://github.com/aspnet/Home/issues/2669#issuecomment-395226030

kichalla:
> This is different from the fix that's described in the issue: aspnet/Home#2669 (comment)

Sure, my bad. Will close this PR and work on the fix suggested.

PR #717, opened 2018-08-15
Use RoutePattern in TemplateBinder

rynowak:


src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -149,36 +166,43 @@ public string BindValues(RouteValueDictionary acceptedValues)
...
                         }
                     }
-                    else if (part.IsParameter)
+                    else if (part is RoutePatternSeparatorPart separatorPart)
+                    {
+                        if (!context.Accept(separatorPart.Content))
+                        {
+                            return null;
+                        }
+                    }
```
rynowak: Extra case is needed here because in the past separators were also literals.

PR #719, opened 2018-08-15
Introducing new **catchAll parameter to allow generating links with u…

### LINKED ISSUE #539: Consider adding a feature for catchall that roundtrips / characters
Tracking issue for https://github.com/aspnet/Home/issues/2669#issuecomment-395226030

Questions: Will new segment type be dispatching only or will it also be added to OG routing?

kichalla:
…nencoded values

[Fixes #539] Consider adding a feature for catchall that roundtrips / characters

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
+++ b/src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
@@ -44,6 +44,11 @@ namespace Microsoft.AspNetCore.Routing.Internal
 
         public bool Accept(string value)
         {
+            return Accept(value, encodeSlashes: true);
+        }
+
+        public bool Accept(string value, bool encodeSlashes)
+        {
             if (string.IsNullOrEmpty(value))
             {
                 if (UriState == SegmentState.Inside || BufferState == SegmentState.Inside)
@@ -67,7 +72,7 @@ namespace Microsoft.AspNetCore.Routing.Internal
             {
                 if (_buffer[i].RequiresEncoding)
                 {
-                    _urlEncoder.Encode(Writer, _buffer[i].Value);
+                    EncodeValue(_buffer[i].Value);
                 }
                 else
                 {
@@ -88,16 +93,17 @@ namespace Microsoft.AspNetCore.Routing.Internal
             UriState = SegmentState.Inside;
 
             _lastValueOffset = _uri.Length;
+
             // Allow the first segment to have a leading slash.
             // This prevents the leading slash from PathString segments from being encoded.
             if (_uri.Length == 0 && value.Length > 0 && value[0] == '/')
             {
                 _uri.Append("/");
-                _urlEncoder.Encode(Writer, value, 1, value.Length - 1);
+                EncodeValue(value, 1, value.Length - 1, encodeSlashes);
             }
             else
             {
-                _urlEncoder.Encode(Writer, value);
+                EncodeValue(value, encodeSlashes);
             }
 
             return true;
@@ -195,6 +201,44 @@ namespace Microsoft.AspNetCore.Routing.Internal
             }
 
             return _uri.ToString();
+        }
+
+        private void EncodeValue(string value)
+        {
+            EncodeValue(value, encodeSlashes: true);
+        }
+
+        private void EncodeValue(string value, bool encodeSlashes)
+        {
+            EncodeValue(value, start: 0, characterCount: value.Length, encodeSlashes);
+        }
+
+        // For testing
+        internal void EncodeValue(string value, int start, int characterCount, bool encodeSlashes)
+        {
+            // Just encode everything if its ok to encode slashes
+            if (encodeSlashes)
+            {
+                _urlEncoder.Encode(Writer, value, start, characterCount);
+            }
+            else
+            {
+                int end;
+                int length = start + characterCount;
+                while ((end = value.IndexOf('/', start, characterCount)) >= 0)
+                {
+                    _urlEncoder.Encode(Writer, value, start, end - start);
+                    _uri.Append("/");
+
+                    start = end + 1;
+                    characterCount = length - start;
+                }
+
+                if (end < 0 && characterCount >= 0)
+                {
+                    _urlEncoder.Encode(Writer, value, start, length - start);
+                }
+            }
         }
 
         private string DebuggerToString()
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
```diff
@@ -27,9 +27,17 @@ public static RoutePatternParameterPart ParseRouteParameter(string parameter)
 
             var startIndex = 0;
             var endIndex = parameter.Length - 1;
+            var encodeValue = true;
 
             var parameterKind = RoutePatternParameterKind.Standard;
-            if (parameter[0] == '*')
+
+            if (parameter.StartsWith("**"))
```
JamesNK: Use overload with StringComparison.Ordinal

JamesNK (approved):
Looks good to me but wait for Ryan to take a look before merging.

FYI I created https://github.com/aspnet/Routing/issues/720 to track updating docs

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
```diff
@@ -88,16 +93,24 @@ public bool Accept(string value)
...
+                }
+                else
+                {
+                    _urlEncoder.Encode(Writer, value);
+                }
             }
             else
             {
-                _urlEncoder.Encode(Writer, value);
+                _uri.Append(value);
```
rynowak: We still need to encode characters other than `/`

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternParameterPart.cs
```diff
@@ -27,13 +37,19 @@ public sealed class RoutePatternParameterPart : RoutePatternPart
...
 
         /// <summary>
         /// Gets the list of constraints associated with this parameter.
         /// </summary>
         public IReadOnlyList<RoutePatternConstraintReference> Constraints { get; }
 
+        /// <summary>
+        /// Gets the value indicating if current parameter's value should be encoded.
+        /// </summary>
+        public bool EncodeValue { get; }
```
rynowak: Should we call this something like `EncodeSlashes`?

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -98,6 +98,40 @@ public void GetLink_MultipleEndpoints_Success2()
...
+            var linkGenerator = CreateLinkGenerator(endpoint);
+            var httpContext = CreateHttpContext(ambientValues: new { controller = "Home", action = "Index" });
+
+            // Act
+            var link = linkGenerator.GetLink(httpContext, new { path = routeValue });
+
+            // Assert
+            Assert.Equal(expected, link);
+        }
+
```
rynowak: Can you add some cases where there are characters other than `/` that need to be url-encoded?

kichalla (commented):
🆙 📅

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
JamesNK: I'm going to guess there is no built in way to exclude encoding character and we don't want to write our own encode method or Barry will smite us.

Split value by `/`, encode each part to the writer, separated by `/`?

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
```diff
@@ -197,6 +203,38 @@ public override string ToString()
...
+            EncodeValue(value, startIndex: 0, characterCount: value.Length, encodeSlashes);
+        }
+
+        private void EncodeValue(string value, int startIndex, int characterCount, bool encodeSlashes)
+        {
+            // Just encode everything if its ok to encode slashes
+            if (!encodeSlashes && value.IndexOf('/', startIndex, characterCount) > 0)
+            {
+                var subValue = value.Substring(startIndex, characterCount);
+                var parts = subValue.Split('/');
```
JamesNK: We can avoid allocating with the `Substring` and `Split` with a loop using `value.IndexOf('/', lastSlashIndex, remainingCharacterCount)` to find the next slash and then writing that set of characters using `_urlEncoder.Encode(Writer, value, lastSlashIndex + 1, nextSlashIndex);` + a find call to Encode with the trailing chars.

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
JamesNK: Some extra tests:
* multiple slashes in a row, e.g. localhost/test//blah
* catchall ends with a trailing slash, e.g. localhost/text/thisiscatchallcontent/

kichalla: yeah, i have those scenarios covered here actually

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
kichalla: Good idea 👍

kichalla (commented):
🆙 📅

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
```diff
@@ -197,6 +203,44 @@ public override string ToString()
...
+            // Just encode everything if its ok to encode slashes
+            if (encodeSlashes)
+            {
+                _urlEncoder.Encode(Writer, value, startIndex, characterCount);
+            }
+            else
+            {
+                var length = startIndex + characterCount;
+                var lastStartIndex = startIndex;
+                for (var i = startIndex; i < length; i++)
```
rynowak: nit: I would expect this loop to look more like: https://github.com/aspnet/Routing/blob/release/2.2/src/Microsoft.AspNetCore.Routing/Matching/FastPathTokenizer.cs#L29

# CHANNEL YOUR INNER C PROGRAMMER

src/Microsoft.AspNetCore.Routing/Patterns/RouteParameterParser.cs
```diff
@@ -27,9 +27,17 @@ public static RoutePatternParameterPart ParseRouteParameter(string parameter)
...
             var startIndex = 0;
             var endIndex = parameter.Length - 1;
+            var encodeValue = true;
 
             var parameterKind = RoutePatternParameterKind.Standard;
-            if (parameter[0] == '*')
+
+            if (parameter.StartsWith("**", StringComparison.Ordinal))
+            {
+                encodeValue = false;
```
rynowak: rename to `encodeSlashes` to match

rynowak (approved):

JamesNK (approved):
:shipit:

PR #723, opened 2018-08-16
Performance for EndpointMetadataCollection

rynowak:
#TLDR

The baseline for these tests is the naive implementation (looping with an as-cast). Given that we typically are working with a generic type parameter and that type will be an interface, this is one of the worst cases for casts.

Using a dictionary-based cache is a better option at a really low threshold (4-5 items). I don't feel like it's worth optimizing for the cases where the item collection is really really small, but we could do that in the future if it proves necessary. You'll see that the perf of the dictionary-based implementation is still pretty close for N=3

# Results

|             Method | Count |      Mean |     Error |    StdDev |    Median |         Op/s | Scaled | ScaledSD | Allocated |
|------------------- |------ |----------:|----------:|----------:|----------:|-------------:|-------:|---------:|----------:|
|           **Baseline** |     **3** |  **27.93 ns** | **0.2281 ns** | **0.2133 ns** |  **27.85 ns** | **35,799,096.6** |   **1.00** |     **0.00** |       **0 B** |
|        GetMetadata |     3 |  36.63 ns | 0.7390 ns | 1.5262 ns |  35.66 ns | 27,302,140.5 |   1.31 |     0.05 |       0 B |
| GetOrderedMetadata |     3 |  57.53 ns | 0.0701 ns | 0.0621 ns |  57.52 ns | 17,383,720.3 |   2.06 |     0.02 |       0 B |
|                    |       |           |           |           |           |              |        |          |           |
|           **Baseline** |    **10** |  **85.18 ns** | **2.0738 ns** | **1.7317 ns** |  **84.56 ns** | **11,740,338.5** |   **1.00** |     **0.00** |       **0 B** |
|        GetMetadata |    10 |  36.69 ns | 0.8027 ns | 1.7279 ns |  35.81 ns | 27,257,759.2 |   0.43 |     0.02 |       0 B |
| GetOrderedMetadata |    10 |  58.89 ns | 1.1702 ns | 2.4168 ns |  57.47 ns | 16,980,163.1 |   0.69 |     0.03 |       0 B |
|                    |       |           |           |           |           |              |        |          |           |
|           **Baseline** |    **25** | **208.92 ns** | **0.6678 ns** | **0.5214 ns** | **208.77 ns** |  **4,786,459.7** |   **1.00** |     **0.00** |       **0 B** |
|        GetMetadata |    25 |  36.52 ns | 0.6842 ns | 0.5713 ns |  36.31 ns | 27,379,923.1 |   0.17 |     0.00 |       0 B |
| GetOrderedMetadata |    25 |  59.58 ns | 1.1892 ns | 2.5852 ns |  57.90 ns | 16,784,812.3 |   0.29 |     0.01 |       0 B |

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.Routing.Performance/EndpointMetadataCollectionBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.Routing.Performance/EndpointMetadataCollectionBenchmark.cs
@@ -1,0 +1,127 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using BenchmarkDotNet.Attributes;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class EndpointMetadataCollectionBenchmark
+    {
+        private object[] _items;
+        private EndpointMetadataCollection _collection;
+
+        [Params(3, 10, 25)]
+        public int Count { get; set; }
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            var seeds = new Type[]
+            {
+                typeof(Metadata1),
+                typeof(Metadata2),
+                typeof(Metadata3),
+                typeof(Metadata4),
+                typeof(Metadata5),
+                typeof(Metadata6),
+                typeof(Metadata7),
+                typeof(Metadata8),
+                typeof(Metadata9),
+            };
+
+            _items = new object[Count];
+            for (var i = 0; i < _items.Length; i++)
+            {
+                _items[i] = seeds[i % seeds.Length];
+            }
+
+            _collection = new EndpointMetadataCollection(_items);
+        }
+
+        // This is a synthetic baseline that visits each item and does an as-cast.
+        [Benchmark(Baseline = true, OperationsPerInvoke = 5)]
+        public void Baseline()
+        {
+            var items = _items;
+            for (var i = items.Length - 1; i >= 0; i--)
+            {
+                GC.KeepAlive(_items[i] as IMetadata1);
+            }
+
+            for (var i = items.Length - 1; i >= 0; i--)
+            {
+                GC.KeepAlive(_items[i] as IMetadata2);
+            }
+
+            for (var i = items.Length - 1; i >= 0; i--)
+            {
+                GC.KeepAlive(_items[i] as IMetadata3);
+            }
+
+            for (var i = items.Length - 1; i >= 0; i--)
+            {
+                GC.KeepAlive(_items[i] as IMetadata4);
+            }
+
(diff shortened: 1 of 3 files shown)


benchmarks/Microsoft.AspNetCore.Routing.Performance/EndpointMetadataCollectionBenchmark.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        // This is a synthetic baseline that visits each item and does an as-cast.
+        [Benchmark(Baseline = true, OperationsPerInvoke = 5)]
+        public void Baseline()
+        {
+            var items = _items;
+            for (var i = items.Length - 1; i >= 0; i--)
+            {
+                GC.KeepAlive(_items[i] as IMetadata1);
```
JamesNK: What does KeepAlive do?

src/Microsoft.AspNetCore.Routing.Abstractions/EndpointMetadataCollection.cs
```diff
@@ -37,6 +41,7 @@ public EndpointMetadataCollection(IEnumerable<object> items)
             }
 
             _items = items.ToArray();
+            _cache = new ConcurrentDictionary<Type, object[]>();
```
JamesNK: Have you considered populating a normal dictionary in the ctor instead of being lazy?

Edit:
As I typed this I realized we need to handle `GetMetadata<IInterface>()` which means reflection to populate the dictionary with all the different base types and interfaces of the values.

JamesNK: I see your perf test will test this scenario but have we got any unit tests for it?

benchmarks/Microsoft.AspNetCore.Routing.Performance/EndpointMetadataCollectionBenchmark.cs
rynowak: KeepAlive prevents the use of the argument from being optimized away. Basically it's a way to force the expression to be evaluated pretty much as written, without being deleted by the JIT.

src/Microsoft.AspNetCore.Routing.Abstractions/EndpointMetadataCollection.cs
rynowak: Yeah, that was exactly my rationale as well, we don't know up front what `TMetadata` will be, so unless we wanted to use all possible types we couldn't do a good job.

rynowak: Yeah, so... oddly enough there are like no tests for EMC. I will remedy that.

PR #725, opened 2018-08-16
Fix #710 - add support for data tokens

### LINKED ISSUE #710: Add support for MVC + Endpoint Routing + DataTokens
We need to have some path to support data tokens.

rynowak:


src/Microsoft.AspNetCore.Routing/EndpointFeature.cs
```diff
@@ -54,7 +54,20 @@ RouteData IRoutingFeature.RouteData
             {
                 if (_routeData == null)
                 {
-                    _routeData = new RouteData(_values);
+                    _routeData = _values == null ? new RouteData() : new RouteData(_values);
+
+                    // Note: DataTokens won't update if someone else overwrites the Endpoint
+                    // after route values has been set. This seems find since endpoints are a new
+                    // feature and DataTokens are for back-compat.
```
rynowak: I didn't have a better idea than this. RouteValues is really commonly read/write. Data tokens isn't frequently written to in application code.

src/Microsoft.AspNetCore.Routing/EndpointFeature.cs
```diff
@@ -54,7 +54,20 @@ RouteData IRoutingFeature.RouteData
             {
                 if (_routeData == null)
                 {
-                    _routeData = new RouteData(_values);
+                    _routeData = _values == null ? new RouteData() : new RouteData(_values);
```
rynowak: Hit this argument null in unit tests, seemed worth fixing.

JamesNK: What is this change for?

test/Microsoft.AspNetCore.Routing.Tests/EndpointFeatureTest.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        [Fact]
+        public void RouteData_CanIntializeDataTokens_WithMetadata()
+        {
+            // Arrange
+            var expected = new RouteValueDictionary(new { foo = 17, bar = "hello", });
+
+            var feature = new EndpointFeature()
+            {
+                Endpoint = new MatcherEndpoint(
```
JamesNK: One day we will merge RouteEndpoint. One day...

rynowak:
DataTokens are like metadata but for conventional routing. You can put a bag-O-stuff on a route, and then figure out later which route matched. Think of them like route values, except static, and not used for action selection.

PR #729, opened 2018-08-17
Add ASCII optimized jump tables

rynowak:
We can do significantly better than the built in methods we have available to call. Partially because the span version of equals creates a copy to try and vectorize, and because we can know ahead of time if we're looking at ASCII for one of the comparands. 

# Jump Table (Single Entry)

|   Method |     Mean |     Error |    StdDev |          Op/s | Scaled | Allocated |
|--------- |---------:|----------:|----------:|--------------:|-------:|----------:|
| Baseline | 5.210 ns | 0.0330 ns | 0.0293 ns | 191,928,943.2 |   1.00 |       0 B |
|  Default | 6.039 ns | 0.0514 ns | 0.0456 ns | 165,595,507.2 |   1.16 |       0 B |
|    Ascii | 3.711 ns | 0.0506 ns | 0.0448 ns | 269,497,429.6 |   0.71 |       0 B |

See, much better. 

# Plaintext

**Before**

|           Method |      Mean |      Error |     StdDev |         Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|----------------- |----------:|-----------:|-----------:|-------------:|-------:|---------:|-------:|----------:|
|         Baseline |  85.70 ns |  0.4858 ns |  0.4056 ns | 11,667,946.8 |   1.00 |     0.00 | 0.0004 |      40 B |
|              Dfa | 141.44 ns |  1.7346 ns |  1.5377 ns |  7,070,193.3 |   1.65 |     0.02 | 0.0017 |     168 B |
| LegacyTreeRouter | 834.57 ns | 16.6036 ns | 15.5310 ns |  1,198,221.7 |   9.74 |     0.18 | 0.0048 |     480 B |
|     LegacyRouter | 617.02 ns | 12.3453 ns | 23.7852 ns |  1,620,702.9 |   7.20 |     0.28 | 0.0029 |     344 B |

**After**

|           Method |      Mean |     Error |    StdDev |         Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|----------------- |----------:|----------:|----------:|-------------:|-------:|---------:|-------:|----------:|
|         Baseline |  84.93 ns | 0.3640 ns | 0.3404 ns | 11,774,231.4 |   1.00 |     0.00 | 0.0004 |      40 B |
|              Dfa | 131.67 ns | 2.2844 ns | 2.1368 ns |  7,595,008.6 |   1.55 |     0.03 | 0.0019 |     168 B |
| LegacyTreeRouter | 875.63 ns | 6.1450 ns | 5.1313 ns |  1,142,036.8 |  10.31 |     0.07 | 0.0048 |     480 B |
|     LegacyRouter | 566.53 ns | 3.3557 ns | 2.9748 ns |  1,765,147.0 |   6.67 |     0.04 | 0.0038 |     344 B |

# Github

**Before**

|   Method |     Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|----------:|----------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 136.2 ns | 0.6978 ns | 0.6185 ns | 7,341,471.1 |   1.00 |     0.00 | 0.0610 |   9.49 KB |
|      Dfa | 489.9 ns | 3.7682 ns | 3.1467 ns | 2,041,165.7 |   3.60 |     0.03 | 0.9766 |  87.58 KB |

**After**

|   Method |     Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|----------:|----------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 134.9 ns | 0.7136 ns | 0.6675 ns | 7,410,382.3 |   1.00 |     0.00 | 0.0610 |   9.49 KB |
|      Dfa | 472.4 ns | 3.0402 ns | 2.3736 ns | 2,116,639.6 |   3.50 |     0.02 | 0.9766 |  87.58 KB |

# Azure

**Before**

|   Method |       Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |-----------:|---------:|---------:|------------:|-------:|---------:|-------:|----------:|
| Baseline |   302.9 ns | 3.702 ns | 3.281 ns | 3,301,570.8 |   1.00 |     0.00 |      - |   3.91 KB |
|      Dfa | 1,089.0 ns | 4.192 ns | 3.500 ns |   918,287.3 |   3.60 |     0.04 | 0.7324 |   62.7 KB |

**After**

|   Method |       Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |-----------:|---------:|---------:|------------:|-------:|---------:|-------:|----------:|
| Baseline |   323.0 ns | 4.096 ns | 3.631 ns | 3,095,719.9 |   1.00 |     0.00 |      - |   3.91 KB |
|      Dfa | 1,120.5 ns | 6.516 ns | 5.441 ns |   892,441.3 |   3.47 |     0.04 | 0.7324 |   62.7 KB |

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
@@ -1,0 +1,75 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.AspNetCore.Routing.Matching
+{
+    internal static class Ascii
+    {
+        // case-sensitive equality comparison when we KNOW that 'a' is in the ASCII range
+        // and we know that the spans are the same length.
+        //
+        // Similar to https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Globalization/CompareInfo.cs#L549
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool AsciiIgnoreCaseEquals(ReadOnlySpan<char> a, ReadOnlySpan<char> b, int length)
+        {
+            // The caller should have checked the length. We enforce that here by THROWING if the
+            // lengths are unequal.
+            if (a.Length < length || b.Length < length)
+            {
+                // This should never happen, but we don't want to have undefined
+                // behavior if it does.
+                ThrowArgumentExceptionForLength();
+            }
+            
+            ref var charA = ref MemoryMarshal.GetReference(a);
+            ref var charB = ref MemoryMarshal.GetReference(b);
+
+            // Iterates each span for the provided length and compares each character
+            // case-insensitively. This looks funky because we're using unsafe operations
+            // to elide bounds-checks.
+            while (length > 0 && AsciiIgnoreCaseEquals(charA, charB))
+            {
+                charA = ref Unsafe.Add(ref charA, 1);
+                charB = ref Unsafe.Add(ref charB, 1);
+                length--;
+            }
+
+            return length == 0;
+        }
+
+        // case-insensitive equality comparison for characters in the ASCII range
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool AsciiIgnoreCaseEquals(char charA, char charB)
+        {
+            const uint AsciiToLower = 0x20;
+            return
+                // Equal when chars are exactly equal
+                charA == charB ||
+
+                // Equal when converted to-lower AND they are letters
+                ((charA | AsciiToLower) == (charB | AsciiToLower) && (uint)((charA | AsciiToLower) - 'a') <= (uint)('z' - 'a'));
+        }
+
+        public static bool IsAscii(string text)
+        {
+            for (var i = 0; i < text.Length; i++)
+            {
+                if (text[i] > (char)0x7F)
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private static void ThrowArgumentExceptionForLength()
+        {
+            throw new ArgumentException("length");
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
@@ -39,7 +39,15 @@ namespace Microsoft.AspNetCore.Routing.Matching
                 return new ZeroEntryJumpTable(defaultDestination, exitDestination);
             }
 
-            // The IL Emit jump table is not faster for a single entry
+            // The IL Emit jump table is not faster for a single entry - but we have an optimized version when all text
+            // is ASCII
+            if (pathEntries.Length == 1 && Ascii.IsAscii(pathEntries[0].text))
+            {
+                var entry = pathEntries[0];
+                return new SingleEntryAsciiJumpTable(defaultDestination, exitDestination, entry.text, entry.destination);
+            }
+
+            // We have a fallback that works for non-ASCII
             if (pathEntries.Length == 1)
             {
                 var entry = pathEntries[0];
--- a/benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableSingleEntryBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableSingleEntryBenchmark.cs
@@ -10,10 +10,10 @@ namespace Microsoft.AspNetCore.Routing.Matching
 {
     public class JumpTableSingleEntryBenchmark
     {
-        private JumpTable _implementation;
-        private JumpTable _prototype;
+        private JumpTable _default;
         private JumpTable _trie;
         private JumpTable _vectorTrie;
+        private JumpTable _ascii;
 
         private string[] _strings;
         private PathSegment[] _segments;
@@ -21,10 +21,10 @@ namespace Microsoft.AspNetCore.Routing.Matching
         [GlobalSetup]
         public void Setup()
         {
-            _implementation = new SingleEntryJumpTable(0, -1, "hello-world", 1);
-            _prototype = new SingleEntryAsciiVectorizedJumpTable(0, -2, "hello-world", 1);
-            _trie = new ILEmitTrieJumpTable(0, -1, new[] { ("hello-world", 1), }, vectorize: false, _implementation);
-            _vectorTrie = new ILEmitTrieJumpTable(0, -1, new[] { ("hello-world", 1), }, vectorize: true, _implementation);
+            _default = new SingleEntryJumpTable(0, -1, "hello-world", 1);
+            _trie = new ILEmitTrieJumpTable(0, -1, new[] { ("hello-world", 1), }, vectorize: false, _default);
+            _vectorTrie = new ILEmitTrieJumpTable(0, -1, new[] { ("hello-world", 1), }, vectorize: true, _default);
+            _ascii = new SingleEntryAsciiJumpTable(0, -1, "hello-world", 1);
 
             _strings = new string[]
             {
@@ -80,7 +80,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
         }
 
         [Benchmark(OperationsPerInvoke = 5)]
-        public int Implementation()
+        public int Default()
         {
             var strings = _strings;
             var segments = _segments;
@@ -88,14 +88,14 @@ namespace Microsoft.AspNetCore.Routing.Matching
             var destination = 0;
             for (var i = 0; i < strings.Length; i++)
             {
-                destination = _implementation.GetDestination(strings[i], segments[i]);
+                destination = _default.GetDestination(strings[i], segments[i]);
             }
 
             return destination;
         }
 
         [Benchmark(OperationsPerInvoke = 5)]
-        public int Prototype()
+        public int Ascii()
         {
             var strings = _strings;
             var segments = _segments;
@@ -103,7 +103,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
             var destination = 0;
             for (var i = 0; i < strings.Length; i++)
             {
-                destination = _prototype.GetDestination(strings[i], segments[i]);
+                destination = _ascii.GetDestination(strings[i], segments[i]);
             }
 
             return destination;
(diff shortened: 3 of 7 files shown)


src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
```diff
@@ -0,0 +1,45 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool AsciiIgnoreCaseEqual(char charA, char charB)
+        {
+            const uint AsciiToLower = 0x20;
+            return
+                // Equal when chars are exactly equal
+                charA == charB ||
+
+                // Equal when converted to-lower AND they are letters
+                ((charA | AsciiToLower) == (charB | AsciiToLower) && (uint)((charA | AsciiToLower) - 'a') <= (uint)('z' - 'a'));
```
JamesNK: What happens with non-alpha characeters? What will lower-casing a digit or `/`, `~` let it compare as equal to something else incorrectly?

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
```diff
@@ -0,0 +1,45 @@
...
+        // and we know that the spans are the same length.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool UnsafeAsciiIgnoreCaseEqual(ReadOnlySpan<char> a, ReadOnlySpan<char> b, int length)
+        {
+            // We're doing unsafe operations here, caller is expected to have already checked
+            // length.
+            ref var charA = ref MemoryMarshal.GetReference(a);
+            ref var charB = ref MemoryMarshal.GetReference(b);
+
+            while (length > 0 && AsciiIgnoreCaseEqual(charA, charB))
```
JamesNK: I think it is worth decribing this in a comment. It is simple what it does but a lot of people won't be fimilar with the method of doing it

e.g. interate through characters and do a case insensitive compare until the end of the text

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
JamesNK: I just saw the comment that you check they are letters 😄

JamesNK: How about removing the check for letters, writing some tests that fail with non-alpha characters, then add the check back. Make sure we have that situation covered test-wise 😸

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace Microsoft.AspNetCore.Routing.Matching
+{
+    internal static class Ascii
+    {
+        // case-sensitive equality comparison when we KNOW that 'a' is in the ASCII range
+        // and we know that the spans are the same length.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool UnsafeAsciiIgnoreCaseEqual(ReadOnlySpan<char> a, ReadOnlySpan<char> b, int length)
+        {
+            // We're doing unsafe operations here, caller is expected to have already checked
```
blowdart: I always worry when I see the caller is "expected to have already checked", are we expecting anything outside of code you write in routing to end up calling this? And is the expectation correct everywhere right now?

JamesNK:
@rynowak Have you considered an IL emit jump table? Something that avoids loops altogether. Or do you think the jump tables optimized for just a few exits makes this unnessessary?

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
JamesNK: What happens if `ref Unsafe.Add(ref charA, 1);` reads beyond the end of charA? An exception? Reading random memory?

rynowak:
> Have you considered an IL emit jump table?

What would the generated code look like? Something like https://github.com/aspnet/KestrelHttpServer/blob/1f3524787e62db9890547b33e8a450441b14810b/src/Kestrel.Core/Internal/Http/HttpHeaders.Generated.cs#L3592 ?

I think it's something we could do, but we'd have to be concerned about worst-case performance, which would be a little hard to test for. In the case of Kestrel, the headers are defined by the spec, but a user could define a set of routes that cause us to do a linear search. It wouldn't be **HORRID** but the advantage of using some kind of hash table is that the performance much less likely to degrade to linear search territory.

Another option would be a trie, but I'm a little worried that the generated code won't be faster, it's going to be **REAL** branchy. The branch predictor would hate us, but it would be O(N). 

--

I'm sure that whatever we do, we'll adopt plain-old-.NET-dictionary as the strategy of last resort. I think the widest node I've seen in testing is 92 cases (Azure). I'd like us to have some upper bound on the number of candidates of anything we build, maybe 100 so that we don't have to harden it against handling a super super super big number of cases.

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
rynowak: I think the test you're actually looking for here is a case like ` ` vs `@` these are 32 bits apart on the ASCII table so if something was wrong with line 42 we might accept them as equal when they are obviously not.

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
rynowak: @blowdart - this is new code that's called in two specific places in routing. 

https://github.com/aspnet/Routing/pull/729/files#diff-fd3f742fd5ae89148383cc0f192ce283R43 - checked at lines: 38 & 42

https://github.com/aspnet/Routing/pull/729/files#diff-c9b9e0a59a91deab3cd4517572d902eeR51 - checked at lines: 41 & 48

We may end up using these more in routing, but always within the same family of internal string processing classes.

Would you feel better if I just inlined this code so it doesn't leak? 

--- 

@JamesNK - **UNDEFINED BEHAVIOUR** 😆

JamesNK:
> What would the generated code look like?

I imagine it would unroll this loop. There would be no need to look up text and destination from entries, those values would be baked into the generated IL.

```
            var entries = _entries;
            for (var i = 0; i < entries.Length; i++)
            {
                var text = entries[i].text;
                if (length == text.Length)
                {
                    // We're relying on Span for bounds checking here, we are doing unsafe operations
                    // in the loop.
                    var a = path.AsSpan(segment.Start, length);
                    var b = text.AsSpan();
                     if (Ascii.UnsafeAsciiIgnoreCaseEqual(a, b, length))
                    {
                        return entries[i].destination;
                    }
                }
            }
```

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
JamesNK: Yes, exactly what I was thinking.

rynowak:
The loops aren't really the major cost, it's the actual string comparisons (and bounds checking of each string). 

For comparison:

|            Method | Count |       Mean |     Error |    StdDev |     Median |          Op/s | Scaled | ScaledSD | Allocated |
|------------------ |------ |-----------:|----------:|----------:|-----------:|--------------:|-------:|---------:|----------:|
|          **Baseline** |     **2** |   **5.130 ns** | **0.0424 ns** | **0.0396 ns** |   **5.143 ns** | **194,936,895.4** |   **1.00** |     **0.00** |       **0 B** |
|      LinearSearch |     2 |  27.937 ns | 0.2525 ns | 0.2109 ns |  27.961 ns |  35,794,192.7 |   5.45 |     0.06 |       0 B |
| LinearSearchAscii |     2 |   8.962 ns | 0.0972 ns | 0.0909 ns |   8.972 ns | 111,576,243.0 |   1.75 |     0.02 |       0 B |
|        Dictionary |     2 | 150.838 ns | 1.4671 ns | 1.3723 ns | 150.865 ns |   6,629,635.2 |  29.41 |     0.34 |       0 B |
|             Ascii |     2 |   7.272 ns | 0.0472 ns | 0.0419 ns |   7.275 ns | 137,515,068.6 |   1.42 |     0.01 |       0 B |
|  DictionaryLookup |     2 |  24.914 ns | 0.2526 ns | 0.2239 ns |  24.868 ns |  40,137,970.8 |   4.86 |     0.06 |       0 B |
|   CustomHashTable |     2 |  21.989 ns | 0.1884 ns | 0.1762 ns |  21.982 ns |  45,477,623.4 |   4.29 |     0.05 |       0 B |
|                   |       |            |           |           |            |               |        |          |           |
|          **Baseline** |     **4** |   **5.174 ns** | **0.0348 ns** | **0.0326 ns** |   **5.177 ns** | **193,271,973.9** |   **1.00** |     **0.00** |       **0 B** |
|      LinearSearch |     4 |  52.388 ns | 0.5316 ns | 0.4972 ns |  52.240 ns |  19,088,227.1 |  10.13 |     0.11 |       0 B |
| LinearSearchAscii |     4 |  14.892 ns | 0.2919 ns | 0.4629 ns |  14.706 ns |  67,148,562.4 |   2.88 |     0.09 |       0 B |
|        Dictionary |     4 | 146.894 ns | 0.7734 ns | 0.6856 ns | 146.884 ns |   6,807,641.9 |  28.39 |     0.22 |       0 B |
|             Ascii |     4 |  10.055 ns | 0.1039 ns | 0.0921 ns |  10.038 ns |  99,452,762.3 |   1.94 |     0.02 |       0 B |
|  DictionaryLookup |     4 |  26.657 ns | 0.1707 ns | 0.1597 ns |  26.663 ns |  37,513,045.5 |   5.15 |     0.04 |       0 B |
|   CustomHashTable |     4 |  17.107 ns | 0.1304 ns | 0.1220 ns |  17.126 ns |  58,454,175.6 |   3.31 |     0.03 |       0 B |
|                   |       |            |           |           |            |               |        |          |           |
|          **Baseline** |     **5** |   **5.162 ns** | **0.0547 ns** | **0.0457 ns** |   **5.168 ns** | **193,707,566.3** |   **1.00** |     **0.00** |       **0 B** |
|      LinearSearch |     5 |  61.968 ns | 0.7017 ns | 0.6564 ns |  62.017 ns |  16,137,305.2 |  12.00 |     0.16 |       0 B |
| LinearSearchAscii |     5 |  18.637 ns | 0.3540 ns | 0.3477 ns |  18.511 ns |  53,655,980.3 |   3.61 |     0.07 |       0 B |
|        Dictionary |     5 | 144.261 ns | 2.8030 ns | 3.4423 ns | 142.152 ns |   6,931,899.5 |  27.95 |     0.69 |       0 B |
|             Ascii |     5 |   8.789 ns | 0.1733 ns | 0.2486 ns |   8.669 ns | 113,782,366.8 |   1.70 |     0.05 |       0 B |
|  DictionaryLookup |     5 |  24.759 ns | 0.4708 ns | 0.4404 ns |  24.576 ns |  40,389,951.1 |   4.80 |     0.09 |       0 B |
|   CustomHashTable |     5 |  20.843 ns | 0.4092 ns | 0.4548 ns |  20.666 ns |  47,977,101.2 |   4.04 |     0.09 |       0 B |
|                   |       |            |           |           |            |               |        |          |           |
|          **Baseline** |    **10** |   **5.239 ns** | **0.1015 ns** | **0.1042 ns** |   **5.215 ns** | **190,891,256.8** |   **1.00** |     **0.00** |       **0 B** |
|      LinearSearch |    10 | 130.258 ns | 2.8923 ns | 2.7055 ns | 129.717 ns |   7,677,060.5 |  24.87 |     0.69 |       0 B |
| LinearSearchAscii |    10 |  32.324 ns | 0.5791 ns | 0.6668 ns |  32.111 ns |  30,936,432.0 |   6.17 |     0.17 |       0 B |
|        Dictionary |    10 | 146.959 ns | 1.5813 ns | 1.3204 ns | 147.028 ns |   6,804,632.1 |  28.06 |     0.59 |       0 B |
|             Ascii |    10 |  14.608 ns | 0.1066 ns | 0.0945 ns |  14.624 ns |  68,454,626.5 |   2.79 |     0.06 |       0 B |
|  DictionaryLookup |    10 |  24.081 ns | 0.4642 ns | 0.6197 ns |  24.290 ns |  41,527,232.6 |   4.60 |     0.15 |       0 B |
|   CustomHashTable |    10 |  22.323 ns | 0.3805 ns | 0.3373 ns |  22.238 ns |  44,797,635.8 |   4.26 |     0.10 |       0 B |
|                   |       |            |           |           |            |               |        |          |           |
|          **Baseline** |    **25** |   **5.487 ns** | **0.0908 ns** | **0.0805 ns** |   **5.479 ns** | **182,246,517.2** |   **1.00** |     **0.00** |       **0 B** |
|      LinearSearch |    25 | 279.137 ns | 2.5398 ns | 2.1209 ns | 278.682 ns |   3,582,467.6 |  50.88 |     0.80 |       0 B |
| LinearSearchAscii |    25 |  80.384 ns | 1.6025 ns | 3.4496 ns |  80.051 ns |  12,440,309.0 |  14.65 |     0.66 |       0 B |
|        Dictionary |    25 | 152.538 ns | 2.4089 ns | 2.2533 ns | 152.267 ns |   6,555,723.1 |  27.81 |     0.56 |       0 B |
|             Ascii |    25 |  25.143 ns | 0.4119 ns | 0.3652 ns |  25.114 ns |  39,772,249.0 |   4.58 |     0.09 |       0 B |
|  DictionaryLookup |    25 |  26.228 ns | 0.5222 ns | 0.5588 ns |  26.208 ns |  38,126,467.6 |   4.78 |     0.12 |       0 B |
|   CustomHashTable |    25 |  21.990 ns | 0.3678 ns | 0.3260 ns |  22.022 ns |  45,475,374.0 |   4.01 |     0.08 |       0 B |

--

A hash-table based data structure starts to win once we get past 5 entries. I think we'll need a different data structure than linear search.

JamesNK:
You could do that without code gen using a sparse array.
```
Entries[][] _entriesByLength;

if (_entriesByLength.Length < length)
   return exit;

Entries[] entries = _entriesByLength[length]
if (entries == null)
   return exit;

// liner search with entries
```
I think jumping all over the place with object references would make it unfriendly to L1/2/3 cache compared to IL emit.

rynowak:
That's still a linear search though and we have to do it case insensitive. Basically if we don't change the data structure or algorthm, it's still the same.

rynowak:
Updated, and made more safer

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (a.Length < length || b.Length < length)
+            {
+                // This should never happen, but we don't want to have undefined
+                // behavior if it does.
+                ThrowArgumentException();
+            }
+
+            // We're doing unsafe operations here, caller is expected to have already checked
+            // length.
+            ref char charA = ref MemoryMarshal.GetReference(a);
```
JamesNK: Curious: what was the type when it was var?

src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
```diff
@@ -45,12 +52,23 @@ public JumpTable Build()
                 return new ZeroEntryJumpTable(DefaultDestination, ExitDestination);
             }
 
+            if (_entries.Count == 1 && !HasNonAsciiEntry)
```
JamesNK: nit: Have `if (!HasNonAsciiEntry) { return SingleEntryAsciiJumpTable } else { return SingleEntryJumpTable }` nested in count check. Likewise with linear jump table.

rynowak:
> Looks good to me. Any other changes or do you consider it done?

I think I consider this about done.... but I made something in the ref-emit PR that might be better than this. SO I guess I want to get that in first and then be able to benchmark this and that side by side 😆

rynowak:
So, revisiting this, I think we should go with the approach in this PR. This is the fastest of the approaches:

|         Method |     Mean |     Error |    StdDev |   Median |          Op/s | Scaled | ScaledSD | Allocated |
|--------------- |---------:|----------:|----------:|---------:|--------------:|-------:|---------:|----------:|
|       Baseline | 7.173 ns | 0.1584 ns | 0.3764 ns | 7.003 ns | 139,405,704.9 |   1.00 |     0.00 |       0 B |
| Implementation | 7.108 ns | 0.0907 ns | 0.0758 ns | 7.082 ns | 140,684,768.1 |   0.99 |     0.05 |       0 B |
|      Prototype | 3.781 ns | 0.0274 ns | 0.0242 ns | 3.784 ns | 264,488,547.6 |   0.53 |     0.03 |       0 B |
|           Trie | 6.577 ns | 0.0927 ns | 0.0867 ns | 6.565 ns | 152,050,556.8 |   0.92 |     0.05 |       0 B |
|     VectorTrie | 6.446 ns | 0.0357 ns | 0.0316 ns | 6.439 ns | 155,146,597.0 |   0.90 |     0.04 |       0 B |
|          Ascii | 2.960 ns | 0.0319 ns | 0.0249 ns | 2.960 ns | 337,784,263.7 |   0.41 |     0.02 |       0 B |

I'm going to rebase this and resolve any outstanding feedback.

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
rynowak: it's char, changed it.

src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
rynowak: meh, I think it's clearer this way.

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
rynowak: @blowdart - to close the loop on this I've removed this comment as it's no longer true. We check the length here now.

rynowak:
@JamesNK - this is updated and ready for a final review. I removed the linear search version of this code because it would only fit into a really really obscure niche now. The IL generated version is just too good.

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
```diff
@@ -0,0 +1,75 @@
...
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private static void ThrowArgumentException()
+        {
+            throw new ArgumentException();
```
pranavkm: "length"

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
```diff
@@ -0,0 +1,75 @@
...
+                // Equal when chars are exactly equal
+                charA == charB ||
+
+                // Equal when converted to-lower AND they are letters
+                ((charA | AsciiToLower) == (charB | AsciiToLower) && (uint)((charA | AsciiToLower) - 'a') <= (uint)('z' - 'a'));
+        }
+
+        public static bool IsAscii(string text)
+        {
+            for (int i = 0; i < text.Length; i++)
```
pranavkm: `var`

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
```diff
@@ -0,0 +1,75 @@
...
+                if (text[i] > (char)0x7F)
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private static void ThrowArgumentExceptionForLength()
```
JamesNK: I'm guessing this doesn't need a NoInlining attribute on it because the throws will prevent it from being inlined?

src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
```diff
@@ -39,7 +39,15 @@ public static JumpTable Build(int defaultDestination, int exitDestination, (stri
...
+            // The IL Emit jump table is not faster for a single entry - but we have an optimized version when all text
+            // is ASCII
+            if (pathEntries.Length == 1 && Ascii.IsAscii(pathEntries[0].text))
+            {
+                var entry = pathEntries[0];
+                return new SingleEntryAsciiJumpTable(defaultDestination, exitDestination, entry.text, entry.destination);
+            }
+
+            // We have a fallback that works for non-ASCII
             if (pathEntries.Length == 1)
```
JamesNK: Could combine this if with line 44 and have the IsAscii as a nested if

src/Microsoft.AspNetCore.Routing/Matching/Ascii.cs
rynowak: Yeah, generally an exception constructor is big enough that it won't be inlined.

src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
rynowak: yeah I think it's clearer this way.

rynowak:
Thanks!

PR #730, opened 2018-08-17
I'll try ref-emit, that's a good trick!

rynowak:
Updated with some perf numbers based on the latest iteration.

@Tornhoof - I updated the `JumpTableMultipleEntryBenchmark` to use different lengths of strings 👍 

## TLDR
- The IL Emit trie is the best data structure we have matching strings with N >=2
- The overall result is about a 35% speedup in our real world benchmarks
- The vectorized version (64bit) has enough scalability to dominate the conversation for all reasonable sizes
- The vectorized version is faster for small numbers of strings
- The non-vectorized version (32bit) is better for most sizes but doesn't scale as well (N >=~50)
- The IL Emit trie is slower than our current implementation for a single string (N == 1)
- I have a prototype vectorized non-IL-Emit implementation that's much better than our current implementation, but will not be following up on it in this PR. It's possible that once I productionize it we'll end up in the same place. 

## JumpTableMultipleEntryBenchmark

|       Method | Count |       Mean |     Error |     StdDev |     Median |          Op/s | Scaled | ScaledSD | Allocated |
|------------- |------ |-----------:|----------:|-----------:|-----------:|--------------:|-------:|---------:|----------:|
|     **Baseline** |     **2** |   **9.157 ns** | **0.0119 ns** |  **0.0093 ns** |   **9.154 ns** | **109,206,844.3** |   **1.00** |     **0.00** |       **0 B** |
| LinearSearch |     2 |  24.462 ns | 0.4854 ns |  1.1536 ns |  23.687 ns |  40,879,272.5 |   2.67 |     0.13 |       0 B |
|   Dictionary |     2 | 239.305 ns | 4.7769 ns |  9.9711 ns | 233.227 ns |   4,178,763.6 |  26.13 |     1.08 |       0 B |
|         Trie |     2 |  12.490 ns | 0.1453 ns |  0.1135 ns |  12.442 ns |  80,064,498.0 |   1.36 |     0.01 |       0 B |
|   VectorTrie |     2 |  12.239 ns | 0.4438 ns |  0.3934 ns |  12.061 ns |  81,709,153.3 |   1.34 |     0.04 |       0 B |
|              |       |            |           |            |            |               |        |          |           |
|     **Baseline** |     **5** |   **9.451 ns** | **0.1893 ns** |  **0.4075 ns** |   **9.215 ns** | **105,810,490.7** |   **1.00** |     **0.00** |       **0 B** |
| LinearSearch |     5 |  59.557 ns | 0.3104 ns |  0.2424 ns |  59.485 ns |  16,790,717.5 |   6.31 |     0.26 |       0 B |
|   Dictionary |     5 | 250.408 ns | 5.4164 ns | 15.7998 ns | 249.498 ns |   3,993,477.5 |  26.54 |     1.99 |       0 B |
|         Trie |     5 |  15.329 ns | 0.0352 ns |  0.0294 ns |  15.326 ns |  65,234,792.0 |   1.62 |     0.07 |       0 B |
|   VectorTrie |     5 |  14.448 ns | 0.2894 ns |  0.6934 ns |  13.977 ns |  69,215,327.3 |   1.53 |     0.10 |       0 B |
|              |       |            |           |            |            |               |        |          |           |
|     **Baseline** |    **10** |   **9.184 ns** | **0.0564 ns** |  **0.0471 ns** |   **9.170 ns** | **108,890,713.8** |   **1.00** |     **0.00** |       **0 B** |
| LinearSearch |    10 | 109.005 ns | 0.4852 ns |  0.3788 ns | 108.837 ns |   9,173,927.8 |  11.87 |     0.07 |       0 B |
|   Dictionary |    10 | 233.432 ns | 2.2215 ns |  1.7344 ns | 232.913 ns |   4,283,900.9 |  25.42 |     0.22 |       0 B |
|         Trie |    10 |  20.698 ns | 0.0195 ns |  0.0163 ns |  20.697 ns |  48,312,758.8 |   2.25 |     0.01 |       0 B |
|   VectorTrie |    10 |  18.475 ns | 0.3657 ns |  0.8621 ns |  18.035 ns |  54,127,825.9 |   2.01 |     0.09 |       0 B |
|              |       |            |           |            |            |               |        |          |           |
|     **Baseline** |    **25** |   **9.179 ns** | **0.0447 ns** |  **0.0349 ns** |   **9.173 ns** | **108,940,288.1** |   **1.00** |     **0.00** |       **0 B** |
| LinearSearch |    25 | 253.955 ns | 1.8727 ns |  1.6601 ns | 253.249 ns |   3,937,709.1 |  27.67 |     0.20 |       0 B |
|   Dictionary |    25 | 251.812 ns | 4.9460 ns |  8.1265 ns | 248.261 ns |   3,971,222.8 |  27.43 |     0.88 |       0 B |
|         Trie |    25 | 130.515 ns | 3.1445 ns |  4.8020 ns | 128.177 ns |   7,661,961.4 |  14.22 |     0.52 |       0 B |
|   VectorTrie |    25 |  28.468 ns | 0.7217 ns |  0.7411 ns |  28.150 ns |  35,127,220.9 |   3.10 |     0.08 |       0 B |
|              |       |            |           |            |            |               |        |          |           |
|     **Baseline** |    **50** |   **9.162 ns** | **0.0149 ns** |  **0.0132 ns** |   **9.158 ns** | **109,152,276.2** |   **1.00** |     **0.00** |       **0 B** |
| LinearSearch |    50 | 422.275 ns | 8.2855 ns | 14.9405 ns | 412.976 ns |   2,368,125.0 |  46.09 |     1.61 |       0 B |
|   Dictionary |    50 | 239.422 ns | 3.3859 ns |  3.0015 ns | 238.496 ns |   4,176,717.7 |  26.13 |     0.32 |       0 B |
|         Trie |    50 | 253.350 ns | 2.1008 ns |  1.8623 ns | 252.924 ns |   3,947,102.3 |  27.65 |     0.20 |       0 B |
|   VectorTrie |    50 |  45.970 ns | 0.9156 ns |  2.2460 ns |  44.898 ns |  21,753,348.8 |   5.02 |     0.24 |       0 B |
|              |       |            |           |            |            |               |        |          |           |
|     **Baseline** |   **100** |   **9.398 ns** | **0.1872 ns** |  **0.3694 ns** |   **9.215 ns** | **106,404,380.9** |   **1.00** |     **0.00** |       **0 B** |
| LinearSearch |   100 | 574.650 ns | 0.7291 ns |  0.6089 ns | 574.604 ns |   1,740,188.6 |  61.23 |     2.25 |       0 B |
|   Dictionary |   100 | 240.242 ns | 0.9625 ns |  0.7515 ns | 239.967 ns |   4,162,464.6 |  25.60 |     0.94 |       0 B |
|         Trie |   100 | 498.034 ns | 1.9317 ns |  1.5081 ns | 497.815 ns |   2,007,893.2 |  53.07 |     1.96 |       0 B |
|   VectorTrie |   100 | 159.908 ns | 3.1971 ns |  8.2528 ns | 154.723 ns |   6,253,581.5 |  17.04 |     1.08 |       0 B |

## JumpTableSingleEntryBenchmark

|         Method |      Mean |     Error |    StdDev |    Median |          Op/s | Scaled | ScaledSD | Allocated |
|--------------- |----------:|----------:|----------:|----------:|--------------:|-------:|---------:|----------:|
|       Baseline |  8.986 ns | 0.0478 ns | 0.0373 ns |  8.976 ns | 111,281,592.4 |   1.00 |     0.00 |       0 B |
| Implementation |  8.322 ns | 0.1743 ns | 0.2865 ns |  8.179 ns | 120,158,894.2 |   0.93 |     0.03 |       0 B |
|      Prototype |  5.752 ns | 0.2388 ns | 0.2117 ns |  5.689 ns | 173,865,156.4 |   0.64 |     0.02 |       0 B |
|           Trie | 12.045 ns | 0.2485 ns | 0.5131 ns | 11.718 ns |  83,022,932.1 |   1.34 |     0.06 |       0 B |
|     VectorTrie | 11.729 ns | 0.3663 ns | 0.3762 ns | 11.621 ns |  85,261,040.1 |   1.31 |     0.04 |       0 B |

## Azure

**Before**

|   Method |       Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |-----------:|---------:|---------:|------------:|-------:|---------:|-------:|----------:|
| Baseline |   643.8 ns | 20.28 ns | 19.92 ns | 1,553,224.9 |   1.00 |     0.00 |      - |   3.91 KB |
|      Dfa | 2,203.0 ns | 41.75 ns | 46.40 ns |   453,932.3 |   3.42 |     0.12 | 0.7324 |   62.7 KB |

**After**

|   Method |       Mean |    Error |   StdDev |     Median |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |-----------:|---------:|---------:|-----------:|------------:|-------:|---------:|-------:|----------:|
| Baseline |   656.0 ns | 12.95 ns | 26.46 ns |   644.3 ns | 1,524,287.5 |   1.00 |     0.00 |      - |   3.91 KB |
|      Dfa | 1,846.8 ns | 35.37 ns | 37.85 ns | 1,839.7 ns |   541,487.6 |   2.82 |     0.12 | 0.4883 |  45.13 KB |

## Github

**Before**

|   Method |     Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|----------:|----------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 252.4 ns |  1.932 ns |  1.713 ns | 3,962,251.3 |   1.00 |     0.00 | 0.1221 |   9.49 KB |
|      Dfa | 926.7 ns | 21.847 ns | 21.457 ns | 1,079,059.1 |   3.67 |     0.09 | 0.9766 |  87.58 KB |

**After**

|   Method |     Mean |      Error |     StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|-----------:|-----------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 248.7 ns |  0.8757 ns |  0.7763 ns | 4,020,625.9 |   1.00 |     0.00 |      - |   9.49 KB |
|      Dfa | 719.5 ns | 14.2196 ns | 15.8050 ns | 1,389,924.7 |   2.89 |     0.06 | 0.7324 |  73.51 KB |


rynowak:
Putting this out for discussion... obviously not finished, but seems functionally correct.

## AND FAST

|           Method | Count |       Mean |     Error |    StdDev |          Op/s | Scaled | ScaledSD | Allocated |
|----------------- |------ |-----------:|----------:|----------:|--------------:|-------:|---------:|----------:|
|         Baseline |    25 |   5.444 ns | 0.0739 ns | 0.0655 ns | 183,674,816.5 |   1.00 |     0.00 |       0 B |
|     LinearSearch |    25 | 290.574 ns | 5.6966 ns | 6.5602 ns |   3,441,462.7 |  53.38 |     1.32 |       0 B |
|       Dictionary |    25 | 165.146 ns | 3.1412 ns | 3.2258 ns |   6,055,261.4 |  30.34 |     0.67 |       0 B |
|            Ascii |    25 |         NA |        NA |        NA |            NA |      ? |        ? |       N/A |
| DictionaryLookup |    25 |  26.721 ns | 0.4535 ns | 0.4020 ns |  37,424,116.5 |   4.91 |     0.09 |       0 B |
|  CustomHashTable |    25 |  20.253 ns | 0.3149 ns | 0.2945 ns |  49,376,385.9 |   3.72 |     0.07 |       0 B |
|             Trie |    25 |  15.233 ns | 0.3414 ns | 0.5610 ns |  65,646,955.5 |   2.80 |     0.11 |       0 B |

rynowak:
Collecting full results

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,167 @@
...
+                return _exitDestination;
+            }
+
+            return _delegate(path, segment.Start, segment.Length);
+        }
+
+        private Func<string, int, int, int> GenerateDelegate()
+        {
+#if NET462
+            var assemblyName = new AssemblyName("DynamicRoutingJumpTable");
```
JamesNK: Why a dynamic assembly over a dynamic method?

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,167 @@
...
+            il.Emit(OpCodes.Stloc, offset);
+
+            // c = path[offset]
+            il.Emit(OpCodes.Ldarg_1);
+            il.Emit(OpCodes.Ldloc, offset);
+            il.Emit(OpCodes.Callvirt, typeof(string).GetProperty("Chars", new Type[] { typeof(int), }).GetMethod);
+            il.Emit(OpCodes.Stloc, c);
+
+            var nextGroup = il.DefineLabel();
+            var groups = entries.GroupBy(e => char.ToUpperInvariant(e.text[index])).ToArray();
```
JamesNK: Going to need more comments because I'm starting to get lost here visualizing what is going on 😬

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
davidfowl: So he can debug the IL by dumping it to a file 😄

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,167 @@
...
+
+                if (char.ToLowerInvariant(group.Key) != group.Key)
+                {
+                    il.Emit(OpCodes.Ldloc, c);
+                    il.Emit(OpCodes.Ldc_I4, char.ToLowerInvariant(group.Key));
+                    il.Emit(OpCodes.Beq, nextGroup);
+                }
+
+                // Process the group
+                EmitTableForStringsOfEqualLength(il, group.ToArray(), index + 1, length, offset, c, returnDefault);
```
JamesNK: If I'm reading this correctly, this is grouping every character through to the end of comparison. At some point when there is just one match possible remaining can the generated IL be simplified?

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,167 @@
...
+
+                // Similar to 'if (c == 'A' || c == 'a') { ... }
+                il.MarkLabel(nextGroup);
+                nextGroup = il.DefineLabel();
+
+                il.Emit(OpCodes.Ldloc, c);
+                il.Emit(OpCodes.Ldc_I4, group.Key);
+                il.Emit(OpCodes.Beq, nextGroup);
+
+                if (char.ToLowerInvariant(group.Key) != group.Key)
```
JamesNK: To not pay the cost of a case insensitive comparison of the char if the does doesn't have multiple cases? Very clever 😁

JamesNK:
I think it is cool and worth exploring. Blows away my idea of just grouping by the length!

I worry about how much IL will be generated though.

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,167 @@
...
+        public RefEmitTrieJumpTable(
+            int defaultDestination,
+            int exitDestination,
+            (string text, int destination)[] entries)
+        {
+            _defaultDestination = defaultDestination;
+            _exitDestination = exitDestination;
+            _entries = entries;
+
+            _delegate = GenerateDelegate();
```
JamesNK: Need to measure what the startup cost of this would be for a large site. Might be an idea to generate the IL lazy on first call to `GetDestination`.

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
rynowak: Yeah, I don't think we actually want to cross-target for net46. I'll likely put this code in a separate tool that can dump the jump tables if you need to run peverify or something.

rynowak:
Yeah, keep in mind that this is at a 'show and tell' level of quality right now 😆 

Numbers for this are very very very good:

## Github

**Before**

|   Method |     Mean |     Error |   StdDev |   Median |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|----------:|---------:|---------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 153.1 ns |  3.006 ns | 5.719 ns | 150.2 ns | 6,530,888.5 |   1.00 |     0.00 | 0.0610 |   9.49 KB |
|      Dfa | 571.9 ns | 10.056 ns | 8.398 ns | 571.3 ns | 1,748,555.9 |   3.74 |     0.14 | 0.9766 |  87.58 KB |

**After**

|   Method |     Mean |    Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|---------:|----------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 158.5 ns | 3.005 ns |  2.811 ns | 6,310,600.4 |   1.00 |     0.00 | 0.1221 |   9.49 KB |
|      Dfa | 393.2 ns | 7.817 ns | 10.164 ns | 2,543,193.3 |   2.48 |     0.08 | 0.8545 |  73.51 KB |

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            var returnDefault = il.DefineLabel();
+
+            var nextGroup = il.DefineLabel();
+            var groups = _entries.GroupBy(e => e.text.Length).ToArray();
+            for (var i = 0; i < groups.Length; i++)
+            {
+                var group = groups[i];
+
+                // Similar to 'if (length != X) { ... }
```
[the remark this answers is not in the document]
rynowak: yeah, I think you're right about that. I'm not sure that it's worth the trouble to try and figure out if the lengths are contiguous. I worried that for practical cases the JIT would turn our switch back into an if/else ladder.

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
rynowak: Yeah sorry, this isn't really ready for a real review. I wanted to put this out there since you shared your ideas about it earlier.

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
rynowak: yes, we could absolutely do something like that. It would end up looking at lot like https://github.com/aspnet/Routing/pull/729

rynowak: I'm not sure it's worth it, both in terms of performance and effort. One idea I'd like to try is to see if this implementation is faster for techempower plaintext. Then we might end up with a single 'fast path'.

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
rynowak: It will get better - https://matthewarcus.wordpress.com/2014/02/12/tolower/

src/Microsoft.AspNetCore.Routing/Matching/RefEmitTrieJumpTable.cs
rynowak: Yeah, definitely. We'll offload this to task and use it once its available.

rynowak:
re: IL 

I'm also getting stoked on the idea of of vectorizing. This would look like reading, normalizing and comparing 4 characters at a time. This will really help with the amount of codegen, because you're doing a lot more with less.

damageboy:
@rynowak sorry if I'm jumping into the middle of the discussion, so this is basically "switching" on the length, then on str[0], str[1] etc. right? I'm a bit slow in reading IL, just thought I'd ask before asking something else..

Tornhoof:
Nice approach @rynowak.
Anyway, I tried porting my json member name matching code from SpanJson to Jumptable: 
https://gist.github.com/Tornhoof/e88bf7ea470d315372229c427efb2652 (case-sensitive). It tries to do a more optimized comparison as described here:
https://github.com/Tornhoof/SpanJson/wiki/Technology-and-Internals#3-comparison-against-integers

But it looks like your approach is faster anyway, my approach is slower (up to twice as slow according to the benchmark) , probably due to the high overhead of calling the delegate for each match (the code was designed to be called in a large loop for deserialization).
Maybe this gives you further ideas for improvements.

benaadams:
`uint` and `ulong` ascii casing https://github.com/GrabYourPitchforks/coreclr/blob/a3c1df25c4225995ffd6b18fd0fc39d6b81fd6a5/src/System.Private.CoreLib/src/System/Text/Utf16Utility.cs courtesy of @GrabYourPitchforks

rynowak:
@damageboy 

> so this is basically "switching" on the length, then on str[0], str[1] etc. right? I'm a bit slow in reading IL, just thought I'd ask before asking something else

yes, thats exactly what this does. 


@Tornhoof - that's pretty rad. @davidfowl  and I were discussing vectorizing this and it looks like you've already done that. 

> Probably due to the high overhead of calling the delegate for each match

Curious which delegate is this that's being called? It seems to me like you might have already built what I was thinking about.  

Looking more closely I think the difference between your approach and what I was imagining is that you're using a flat list of cases, where I was planning to build a 'tree' - resulting in each region of memory being read and considered in sequence. I need this to scale to a high number of entries that you wouldn't find when looking at class members. 

@benaadams cool beans 👍 https://github.com/GrabYourPitchforks/coreclr/blob/a3c1df25c4225995ffd6b18fd0fc39d6b81fd6a5/src/System.Private.CoreLib/src/System/Text/Utf16Utility.cs#L81 is exactly like what I was planning. 


---------------

I assume that @davidfowl must have tweeted this lol. I'm glad to see all of the interest. 

What I have in mind for the next iteration of this is to turn it into something like the following (psuedocode)

```
uint64 bits;
var span = path.AsSpan(segment.Start, segment.Length);
switch (span.Length)
{
    case 8:
        bits = ReadUInt64(span, 0);
        bits = ToLowerInvariantAsciiQWord(bits);
        
        if (bits == 0xSomePrecomputed64bitNumber)
        {
            bits = ReadUInt64(span, 4);
            bits = ToLowerInvariantAsciiQWord(bits);

            if (bits == 0xTheNextPrecomputed64bitNumber)
            {
                return 5; // Return the state that matches this entry
            }

            return _defaultDesintation;
        }

        if (bits == 0xSomeOtherPrecomputed64bitNumber)
        {
            ...
        }

        ....
        return _defaultDesintation;

    case 17:
     .....
}
```

rynowak:
Also for anyone who's not familiar with how we're testing this - we've built a tool that can convert swagger files to benchmarks for routing. 
https://github.com/aspnet/Routing/tree/release/2.2/benchmarks/Swaggatherer
https://github.com/aspnet/Routing/blob/release/2.2/benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/MatcherGithubBenchmark.generated.cs (warning BIG)
https://github.com/aspnet/Routing/blob/release/2.2/benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/MatcherAzureBenchmark.generated.cs (warning BIGGER)

Tornhoof:
@rynowak In my version, I generate a delegate via expression trees and that one has a certain amount of overhead and as we're talking nano seconds (atleast that's my current guess, might be a porting error on my side, I did it during my lunch break, haven't had time to look at the disassembly to find out in detail).

Yeah, the pseudo code snippet is basically what I'm doing in SpanJson, just without the logic for case-insensitive comparison (which is not necessary in my use case) and I don't have switch, then if, but the length check is part of the if.
I have a code-generated version (without expression trees), the output looks then like this:
from https://github.com/Tornhoof/SpanJson/blob/Codegen/SpanJson.Benchmarks/Generated/AnswerUtf16Formatter.cs
```csharp
var name = reader.ReadUtf16NameSpan();
var length = name.Length;
ref var b = ref MemoryMarshal.GetReference(MemoryMarshal.AsBytes(name));
if (length >= 4 && ReadUInt64(ref b, 0) == 34058901685993570UL)
{
    if (length == 13 && ReadUInt64(ref b, 8) == 32088563963986015UL && ReadUInt64(ref b, 16) == 33495998976491627UL &&
        ReadUInt16(ref b, 24) == 110)
    {
        result.body_markdown = StringUtf16Formatter<ExcludeNullsOriginalCaseResolver<char>>.Default.Deserialize(ref reader);
        continue;
    }

    if (length == 4)
    {
        result.body = StringUtf16Formatter<ExcludeNullsOriginalCaseResolver<char>>.Default.Deserialize(ref reader);
        continue;
    }

    reader.SkipNextUtf16Segment();
    continue;
}

if (length == 9 && ReadUInt64(ref b, 0) == 33496016157016161UL && ReadUInt64(ref b, 8) == 29555280583983205UL && ReadUInt16(ref b, 16) == 100)
{
    result.answer_id = NullableInt32Utf16Formatter<ExcludeNullsOriginalCaseResolver<char>>.Default.Deserialize(ref reader);
    continue;
}
```
The code in my gist is basically the same, just using expression trees and some modifications due to expressiont tree limitations, just with return result instead of the reader code.

damageboy:
@rynowak I realize this is pure kibbitzing on my behalf,
But would you be interested in seeing a similar dictionary like generator (at runtime as well) base on Expression trees?

I can dig it up from my code base, and isolate it...

It does one optimization on top of what you did, assuming I read your code properly (sorry if I missed it):
At any given "step" it will evaluate what is the best character position in terms of distribution of actual values to switch upon.

For example:
If you have 40 strings all starting with `/Get`, it is highly likely that the 5<sup>th</sup> char (for example) will have a very high distribution (certainly compared to `0:4`) so it switches on that 5<sup>th</sup> char first, then proceeds to the next most distributed index/char according to the order of distributions (descending of course...).

Anyway... great work!
Let me know if you would care to see that code, or use the optimization on top of your existing version.
I'm curious to see how this goes :)

This is really the sort of thing that makes JIT environments sometimes outperform statically compiled languages...
Great work!

benaadams:
I like it very much :)

Vectoring wouldn't allow the match to be as static; but might do something like (if was using Vectors)
```csharp
private static bool IsMatch(ReadOnlySpan<char> input, ReadOnlySpan<char> lowercaseRoutePart)
{
    if (input.Length != lowercaseRoutePart.Length)
    {
        // No match
        return false;
    }

    ref char inputStart = ref MemoryMarshal.GetReference(input);
    ref char routeStart = ref MemoryMarshal.GetReference(lowercaseRoutePart);

    if (Vector.IsHardwareAccelerated)
    {
        var A = new Vector<short>((short)'A');
        var Z = new Vector<short>((short)'Z');

        for (var i = 0; i < input.Length - Vector<short>.Count; i += Vector<short>.Count)
        {
            var vector = Unsafe.As<char, Vector<short>>(ref Unsafe.Add(ref inputStart, i));
            // Lower case
            vector = Vector.ConditionalSelect(
                Vector.BitwiseAnd(
                    Vector.GreaterThanOrEqual(vector, A),
                    Vector.LessThanOrEqual(vector, Z)
                ),
                Vector.BitwiseOr(new Vector<short>(0x20), vector), // 0010 0000
                vector
            );

            if (!vector.Equals(Unsafe.As<char, Vector<short>>(ref Unsafe.Add(ref routeStart, i))))
            {
                // No match
                return false;
            }
        }
    }

    // Non vector bit
}
```
Though @GrabYourPitchforks would probably tell me I need to worry about alignment 😉

rynowak:
@damageboy 

> At any given "step" it will evaluate what is the best character position in terms of distribution of actual values to switch upon.

If you have an example of this it would be great to see!

JamesNK:
> I assume that @davidfowl must have tweeted this lol. I'm glad to see all of the interest.

I @ @benaadams. I guess a lot of people who dive into perf follow both of us 😄

rynowak:
Every time @benaadams shows up I feel like I'm bumped into nicolas cage at the grocery store

Tornhoof:
@rynowak  Is it intended that your benchmark always has the same path length (a full guid) as the input? This kinda defeats your whole if(length ==) check:
See:
https://github.com/aspnet/Routing/blob/031e8ad6e5b95c9f8611120909ead3b75fce5523/benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableMultipleEntryBenchmark.cs#L30
and:
https://github.com/aspnet/Routing/blob/031e8ad6e5b95c9f8611120909ead3b75fce5523/benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableMultipleEntryBenchmark.cs#L184

The input to ``GetStringths`` method is always 100, making the string always the full guid length. I think you're missing a ``Random`` in that method.

rynowak:
I  coded up a version that's using uint64 to vectorize, and it's slower 😞  - not sure why yet. 

|   Method |     Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|----------:|----------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 155.9 ns | 0.7183 ns | 0.5194 ns | 6,414,677.8 |   1.00 |     0.00 | 0.0610 |   9.49 KB |
|      Dfa | 392.2 ns | 5.0353 ns | 4.7100 ns | 2,549,612.4 |   2.52 |     0.03 | 0.8545 |  73.51 KB |

rynowak:
@Tornhoof  - yeah, I think you're right. That particular benchmark needs to be revisited. The more meaningful result would be something like `GithubMatcherBenchmark` with is generated from github's public API surface.

rynowak:
For anyone curious about the latest iteration, here's what it generates: 
https://gist.github.com/rynowak/31d5e6d90e554f219e1b65634276d187

rynowak:
OK, I'm getting a much better result now with the `MatcherGithubBenchmark`. The problem was that that this route table has lots of small nodes where it was still using linear search (< 10 items). The generated ref-emit table is faster than a vanilla string comparison with even a single candidate, so I want to use it for all nodes 🔥 

Here's the new results:

|   Method |     Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|---------:|---------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 156.4 ns | 2.604 ns | 2.174 ns | 6,392,827.2 |   1.00 |     0.00 | 0.0610 |   9.49 KB |
|      Dfa | 334.0 ns | 8.785 ns | 8.628 ns | 2,993,821.5 |   2.14 |     0.06 | 0.7324 |  73.51 KB |

Compare to the original at 3.74x of the baseline. 

Here's what the profile looks like now:
image

JamesNK:
Nice!

src/Microsoft.AspNetCore.Routing/Matching/VectorizedRefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,548 @@
...
+            /// </summary>
+            public Label ReturnDefault { get; set; }
+
+            /// <summary>
+            /// Label to goto that will return a sentinel value for non-ascii text.
+            /// </summary>
+            public Label ReturnNotAscii { get; set; }
+        }
+
+        private class Methods
```
davidfowl: The funny thing about the Unsafe class is that it only exists to do things that aren't expressible in C# but can be done in IL. You can skip generating these calls and instead use the IL directly:

https://github.com/dotnet/corefx/blob/e6506dbd5669b2996b3a1479e594e6e8a3a89774/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il

JamesNK: Good point. Though, the problem with using the IL directly is you could no longer translate the generated code back to C# to debug it.

src/Microsoft.AspNetCore.Routing/Matching/VectorizedRefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,548 @@
...
+                ReturnDefault = il.DefineLabel(),
+                ReturnNotAscii = il.DefineLabel(),
+            };
+
+            // See comments on Methods for details
+            var methods = new Methods();
+
+            // Initializing top-level locals - this is similar to...
+            // ReadOnlySpan<char> span = arg0.AsSpan(arg1, arg2);
+            // ref byte p = ref Unsafe.As<char, byte>(MemoryMarshal.GetReference<char>(span))
```
Tornhoof: I imagine it's the same code internally, but imho ref var b = ref MemoryMarshal.GetReference(MemoryMarshal.AsBytes(name)); is more readable.

src/Microsoft.AspNetCore.Routing/Matching/VectorizedRefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,548 @@
...
+                    .Where(m => m.GetGenericArguments().Length == 2)
+                    .Where(m => m.GetParameters().Length == 1)
+                    .FirstOrDefault()
+                    ?.MakeGenericMethod(typeof(char), typeof(byte));
+                if (Add == null)
+                {
+                    throw new InvalidOperationException("Failed to find Unsafe.As{TFrom, TTo}(ref TFrom)");
+                }
+
+                AsSpan = typeof(MemoryExtensions).GetMethod(
```
Tornhoof: If you want to make that GetMethod code more readable, you can cheat with Expression Trees:
```csharp
Expression<Action<string>> functor = s => s.AsSpan(0,0);
var asSpanMethodInfo = (functor.Body as MethodCallExpression).Method;
```

src/Microsoft.AspNetCore.Routing/Matching/VectorizedRefEmitTrieJumpTable.cs
benaadams:
```csharp
ref MemoryMarshal.GetReference(MemoryMarshal.AsBytes(name));
```
Would first get the reference to the first element of span, `Unsafe.As` that then [create a new Span](https://github.com/dotnet/coreclr/blob/b2031a14a3b110f8e712abdb0cd392c54cd2e839/src/System.Private.CoreLib/shared/System/Runtime/InteropServices/MemoryMarshal.Fast.cs#L28) and then get the reference of the first element of that:
```csharp
public static Span<byte> AsBytes<T>(Span<T> span)
    where T : struct
{
    if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
        ThrowHelper.ThrowInvalidTypeWithPointersNotSupported(typeof(T));

    return new Span<byte>(
        ref Unsafe.As<T, byte>(ref GetReference(span)),
        checked(span.Length * Unsafe.SizeOf<T>()));
}
```
So would be 2x `GetReference`; 1x `Unsafe.As`; 1x `new Span<byte>`

Current is 1x `GetReference`; 1x `Unsafe.As`?

Tornhoof: Urgh, ok, forget that then, bad idea.

Tornhoof:
I'm not sure what the guidelines for endianess are, but I guess you probably need to handle the big-endian case, at least that lower-case bitfiddling is affected.

rynowak:
> I'm not sure what the guidelines for endianess are, but I guess you probably need to handle the big-endian case, at least that lower-case bitfiddling is affected.

My understanding is that I don't have an endianness problem because everything used here is specified in a literal with the same size as what it's being compared to. For instance, all of the uint64 literals would be converted to machine-endianness without me having to write any special case code. 

At no point does this code process individual bytes, it always looks at 64bit uints or 16bit uints.

Tornhoof:
> At no point does this code process individual bytes, it always looks at 64bit uints or 16bit uints.
You don't need to change the bitmasks? e.g.
https://github.com/aspnet/Routing/blob/7e757b7814ceab2fea63ba6678fa392aebaeda31/test/Microsoft.AspNetCore.Routing.Tests/EmitTest.cs#L60-L61

I imagined I would need to write something like ``0x8000800080008000UL`` for big endian?

benaadams:
Its over the two byte `char` which already has the effect of endianness for uppercasing, is ascii matching patterns.

e.g. `A` == `0x0041` (in code) == `64`

Not sure about string compares to long constants; depends when they were generated (e.g. as soruce code on the opposite endian platform, might be a problem?)

Tornhoof:
@benaadams Ok,  I'm kind of at a loss here, is that a yes or a no?
'a' in Little Endian is 0x0041, in Big Endian it should be 0x4100 and then the masks need to swapped too or are strings/chars handled differently somehow on big endian?

> Not sure about string compares to long constants; depends when they were generated (e.g. as soruce code on the opposite endian platform, might be a problem?)

as @rynowak  said those are in the current iteration not a problem, they are all generated at runtime and until this system is kept, it should be correct.

benaadams:
> Ok, I'm kind of at a loss here, is that a yes or a no?
'a' in Little Endian is 0x0041, in Big Endian it should be 0x4100 and then the masks need to swapped too or are strings/chars handled differently somehow on big endian?

'a' is the decimal value 97
`0x0061` in code, is also the decimal value 97

The endianness of the platform doesn't change its meaning; `0x0061` doesn't become 24832 `0x6100`; any more than 97 becomes 24832

benaadams:
> Ok, I'm kind of at a loss here, is that a yes or a no?
'a' in Little Endian is 0x0041, in Big Endian it should be 0x4100 and then the masks need to swapped too or are strings/chars handled differently somehow on big endian?

'a' is the decimal value 97
`0x0061` in code, is also the decimal value 97

The endianness of the platform doesn't change its meaning; `0x0061` doesn't become 24832 `0x6100`; any more than 97 becomes 24832

`char` is 2 bytes so already effected by endianness; so as `0x0061` switches to become `0x6100` so does `0x0080` switch to become `0x8000` as they both have to to retain the same numerical meanings.

Tornhoof:
@benaadams  Ok yeah, I was mixing serialization in there. Thanks.

rynowak:
I'll probably pick this up again this weekend and try to get a good set of tests and other things in place.

I think the main concerns left are:
1. Cleanup of how we debug the emitted code (right now I'm using hacks)
2. Make the generation of ref-emit code deferred, right now it's synchronous
3. Add proper fallback paths for non-ASCII (right now it will just fail)
4. tests 🤣

src/Microsoft.AspNetCore.Routing/Matching/VectorizedRefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,548 @@
...
+            {
+                EmitVectorizedTable(il, entries, index, length, locals, labels, methods);
+                return;
+            }
+
+            // Fall back to processing a character at a time.
+            EmitSingleCharacterTable(il, entries, index, length, locals, labels, methods);
+        }
+
+        private static void EmitVectorizedTable(
```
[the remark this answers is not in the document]
rynowak: Yeah the thing about using proper vectors is that they couldn't be constants. We're generally dealing with very small strings anyway, so this seemed nicer than opening that can of worms. And it's really fast compared to the other techniques we've tried.

src/Microsoft.AspNetCore.Routing/Matching/VectorizedRefEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,548 @@
...
+                if (ReadUnalignedUInt16 == null)
+                {
+                    throw new InvalidOperationException("Failed to find Unsafe.ReadUnaligned{T}(ref byte)");
+                }
+            }
+
+            /// <summary>
+            /// <see cref="Unsafe.Add{T}(ref T, int)"/> - Add[ref byte]
+            /// </summary>
+            public MethodInfo Add { get; set; }
```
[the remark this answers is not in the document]
rynowak: Yeah I'm converting this thing to be static + readonly properties, hold tight!

src/Microsoft.AspNetCore.Routing/Matching/VectorizedRefEmitTrieJumpTable.cs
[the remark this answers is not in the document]
rynowak: I'm going to use a cached instance

rynowak: I think the effect is the same though no? I like that we can at least show the code that we're using in terms of c#.

src/Microsoft.AspNetCore.Routing/Matching/VectorizedRefEmitTrieJumpTable.cs
rynowak: I briefly thought about it, but this is already done 😆

Directory.Build.props
```diff
@@ -18,6 +18,6 @@
     <PackageSigningCertName>MicrosoftNuGet</PackageSigningCertName>
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <LangVersion>7.2</LangVersion>
+    <LangVersion>7.3</LangVersion>
```
rynowak: ref assignment

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableMultipleEntryBenchmark.cs
```diff
@@ -67,15 +64,24 @@ public int Baseline()
...
+                {
+                    destination = string.Compare(
                         @string,
                         segment.Start,
                         @string,
                         0,
-                        @string.Length,
+                        segment.Length,
                         StringComparison.OrdinalIgnoreCase);
+                }
```
rynowak: This code should do the same thing, but it's now a bit clearer that it's correct in the context of a jump table.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableMultipleEntryBenchmark.cs
```diff
@@ -164,7 +155,7 @@ private static string[] GetStrings(int count)
                 var guid = Guid.NewGuid().ToString();
 
                 // Between 5 and 36 characters
-                var text = guid.Substring(0, Math.Max(5, Math.Min(count, 36)));
+                var text = guid.Substring(0, Math.Max(5, Math.Min(i, 36)));
```
rynowak: This was always generating the same thing length, which wasn't intended, now it generates a range of different string lengths.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableSingleEntryBenchmark.cs
```diff
@@ -40,21 +50,30 @@ public int Baseline()
             var strings = _strings;
             var segments = _segments;
 
-            var destination = 0;
-            for (var i = 0; i < strings.Length; i++)
+            int destination = 0;
+            for (int i = 0; i < strings.Length; i++)
```
rynowak: Ugh, the new VS 'code cleanup' feature did this. I'll never trust VS to do anything again.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableSingleEntryBenchmark.cs
```diff
@@ -67,12 +86,234 @@ public int Implementation()
...
+            var destination = 0;
+            for (int i = 0; i < strings.Length; i++)
+            {
+                destination = _vectorTrie.GetDestination(strings[i], segments[i]);
+            }
+
+            return destination;
+        }
+
+        private class SingleEntryAsciiVectorizedJumpTable : JumpTable
```
rynowak: So I plan to do another PR to merge this in. Probably subsuming the ASCII work. This is about half the cost of a string.Compare for a since string. I wanted to include it here for reference because this is basically the algorithm used in the generated code but that has more ifs.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableSingleEntryBenchmark.cs
```diff
@@ -67,12 +86,234 @@ public int Implementation()
...
+                string text,
+                int destination)
+            {
+                _defaultDestination = defaultDestination;
+                _exitDestination = exitDestination;
+                _text = text;
+                _destination = destination;
+
+                int length = text.Length;
+                ReadOnlySpan<char> span = text.ToLowerInvariant().AsSpan();
```
JamesNK: var

JamesNK: var all over the class

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
```diff
@@ -0,0 +1,588 @@
...
+
+namespace Microsoft.AspNetCore.Routing.Matching
+{
+    internal static class ILEmitTrieFactory
+    {
+        // The algorthm we use only works for ASCII text. If we find non-ASCII text in the input
+        // we need to reject it and let is be processed with a fallback technique.
+        public const int NotAscii = Int32.MinValue;
+
+        public static Func<string, int, int, int> Create(
```
JamesNK: Add a brief summary of what this is creating. It isn't immediately obvious from args and return type.

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
```diff
@@ -0,0 +1,588 @@
...
+            // There's no value in vectorizing the computation if we're on 32bit or
+            // if no string is long enough. We do the vectorized comparison with uint64 ulongs
+            // which isn't beneficial if they don't map to the native size of the CPU. The
+            // vectorized algorithm introduces additional overhead for casing.
+            //
+            // Vectorize by default on 64bit (allow override for testing)
+            vectorize = vectorize ?? (IntPtr.Size == 8);
+
+            // Don't vectorize if all of the strings are small (prevents allocating unused locals)
+            vectorize &= entries.Any(e => e.text.Length >= 4);
```
JamesNK: Since this doesn't effect the result, how about splitting the preceding logic out into a `ResolveIsVectorized` method and add some unit tests?

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
```diff
@@ -0,0 +1,588 @@
...
+            il.Emit(OpCodes.And);
+            il.Emit(OpCodes.Brtrue, labels.ReturnNotAscii);
+
+            // Since we're handling a single character at a time, it's easier to just
+            // generate an 'if' with two comparisons instead of doing complicated conversion
+            // logic.
+
+            // Now we generate an 'if' ladder with an entry for each of the unique
+            // characters in the group. 
+            var groups = entries.GroupBy(e => GetUInt16Key(e.text, index));
```
JamesNK: Why group by ushort instead of char? Add a comment explaining why?

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
```diff
@@ -0,0 +1,588 @@
...
+                    UInt64Value = il.DeclareLocal(typeof(ulong));
+                    UInt64LowerIndicator = il.DeclareLocal(typeof(ulong));
+                    UInt64UpperIndicator = il.DeclareLocal(typeof(ulong));
+                }
+            }
+
+            /// <summary>
+            /// Holds current character when processing a character at a time.
+            /// </summary>
+            public LocalBuilder UInt16Value { get; set; }
```
JamesNK: Do these properties need setters?

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
```diff
@@ -0,0 +1,588 @@
...
+
+            /// <summary>
+            /// Label to goto that will return a sentinel value for non-ascii text.
+            /// </summary>
+            public Label ReturnNotAscii { get; set; }
+        }
+
+        private class Methods
+        {
+            public static readonly Methods Instance = new Methods();
```
JamesNK: Add a comment explaining why there is a static instance instead of using a static class.

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            _getDestination = FallbackGetDestination;
+        }
+
+        public override int GetDestination(string path, PathSegment segment)
+        {
+            return _getDestination(path, segment);
+        }
+        
+        private int FallbackGetDestination(string path, PathSegment segment)
```
JamesNK: Add a comment summarizing what is going on here and why

e.g. want to avoid initializing IL generation until needed, don't want to block so fallback to other jump table, etc

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                InitializeILDelegate();
+            });
+        }
+
+        // Internal for testing
+        internal void InitializeILDelegate()
+        {
+            var generated = ILEmitTrieFactory.Create(_defaultDestination, _exitDestination, _entries, _vectorize);
+            _getDestination = (string path, PathSegment segment) =>
```
JamesNK: Invoking this func then the generated func isn't a big performance cost?

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,102 @@
...
+            await Task.Run(() =>
+            {
+                InitializeILDelegate();
+            });
+        }
+
+        // Internal for testing
+        internal void InitializeILDelegate()
+        {
+            var generated = ILEmitTrieFactory.Create(_defaultDestination, _exitDestination, _entries, _vectorize);
```
JamesNK: Does it at all make sense to have a reusable cache of generated functions? I was thinking that for endpoints `Home/Index` and `Login/Index` we could generate IL for `Index` once.

But since this is a trie, it isn't a simple comparison against one string but all the possible strings, so duplicates would be very rare.

src/Microsoft.AspNetCore.Routing/Microsoft.AspNetCore.Routing.csproj
```diff
@@ -4,13 +4,25 @@
...
+    <!-- 
+      RefEmit is supported in netcoreapp. 
+      
+      The ability to save compiled assemblies is for testing and debugging, not shipped in the product.
+    -->
+    <ILEmit Condition="'$(TargetFramework)'!='netstandard2.0'">true</ILEmit>
+    <ILEmitSaveAssemblies Condition="'$(ILEmitSaveAssemblies)'==''">false</ILEmitSaveAssemblies>
+    <DefineConstants Condition="'$(ILEmit)'=='true'">IL_EMIT;$(DefineConstants)</DefineConstants>
+    <DefineConstants Condition="'$(ILEmitSaveAssemblies)'=='true'">IL_EMIT_SAVE_ASSEMBLIES;$(DefineConstants)</DefineConstants>
+  </PropertyGroup>
```
JamesNK: nice 👍

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
JamesNK: And of course the exits would always be different anyway. Never mind 😸

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
```diff
@@ -0,0 +1,102 @@
...
+        
+        private int FallbackGetDestination(string path, PathSegment segment)
+        {
+            if (path.Length == 0)
+            {
+                return _exitDestination;
+            }
+
+            // We only hit this code path if the IL delegate is still initializing.
+            LazyInitializer.EnsureInitialized(ref _task, ref _initializing, ref _lock, InitializeILDelegateAsync);
```
JamesNK: What happens if two threads hit this at the same time? Could InitializeILDelegateAsync be run more than once?

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/JumpTableSingleEntryBenchmark.cs
rynowak: YEAH VS did this. I had to fix my settings....

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
rynowak: I just did it for consistency. It's easier for me to think about this uint16/uint64 vs char/ulong

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
rynowak: Adding

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
rynowak: fixed

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieFactory.cs
rynowak: I explained that this does caching. I didn't explain that particular tradeoff because it's not important. someone could change it in the future and I don't care 👍

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
rynowak: adding

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
rynowak: No, LazyInitializer is cool like that. If you use the overload that takes a lock , it will do double-checked locking. If you use the overload that doesn't take a lock then it does single-checked initialization.

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
rynowak: Yeah it's an appealing idea, but I'm not sure how we'd manage it.

src/Microsoft.AspNetCore.Routing/Matching/ILEmitTrieJumpTable.cs
rynowak: Do you have an alternative in mind?

Issue #678, opened 2018-08-01
Evaluate MatchProcessor design


JamesNK:
Review MatchProcessor design. MatchProcessor vs MatchProcessorBase.

JamesNK:
@rynowak Are you happy with this area now?

PR #734, opened 2018-08-21
Remove MatchProcessor, add IParameterPolicy

JamesNK:
https://github.com/aspnet/Routing/issues/678


src/Microsoft.AspNetCore.Routing/Patterns/RoutePattern.cs
```diff
@@ -48,10 +48,10 @@ public sealed class RoutePattern
         public IReadOnlyDictionary<string, object> Defaults { get; }
 
         /// <summary>
-        /// Gets the set of constraint references for the route pattern.
-        /// The keys of <see cref="Constraints"/> are the route parameter names.
+        /// Gets the set of parameter policy references for the route pattern.
+        /// The keys of <see cref="ParameterPolicies"/> are the route parameter names.
         /// </summary>
-        public IReadOnlyDictionary<string, IReadOnlyList<RoutePatternConstraintReference>> Constraints { get; }
+        public IReadOnlyDictionary<string, IReadOnlyList<RoutePatternParameterPolicyReference>> ParameterPolicies { get; }
```
JamesNK: What should happen here? Should all reference to constraints on `RoutePattern` and `RoutePatternFactory` be replaced with parameter policy?

src/Microsoft.AspNetCore.Routing.Abstractions/IParameterPolicy.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public interface IParameterPolicy
```
rynowak: docs?

src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -240,15 +240,16 @@ public override LinkGenerationTemplate GetTemplate(HttpContext httpContext, stri
...
                 var parameter = endpoint.RoutePattern.GetParameter(kvp.Key); // may be null, that's ok
                 var constraintReferences = kvp.Value;
                 for (var i = 0; i < constraintReferences.Count; i++)
                 {
                     var constraintReference = constraintReferences[i];
-                    var matchProcessor = _matchProcessorFactory.Create(parameter, constraintReference);
-                    if (!matchProcessor.ProcessOutbound(httpContext, routeValues))
+                    var parameterPolicy = _parameterPolicyFactory.Create(parameter, constraintReference);
+                    if (parameterPolicy is IRouteConstraint routeConstraint
+                        && !routeConstraint.Match(httpContext, NullRouter.Instance, kvp.Key, routeValues, RouteDirection.UrlGeneration))
```
rynowak: Beautiful, this is just about how I was hoping this would work out.

src/Microsoft.AspNetCore.Routing/Matching/DefaultParameterPolicyFactory.cs
```diff
@@ -2,45 +2,46 @@
...
-        public override MatchProcessor Create(string parameterName, MatchProcessor value, bool optional)
+        public override IParameterPolicy Create(string parameterName, IParameterPolicy value, bool optional)
         {
             if (value == null)
             {
                 throw new ArgumentNullException(nameof(value));
             }
 
-            return InitializeMatchProcessor(parameterName, optional, value, argument: null);
+            return value;
```
rynowak: This should call InitializeRouteConstraint (if it is a route constraint)

src/Microsoft.AspNetCore.Routing/Matching/ParameterPolicyFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Diagnostics;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Routing.Matching
+{
+    internal abstract class ParameterPolicyFactory
+    {
+        public abstract IParameterPolicy Create(string parameterName, string value, bool optional);
+
+        public abstract IParameterPolicy Create(string parameterName, IRouteConstraint value, bool optional);
```
rynowak: Is there any real reason we need this overload along with the `IParameterPolicy` overload?

src/Microsoft.AspNetCore.Routing/Matching/ParameterPolicyFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System;
+using System.Diagnostics;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Routing.Matching
+{
+    internal abstract class ParameterPolicyFactory
+    {
+        public abstract IParameterPolicy Create(string parameterName, string value, bool optional);
```
rynowak: I'm wondering if we want to have all of these methods take `RoutePatternParameterPart`. If we add more stuff to parameters in the future, it will be there.

JamesNK: It is an internal type. Does it matter that much?

Or do you foresee the factory becoming public?

rynowak: I think we were talking about tusing this from MVC during the building of endpoints right?

JamesNK: Good point.

I'm not a big fan of coupling it to patterns. I do think this class + abstract base class should be moved to Microsoft.AspNetCore.Routing namespace though.

JamesNK: Derp, it is already coupled to patterns.

src/Microsoft.AspNetCore.Routing/DefaultParameterPolicyFactory.cs
```diff
@@ -2,49 +2,46 @@
...
+using Microsoft.AspNetCore.Routing.Constraints;
+using Microsoft.AspNetCore.Routing.Patterns;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 
-namespace Microsoft.AspNetCore.Routing.Matching
+namespace Microsoft.AspNetCore.Routing
 {
-    internal class DefaultMatchProcessorFactory : MatchProcessorFactory
+    public class DefaultParameterPolicyFactory : ParameterPolicyFactory
```
rynowak: Can we keep the implementation of this internal? Why does this need to be made public. Apologies if I said somethign configusing.

JamesNK: What would MVC use to resolve inline policies? Re-implement `DefaultParameterPolicyFactory`?

rynowak: It would resolve `ParameterPolicyFactory` from DI? I'm just talking about the **implementation**

JamesNK: Of course. For a moment there I wasn't Thinking With DI.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -607,141 +608,147 @@ public static RoutePatternParameterPart ParameterPart(string parameterName, obje
...
         /// <summary>
-        /// Creates a <see cref="RoutePatternConstraintReference"/> from the provided object.
+        /// Creates a <see cref="RoutePatternParameterPolicyReference"/> from the provided contraint.
         /// </summary>
         /// <param name="constraint">
-        /// The constraint object, which must be of type <see cref="IRouteConstraint"/>,
-        /// <see cref="MatcherPolicy"/>, or <see cref="String"/>. If the constraint object
-        /// is a <see cref="String"/> it will be tranformed into an instance of 
-        /// <see cref="RegexRouteConstraint"/>.
+        /// The constraint object, which must be of type <see cref="IRouteConstraint"/>
```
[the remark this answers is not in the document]
rynowak: Yeah it is a bit awkward. That's the legacy behavior of the apis that accept an object literal of constraints. The 'inline' strings get looked up, and the 'object literal' ones get turned intoa  regex.

PR #735, opened 2018-08-21
Fix broken benchmarks

rynowak:
The tests for matching still compile but do the wrong thing, because
they aren't setting up the HTTP method metadata correctly.

Adding back an overload of CreateEndpoint that's like what was there
before.


rynowak:
I thought we had that, so it was much to my chagrin that I discovered it was broken. If you know what we need to do to enable this please share it :)

rynowak:
Ah, I figured out what was missing. On it.

JamesNK:
How was it breaking? If it was throwing an error then I believe you can check for this during CI. The SignalR build ran the micro-benchmarks once to check that there wasn't an error.

If it is wrong logic producing an incorrect result then assert the result in the benchmark test (unless you know of a better way).

rynowak:
I'd done the needful. We were missing some MSBuild junk to make the benchmark validation actually run. The vast majority of our benchmarks do validate their output and throw already.

benchmarks/Microsoft.AspNetCore.Routing.Performance/Matching/MatcherFindCandidateSetAzureBenchmark.cs
```diff
@@ -10,8 +10,8 @@ namespace Microsoft.AspNetCore.Routing.Matching
     public partial class MatcherFindCandidateSetAzureBenchmark : EndpointRoutingBenchmarkBase
     {
         // SegmentCount should be max-segments + 1, but we don't have a good way to compute
-        // it here, so using 16 as a safe guess.
-        private const int SegmentCount = 16;
+        // it here, so using 32 as a safe guess.
+        private const int SegmentCount = 32;
```
rynowak: Oops. This was not a safe guess.

rynowak:
@JamesNK updated

Issue #648, opened 2018-07-23
Tests with non-MatcherEndpoint


JamesNK:
See https://github.com/aspnet/Routing/pull/646#discussion_r204257353

Make sure that new functionality works with non-MatcherEndpoints as expected.

rynowak:
Note that how this is shaping up, we'll have `Endpoint` and `RouteEndpoint`. We ascribe 'routing' semantics to `RouteEndpoint` and 'execution' semantics to `Endpoint`.

I think it would be enough for this work item to change `HttpMethodMatcherPolicy` to use `Endpoint` instead of `RouteEndpoint`. We already have tests that we don't inject non-RouteEndpoint types into the routing system, and this is the one other place it can happen. We don't really have places in the code that I can find where we downcast and it isn't intended.

JamesNK:
I recall my main concern was non-MatcherEndpoints being passed to the matcher. The casting going on inside matching made me nervious. I want to ensure that the DFA matcher doesn't let non-MatcherEndpoints slip through and throw InvalidCastExceptions.

rynowak:
Ah yes, I found it. https://github.com/aspnet/Routing/blob/release/2.2/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs#L550 This is the one.

JamesNK:
Ka-boom
```
System.InvalidCastException: Unable to cast object of type &#x27;Microsoft.AspNetCore.Http.Endpoint&#x27; to type &#x27;Microsoft.AspNetCore.Routing.RouteEndpoint&#x27;.
   at System.Linq.Enumerable.CastIterator[TResult](IEnumerable source)&#x2B;MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.AspNetCore.Routing.Matching.DfaMatcherBuilder.ApplyPolicies(DfaNode node) in /_/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs:line 547
   at Microsoft.AspNetCore.Routing.Matching.DfaNode.Visit(Action`1 visitor) in /_/src/Microsoft.AspNetCore.Routing/Matching/DfaNode.cs:line 69
[...]

PR #739, opened 2018-08-22
Fix DfaMatcher to work with non-RouteEndpoint

JamesNK:
Fixes https://github.com/aspnet/Routing/issues/648


src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -367,76 +368,80 @@ internal Candidate[] CreateCandidates(IReadOnlyList<RouteEndpoint> endpoints)
...
-            foreach (var kvp in endpoint.RoutePattern.Defaults)
-            {
-                assignments.Add(kvp.Key, assignments.Count);
-                slots.Add(kvp);
-            }
+            var complexSegments = new List<(RoutePatternPathSegment pathSegment, int segmentIndex)>();
+            var constraints = new List<KeyValuePair<string, IRouteConstraint>>();
 
-            for (var i = 0; i < endpoint.RoutePattern.PathSegments.Count; i++)
+            if (endpoint is RouteEndpoint routeEndpoint)
```
JamesNK: Skip all route endpoint specific logic here

src/Microsoft.AspNetCore.Routing/Matching/EndpointComparer.cs
```diff
@@ -0,0 +1,162 @@
...
+                }
+
+                return 1;
+            }
+            else if (routeEndpointY != null)
+            {
+                return -1;
+            }
+
+            return 0;
```
JamesNK: This check that the Endpoint is a RouteEndpoint is a bit verbose in each compare method :(

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -250,15 +250,13 @@ public PolicyJumpTable BuildJumpTable(int exitDestination, IReadOnlyList<PolicyJ
         private Endpoint CreateRejectionEndpoint(IEnumerable<string> httpMethods)
         {
             var allow = string.Join(", ", httpMethods);
-            return new RouteEndpoint(
+            return new Endpoint(
```
JamesNK: This provides test coverage. Think we need more specific non-RouteEndpoint tests?

src/Microsoft.AspNetCore.Routing/Matching/EndpointComparer.cs
```diff
@@ -0,0 +1,162 @@
...
+    // for tests, errors, etc.
+    //
+    // When we group endpoints we don't consider the route template, because we're trying
+    // to group endpoints not separate them.
+    //
+    // TLDR: 
+    //  IComparer implementation considers the template string as a tiebreaker.
+    //  IEqualityComparer implementation does not.
+    //  This is cool and good.
+    internal class EndpointComparer : IComparer<Endpoint>, IEqualityComparer<Endpoint>
```
rynowak: Yeah this was the one part I was a little unsure what to do about. The built in comparers look a route-endpoint specific stuff.

src/Microsoft.AspNetCore.Routing/Matching/EndpointComparer.cs
rynowak: yeah :-/

src/Microsoft.AspNetCore.Routing/Matching/EndpointComparer.cs
```diff
@@ -0,0 +1,162 @@
...
+            Debug.Assert(x != null);
+            Debug.Assert(y != null);
+
+            var compare = CompareCore(x, y);
+
+            // Since we're sorting, use the route template as a last resort.
+            return compare == 0 ? ComparePattern(x, y) : compare;
+        }
+
+        private int ComparePattern(Endpoint x, Endpoint y)
```
rynowak: Worth a comment I think that a RouteEndpoint always comes before a non-Route-endpoint in the rules of this method.

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
rynowak: I think it would be good to have one unit test in case this implementation changes more. Curently the only way to do this is with a policy - see TestMetadata1MatcherPolicy in DfaMatcherBuilderTest

rynowak:
Thanks for doing this 👍 I was planning to last night, but then I got into the weeds with the comparer thing. I slept on it and decided on a solution, then I came to work and found that someone had already coded it up for me. Unless this is a smoke-fueled fever dream...

src/Microsoft.AspNetCore.Routing/Matching/EndpointComparer.cs
JamesNK: Hmmm, I wonder if these comparers should return 0 if both endpoints are not RouteEndpoint, and then this logic is handled in the final pattern comparison. That will let `IEndpointComparerPolicy` have a chance to have logic for comparing a RouteEndpoint with a non-RouteEndpoint.

What do you think?

JamesNK:
🆙 📅

src/Microsoft.AspNetCore.Routing/Matching/EndpointComparer.cs
rynowak: >  have a chance to have logic for comparing a RouteEndpoint with a non-RouteEndpoint.

I'm not sure how useful it is. The use case for non-`RouteEndpoint` is to inject it as a failure case. It seems natural to me to sort those to the end of the list. 

I would expect that a `RouteEndpoint` > (other) for all cases

PR #741, opened 2018-08-22
Set the endpoint feature only on success

rynowak:
@JamesNK - what do you think of this idea? Obvi no test changes yet, just continuing discussion from your PR

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/EndpointMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Routing/EndpointMiddleware.cs
@@ -32,26 +32,18 @@ namespace Microsoft.AspNetCore.Routing
 
         public async Task Invoke(HttpContext httpContext)
         {
-            var feature = httpContext.Features.Get<IEndpointFeature>();
-            if (feature == null)
+            var endpoint = httpContext.Features.Get<IEndpointFeature>()?.Endpoint;
+            if (endpoint?.RequestDelegate != null)
             {
-                var message = $"Unable to execute an endpoint because the {nameof(EndpointRoutingMiddleware)} was not run for this request. " +
-                    $"Ensure {nameof(EndpointRoutingMiddleware)} is added to the request execution pipeline before {nameof(EndpointMiddleware)} in application startup code.";
-
-                throw new InvalidOperationException(message);
-            }
-
-            if (feature.Endpoint?.RequestDelegate != null)
-            {
-                Log.ExecutingEndpoint(_logger, feature.Endpoint);
+                Log.ExecutingEndpoint(_logger, endpoint);
 
                 try
                 {
-                    await feature.Endpoint.RequestDelegate(httpContext);
+                    await endpoint.RequestDelegate(httpContext);
                 }
                 finally
                 {
-                    Log.ExecutedEndpoint(_logger, feature.Endpoint);
+                    Log.ExecutedEndpoint(_logger, endpoint);
                 }
 
                 return;
--- a/src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
@@ -55,10 +55,7 @@ namespace Microsoft.AspNetCore.Routing
 
         public async Task Invoke(HttpContext httpContext)
         {
-            // For back-compat EndpointRouteValuesFeature implements IEndpointFeature, IRouteValuesFeature and IRoutingFeature
             var feature = new EndpointFeature();
-
-            SetEndpointFeature(httpContext, feature);
 
             // There's an inherent race condition between waiting for init and accessing the matcher
             // this is OK because once `_matcher` is initialized, it will not be set to null again.
@@ -67,6 +64,10 @@ namespace Microsoft.AspNetCore.Routing
             await matcher.MatchAsync(httpContext, feature);
             if (feature.Endpoint != null)
             {
+                // Set the endpoint feature only on success. This means we won't overwrite any
+                // existing state for related features unless we did something.
+                SetEndpointFeature(httpContext, feature);
+
                 Log.MatchSuccess(_logger, feature);
             }
             else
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
```diff
@@ -55,18 +55,19 @@ internal sealed class EndpointRoutingMiddleware
...
             // There's an inherent race condition between waiting for init and accessing the matcher
             // this is OK because once `_matcher` is initialized, it will not be set to null again.
             var matcher = await InitializeAsync();
 
             await matcher.MatchAsync(httpContext, feature);
             if (feature.Endpoint != null)
             {
+                // Set the endpoint feature only on success. This means we won't overwrite any
+                // existing state for related features unless we did something.
+                SetEndpointFeature(httpContext, feature);
```
rynowak: @davidfowl @Tratcher - what's the right thing to do here. Is it good or bad to avoid setting the feature if we don't match?

rynowak: Ping. Do you guys have an opinion on this?

davidfowl: I think it makes sense.

rynowak:
@JamesNK  -updated

src/Microsoft.AspNetCore.Routing/EndpointMiddleware.cs
```diff
@@ -32,26 +32,18 @@ public EndpointMiddleware(ILogger<EndpointMiddleware> logger, RequestDelegate ne
 
         public async Task Invoke(HttpContext httpContext)
         {
-            var feature = httpContext.Features.Get<IEndpointFeature>();
-            if (feature == null)
+            var endpoint = httpContext.Features.Get<IEndpointFeature>()?.Endpoint;
+            if (endpoint?.RequestDelegate != null)
             {
-                var message = $"Unable to execute an endpoint because the {nameof(EndpointRoutingMiddleware)} was not run for this request. " +
```
rynowak: With hosting taking the responsibility of invoking the endpoint in 3.0 we won't need this anymore.

JamesNK: Hmmm, tbh it depends on what we want from this middleware.

Is the purpose of it:
1. I have an endpoint, stop now and don't execute anymore middlware (and if there isn't then throw an exception)
2. Stop if we have an endpoint, otherwise keep going

IMO I think (1) is what people would expect as a default. Maybe it should be an option?

src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
JamesNK: From a backwards compat perspective, what does the current routing do? Will it always set IRoutingFeature? Does it matter if that behavior changes?

rynowak: Current routing middleware only sets the feature if it found something (and executes it immediately). It's not super meaningful outside of MVC because old routing is terminal.

src/Microsoft.AspNetCore.Routing/EndpointMiddleware.cs
rynowak: So in 3.0 I don't think there's a purpose for this middleware at all - because this becomes a built in behavior - so the question is academic. Does that match your understanding?

I would definitely expect 2. Middleware generally don't have preconditions to validate, they no op if there's nothing to do.

src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
JamesNK: What happens if UseMvc doesn't match anything? Does it throw/return then or does the host return 404 at the end?

src/Microsoft.AspNetCore.Routing/EndpointMiddleware.cs
JamesNK: In my mind this is like a return statement from a method. Even if a method implicitly returned a value at the end, I still then there are cases where you might want to explicitly return earlier.

> I would definitely expect 2. Middleware generally don't have preconditions to validate, they no op if there's nothing to do.

Ok, but I wonder if this middleware should be stricter since it is an endpoint rather a step along a pipline.

Maybe I'm overthinking this 😬

src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
rynowak: UseMvc just yields control the next middleware

src/Microsoft.AspNetCore.Routing/EndpointMiddleware.cs
rynowak: I don't think you're overthinking, I think you're ascribing some personality characteristics to middleware that they don't/shouldn't have. In general middleware are designed for 
 maximum composability.

Put another way, if you don't want to handle something, maybe someone else does. It's much better to defer the responsiblity to decide how to response than it is to report a 404 or a 500 yourself. It's the chain of responsibility pattern.

JamesNK: Ok. If someone does want to force a hard exit then they have tools to do so easily without a helper middleware.

src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
JamesNK: Great. Might be worth having a functional test of middleware that follows UseMvc to check compatibility with endpoint routing.

JamesNK:
Include this for preview 2?

Test MVC with a feature branch before merging.

src/Microsoft.AspNetCore.Routing/EndpointRoutingMiddleware.cs
Tratcher: If you did set the feature for a non-match, what would it contain? Nulls? Then you're better off not setting the feature.

Behaviorally all feature interfaces should be treated as optional except for request and response.

rynowak:
https://github.com/aspnet/Mvc/pull/8364

rynowak:
The new functional test in MVC passes without these changes, currently resolving a build break to unblock testing a feature branch.

rynowak:
new test in MVC passes with and without these changes

PR #744, opened 2018-08-23
Improvements for IEndpointSelectorPolicy

rynowak:
These changes are based on our discussion earlier this week. Adding
async, and making it possible to short circuit, which should be better
aligned with the requirments of versioning.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/DefaultEndpointSelector.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/DefaultEndpointSelector.cs
@@ -24,16 +24,30 @@ namespace Microsoft.AspNetCore.Routing.Matching
             _selectorPolicies = matcherPolicies.OrderBy(p => p.Order).OfType<IEndpointSelectorPolicy>().ToArray();
         }
 
-        public override Task SelectAsync(
+        public override async Task SelectAsync(
             HttpContext httpContext,
             EndpointFeature feature,
             CandidateSet candidateSet)
         {
+            var selectorPolicies = _selectorPolicies;
             for (var i = 0; i < _selectorPolicies.Length; i++)
             {
-                _selectorPolicies[i].Apply(httpContext, candidateSet);
+                await selectorPolicies[i].ApplyAsync(httpContext, feature, candidateSet);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Routing/Matching/DefaultEndpointSelector.cs
```diff
@@ -24,16 +24,30 @@ public DefaultEndpointSelector(IEnumerable<MatcherPolicy> matcherPolicies)
...
+                {
+                    // This is a short circuit, the selector chose an endpoint.
+                    return;
+                }
             }
 
+            ProcessFinalCandidates(httpContext, feature, candidateSet);
+        }
+
+        private static void ProcessFinalCandidates(
```
rynowak: FYI this isn't just for cleanliness, we're using ref var which requires us to make it non-async. We might need to to think about changes to this data structure if it's too hard to use in async methods.

src/Microsoft.AspNetCore.Routing/Matching/DefaultEndpointSelector.cs
```diff
@@ -24,16 +24,30 @@ public DefaultEndpointSelector(IEnumerable<MatcherPolicy> matcherPolicies)
...
+        public override async Task SelectAsync(
             HttpContext httpContext,
             EndpointFeature feature,
             CandidateSet candidateSet)
         {
+            var selectorPolicies = _selectorPolicies;
             for (var i = 0; i < _selectorPolicies.Length; i++)
             {
-                _selectorPolicies[i].Apply(httpContext, candidateSet);
+                await selectorPolicies[i].ApplyAsync(httpContext, feature, candidateSet);
```
JamesNK: What is the perf hit of this? If it is called once per-match then state machine overhead will be tiny but I assume there is a Task allocation. Is it worth providing a non-async path?

JamesNK: Ok! I couldn't remember exactly when a Task was allocated. Optimizing a state machine called once per match is def not worth it.

rynowak:
Thanks 👍

PR #747, opened 2018-08-23
RoutePattern does not throw on same duplicate default values

JamesNK:


JamesNK:
@rynowak I'm merging this because both this and IEndpointSelector need a feature branch, and IEndpointSelector has breaking changes for MVC that need to be fixed before they break the build.

JamesNK:
Will deal with any comments you have in a follow up.

PR #750, opened 2018-08-26
Initial ParameterTransformer support

JamesNK:
Basic ParameterTransformer implementation.

This reaches into `TemplateBinder` and `UriBuildingContext`. If they aren't going to be used by the final link generation iteration then these changes could be rolled back later.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/IParameterTransformer.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/IParameterTransformer.cs
@@ -3,8 +3,8 @@
 
 namespace Microsoft.AspNetCore.Routing
 {
-    public abstract class ParameterTransformer : IParameterPolicy
+    public interface IParameterTransformer : IParameterPolicy
     {
-        public abstract string Transform(string value);
+        string Transform(string value);
     }
 }
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/IParameterTransformer.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/IParameterTransformer.cs
@@ -3,8 +3,16 @@
 
 namespace Microsoft.AspNetCore.Routing
 {
+    /// <summary>
+    /// Defines the contract that a class must implement to transform parameter values.
+    /// </summary>
     public interface IParameterTransformer : IParameterPolicy
     {
+        /// <summary>
+        /// Transforms the specified parameter value.
+        /// </summary>
+        /// <param name="value">The parameter value to transform.</param>
+        /// <returns>The transformed value.</returns>
         string Transform(string value);
     }
 }
--- a/src/Microsoft.AspNetCore.Routing/Internal/NullRouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/Internal/NullRouteConstraint.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Internal
+{
+    internal class NullRouteConstraint : IRouteConstraint
+    {
+        public static readonly NullRouteConstraint Instance = new NullRouteConstraint();
+
+        private NullRouteConstraint()
+        {
+        }
+
+        public bool Match(HttpContext httpContext, IRouter route, string routeKey, RouteValueDictionary values, RouteDirection routeDirection)
+        {
+            return true;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
+++ b/src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
@@ -59,7 +59,7 @@ namespace Microsoft.AspNetCore.Routing.Internal
             return Accept(value, encodeSlashes: true, parameterTransformer: null);
         }
 
-        public bool Accept(string value, bool encodeSlashes, ParameterTransformer parameterTransformer)
+        public bool Accept(string value, bool encodeSlashes, IParameterTransformer parameterTransformer)
         {
             if (string.IsNullOrEmpty(value))
             {
@@ -266,13 +266,13 @@ namespace Microsoft.AspNetCore.Routing.Internal
             EncodeValue(value, encodeSlashes: true, parameterTransformer: null);
         }
 
-        private void EncodeValue(string value, bool encodeSlashes, ParameterTransformer parameterTransformer)
+        private void EncodeValue(string value, bool encodeSlashes, IParameterTransformer parameterTransformer)
         {
             EncodeValue(value, start: 0, characterCount: value.Length, encodeSlashes, parameterTransformer);
         }
 
         // For testing
-        internal void EncodeValue(string value, int start, int characterCount, bool encodeSlashes, ParameterTransformer parameterTransformer)
+        internal void EncodeValue(string value, int start, int characterCount, bool encodeSlashes, IParameterTransformer parameterTransformer)
         {
             if (parameterTransformer != null)
             {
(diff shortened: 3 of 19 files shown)


src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
```diff
@@ -205,17 +205,24 @@ public override string ToString()
...
         // For testing
-        internal void EncodeValue(string value, int start, int characterCount, bool encodeSlashes)
+        internal void EncodeValue(string value, int start, int characterCount, bool encodeSlashes, ParameterTransformer parameterTransformer)
         {
+            if (parameterTransformer != null)
+            {
+                value = parameterTransformer.Transform(value.Substring(0, characterCount));
+                start = 0;
+                characterCount = value.Length;
+            }
```
rynowak: Hmm this is a little wierd, and maybe I paged out my understanding of this code at a bad time. When is this called with a subsection of a value?

src/Microsoft.AspNetCore.Routing/ParameterTransformer.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public abstract class ParameterTransformer : IParameterPolicy
```
rynowak: I think this should probably be an interface, since you can be more than one class. Any good reason for this to be a class?

src/Microsoft.AspNetCore.Routing/ParameterTransformer.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        public abstract string Transform(string value);
```
rynowak: I think we need the ability to run this on an inbound value as well don't we?

rynowak: I guess when you consider that, you'd want to have another argument that indicates the direction (or a separate method, which I'd probably prefer).

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -273,6 +277,26 @@ private string BindValues(UriBuildingContext context, RouteValueDictionary accep
...
         }
 
+        private ParameterTransformer GetParameterTransformer(RoutePatternParameterPart parameterPart)
+        {
+            if (_parameterPolicyFactory == null)
+            {
+                return null;
+            }
+
+            for (int i = 0; i < parameterPart.ParameterPolicies.Count; i++)
```
rynowak: var

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -273,6 +277,26 @@ private string BindValues(UriBuildingContext context, RouteValueDictionary accep
...
+            {
+                // Use the first parameter transformer
+                var parameterPolicy = _parameterPolicyFactory.Create(parameterPart, parameterPart.ParameterPolicies[0]);
+                if (parameterPolicy is ParameterTransformer parameterTransformer)
+                {
+                    return parameterTransformer;
+                }
```
rynowak: Is there a problem with supporting multiple? Should we throw? What do you think is most reasonable?

src/Microsoft.AspNetCore.Routing/ParameterTransformer.cs
```diff
@@ -0,0 +1,10 @@
...
+    }
+}
```
rynowak: Philosophical question - should this run before or after a constraint?

rynowak: If it's before - can you reject on invalid data?

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -830,6 +830,35 @@ public void GetLink_InlineConstraints_CompositeInlineConstraint_Fails()
...
+            var services = GetBasicServices();
+            services.AddSingleton(typeof(UpperCaseParameterTransform), new UpperCaseParameterTransform());
+
+            var linkGenerator = CreateLinkGenerator(new[] { endpoint }, routeOptions, services);
+
+            // Act
+            var link = linkGenerator.GetLink(new { controller = "Home", name = "Test" });
+
+            // Assert
+            Assert.Equal("/HOME/Test", link);
```
rynowak: I would also expect this to be part of DFA matcher when the parameter actually appears in the route template.

So something like an `id` parameter would run the transformer per-request. Something like an `action` parameter would run statically when building the endpoint, and then be ignored.

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
JamesNK: It is used here:

```
// Allow the first segment to have a leading slash.
// This prevents the leading slash from PathString segments from being encoded.
if (_path.Length == 0 && value.Length > 0 && value[0] == '/')
{
    _path.Append("/");
    EncodeValue(value, 1, value.Length - 1, encodeSlashes, parameterTransformer);
}
else
{
    EncodeValue(value, encodeSlashes, parameterTransformer);
}
```

JamesNK:
TODO: Add to DFA matcher for non static params

JamesNK:
@rynowak I'm not sure calling a parameter transformer for non-static parameters makes sense. Transforming the parameter value will have happened when the link is generated. Transforming it again by the matcher would likely cause bugs.

rynowak:
> Transforming the parameter value will have happened when the link is generated. Transforming it again by the matcher would likely cause bugs.

OK, let's agree for now to use parameter transformers for 'static' parameters at startup time for now. I want to consider a set of features where we support their use at runtime for 'data' parameters, but I think we can consider that separately.

src/Microsoft.AspNetCore.Routing.Abstractions/IParameterTransformer.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public interface IParameterTransformer : IParameterPolicy
```
rynowak: Needs docs. We should likely explain that this is uni-directional, and only applies to static parameter replacement.

src/Microsoft.AspNetCore.Routing/Internal/NullRouteConstraint.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Internal
```
rynowak: No need for `.Internal` here. This should live with the other route constraint implementations.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -265,7 +269,7 @@ private bool TryBindValuesCore(UriBuildingContext context, RouteValueDictionary
                             // Example: template = {id}.{format?}. parameters: id=5
                             // In this case after we have generated "5.", we wont find any value 
                             // for format, so we remove '.' and generate 5.
-                            if (!context.Accept(converted, parameterPart.EncodeSlashes))
+                            if (!context.Accept(converted, parameterPart.EncodeSlashes, GetParameterTransformer(parameterPart)))
```
rynowak: Wait OK, I'm confused now, I thought you were saying elsewhere that it made sense for parameter tranformers to only apply staticly?

rynowak: Discussed on the call, makes sense.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -71,6 +74,7 @@ public class TemplateBinder
             _pool = pool;
             _pattern = pattern;
             _defaults = defaults;
+            _parameterPolicyFactory = parameterPolicyFactory;
```
rynowak: FYI I will end up adding caching for parameter policies as part of my work.

PR #753, opened 2018-08-28
Support policy arguments and resolving services by constructors 

JamesNK:
I looked into `ActivatorUtilities` and it isn't suitable. The way its constructor matching works is it relies on the parameters being the correct type, and then assumes everything else comes from DI. For inline policies, the arguments for a constructor will always be strings, so that won't work.

Meanwhile constraint constructors are matched on the number of parameters, and then args are converted to the correct types.

What I have done is assume that anything that is populated from the inline argument is `IConvertible` (today `Convert.ChangeType` is used in constraint resolver so that is an ok assumption), and everything else is attempted to be resolved from DI.

I want `DefaultInlineConstraintResolver` and `DefaultParameterPolicyFactory` to use the same code internally. Both will be thin wrappers over the same logic, the difference being the constraint resolver will blow up if the resolve type isn't a constraint.

@rynowak Let me know whether you are ok with this approach. After that I will finish cleaning it up.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Constraints/HttpMethodRouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Routing/Constraints/HttpMethodRouteConstraint.cs
@@ -41,16 +41,6 @@ namespace Microsoft.AspNetCore.Routing.Constraints
             RouteValueDictionary values,
             RouteDirection routeDirection)
         {
-            if (httpContext == null)
-            {
-                throw new ArgumentNullException(nameof(httpContext));
-            }
-
-            if (route == null)
-            {
-                throw new ArgumentNullException(nameof(route));
-            }
-
             if (routeKey == null)
             {
                 throw new ArgumentNullException(nameof(routeKey));
@@ -64,6 +54,12 @@ namespace Microsoft.AspNetCore.Routing.Constraints
             switch (routeDirection)
             {
                 case RouteDirection.IncomingRequest:
+                    // Only required for constraining incoming requests
+                    if (httpContext == null)
+                    {
+                        throw new ArgumentNullException(nameof(httpContext));
+                    }
+
                     return AllowedMethods.Contains(httpContext.Request.Method, StringComparer.OrdinalIgnoreCase);
 
                 case RouteDirection.UrlGeneration:
--- a/src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
+++ b/src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
@@ -18,6 +18,7 @@ namespace Microsoft.AspNetCore.Routing
     public class DefaultInlineConstraintResolver : IInlineConstraintResolver
     {
         private readonly IDictionary<string, Type> _inlineConstraintMap;
+        private readonly IServiceProvider _serviceProvider;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DefaultInlineConstraintResolver"/> class.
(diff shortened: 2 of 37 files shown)


src/Microsoft.AspNetCore.Routing/Constraints/HttpMethodRouteConstraint.cs
```diff
@@ -64,6 +54,12 @@ public HttpMethodRouteConstraint(params string[] allowedMethods)
             switch (routeDirection)
             {
                 case RouteDirection.IncomingRequest:
+                    // Only required for constraining incoming requests
+                    if (httpContext == null)
+                    {
+                        throw new ArgumentNullException(nameof(httpContext));
+                    }
```
rynowak: Makes sense because you'd always have on on incoming requests.

src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
```diff
@@ -26,8 +25,14 @@ public class DefaultInlineConstraintResolver : IInlineConstraintResolver
         /// Accessor for <see cref="RouteOptions"/> containing the constraints of interest.
         /// </param>
         public DefaultInlineConstraintResolver(IOptions<RouteOptions> routeOptions)
```
rynowak: Should we obsolete this? If anyone is using it, we'd recommend that we use the other constructor now.

src/Microsoft.AspNetCore.Routing/Internal/ParameterPolicyActivator.cs
```diff
@@ -0,0 +1,164 @@
+﻿using System;
```
rynowak: copyright header

src/Microsoft.AspNetCore.Routing/Internal/ParameterPolicyActivator.cs
```diff
@@ -0,0 +1,164 @@
...
+    internal static class ParameterPolicyActivator
+    {
+        public static T ResolveParameterPolicy<T>(IDictionary<string, Type> inlineParameterPolicyMap, IServiceProvider serviceProvider, string inlineParameterPolicy, out string parameterPolicyKey)
+            where T : IParameterPolicy
+        {
+            if (inlineParameterPolicyMap == null)
+            {
+                throw new ArgumentNullException(nameof(inlineParameterPolicyMap));
+            }
+
```
rynowak: Should we null check the service provider?

src/Microsoft.AspNetCore.Routing/Internal/ParameterPolicyActivator.cs
```diff
@@ -0,0 +1,164 @@
...
+            {
+                throw new RouteCreationException(
+                    $"An error occurred while trying to create an instance of '{parameterPolicyType.FullName}'.",
+                    exception);
+            }
+        }
+
+        private static IParameterPolicy CreateParameterPolicy(IServiceProvider serviceProvider, Type constraintType, string argumentString)
+        {
+            var constraintTypeInfo = constraintType.GetTypeInfo();
```
rynowak: This  should just be `GetType` - all the type info stuff is legacy from the bad old days 😆

src/Microsoft.AspNetCore.Routing/DefaultInlineConstraintResolver.cs
JamesNK: Sure

src/Microsoft.AspNetCore.Routing/Internal/ParameterPolicyActivator.cs
```diff
@@ -0,0 +1,164 @@
...
+            else
+            {
+                var arguments = !string.IsNullOrEmpty(argumentString)
+                    ? argumentString.Split(',').Select(argument => argument.Trim()).ToArray()
+                    : Array.Empty<string>();
+
+                var matchingConstructors = constructors
+                    .Where(ci => GetNonConvertableParameterTypeCount(serviceProvider, ci.GetParameters()) == arguments.Length)
+                    .OrderByDescending(ci => ci.GetParameters().Length)
+                    .ToArray();
```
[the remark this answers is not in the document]
JamesNK: It is. `ActivatorUtilities` considers a best match the ctor that matches the parameters and then has the most number of arguments to resolve from DI.

PR #755, opened 2018-08-29
Unit test constraints without HttpContext

JamesNK:


JamesNK:
Opps, hit merge by mistake. If you have any feedback I'll fix in a follow up PR.

PR #765, opened 2018-09-02
Reuse collections in DfaMatcherBuilder

JamesNK:
Before:
```
    Dfa | 357.2 ms | 7.020 ms | 6.567 ms | 2.799 | 937.5000 | 875.0000 | 812.5000 | 195.18 MB |
```

After:
```
    Dfa | 308.8 ms | 5.961 ms | 5.284 ms | 3.238 | 937.5000 | 937.5000 | 812.5000 | 150.32 MB |
```

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
@@ -9,33 +9,6 @@ using Microsoft.AspNetCore.Routing.Patterns;
 
 namespace Microsoft.AspNetCore.Routing.Matching
 {
-    internal ref struct CandidateBuilderState
-    {
-        public Dictionary<string, int> assignments;
-        public List<KeyValuePair<string, object>> slots;
-        public List<(string parameterName, int segmentIndex, int slotIndex)> captures;
-        public List<(RoutePatternPathSegment pathSegment, int segmentIndex)> complexSegments;
-        public List<KeyValuePair<string, IRouteConstraint>> constraints;
-
-        public void Initialize()
-        {
-            assignments = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
-            slots = new List<KeyValuePair<string, object>>();
-            captures = new List<(string parameterName, int segmentIndex, int slotIndex)>();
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -9,6 +9,33 @@
 
 namespace Microsoft.AspNetCore.Routing.Matching
 {
+    internal ref struct CandidateBuilderState
```
JamesNK: I started doing this thing to see what it felt like and quickly felt it was stupid. Validate this is stupid for me and I'll move this state to fields on DfaMatcherBuilder

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -362,7 +362,7 @@ public override bool Equals(object obj)
             public override int GetHashCode()
             {
                 var hash = new HashCodeCombiner();
-                hash.Add(IsCorsPreflightRequest);
+                hash.Add(IsCorsPreflightRequest ? 1 : 0);
```
JamesNK: There was no bool overload so it was boxing to object.

JamesNK:
🆙 📅

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -370,20 +414,20 @@ internal Candidate[] CreateCandidates(IReadOnlyList<Endpoint> endpoints)
         // internal for tests
         internal Candidate CreateCandidate(Endpoint endpoint, int score)
         {
-            var assignments = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
-            var slots = new List<KeyValuePair<string, object>>();
-            var captures = new List<(string parameterName, int segmentIndex, int slotIndex)>();
-            (string parameterName, int segmentIndex, int slotIndex) catchAll = default;
+            _assignments.Clear();
```
drieseng: Do you want to clear these lists at all when **endpoint** is not a **RouteEndpoint**?
In that case you could create a **Candidate** instance with only **Array.Empty&lt;T&gt;** instances
No idea how common this is though.

PR #766, opened 2018-09-02
Create collections on DfaNode as needed

JamesNK:
Before:
```
    Dfa | 308.8 ms | 5.961 ms | 5.284 ms | 3.238 | 937.5000 | 937.5000 | 812.5000 | 150.32 MB |
```

After:
```
    Dfa | 337.8 ms | 4.650 ms | 4.122 ms | 2.960 | 812.5000 | 750.0000 | 687.5000 | 138.66 MB |
```

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/DfaNode.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/DfaNode.cs
@@ -14,13 +14,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
     [DebuggerDisplay("{DebuggerToString(),nq}")]
     internal class DfaNode
     {
-        public DfaNode()
-        {
-            Literals = new Dictionary<string, DfaNode>(StringComparer.OrdinalIgnoreCase);
-            Matches = new List<Endpoint>();
-            PolicyEdges = new Dictionary<object, DfaNode>();
-        }
-
         // The depth of the node. The depth indicates the number of segments
         // that must be processed to arrive at this node.
         //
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Routing/Matching/DfaNode.cs
```diff
@@ -30,23 +23,66 @@ public DfaNode()
...
+        {
+            if (Matches == null)
+            {
+                Matches = new List<Endpoint>();
+            }
+
+            Matches.Add(endpoint);
+        }
+
+        public void AddMatches(IEnumerable<Endpoint> endpoints)
```
drieseng: Nit: 
Would it be better to use **IReadOnlyList&lt;EndPoint&gt;** instead of **IEnumerable&lt;EndPoint&gt;**?
That way if ever `List<T>.AddRange(IEnumerable<T>)` specializes for **IReadOnlyList&lt;T&gt;** we can reap the benefits of this.

src/Microsoft.AspNetCore.Routing/Matching/DfaNode.cs
```diff
@@ -30,23 +23,66 @@ public DfaNode()
...
+        {
+            if (Matches == null)
+            {
+                Matches = new List<Endpoint>();
```
drieseng: Immediately pass the endpoints in the .ctor of **List&lt;Endpoint&gt;** when **Matches** is **null**?

src/Microsoft.AspNetCore.Routing/Matching/DfaNode.cs
JamesNK: Hmm, I don't think that would happen.

drieseng: I agree it's not likely. Ignore my remark.

drieseng:
@JamesNK where do I find the benchmark that you used for these results?

PR #767, opened 2018-09-03
Avoid resizing large struct arrays

JamesNK:
Before:
```
    Dfa | 337.8 ms | 4.650 ms | 4.122 ms | 2.960 | 812.5000 | 750.0000 | 687.5000 | 138.66 MB |
```

After:
```
    Dfa | 324.0 ms | 6.375 ms | 14.52 ms | 3.087 | 562.5000 | 500.0000 | 375.0000 | 130.02 MB |
```

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
@@ -26,6 +26,8 @@ namespace Microsoft.AspNetCore.Routing.Matching
         private readonly List<(RoutePatternPathSegment pathSegment, int segmentIndex)> _complexSegments;
         private readonly List<KeyValuePair<string, IRouteConstraint>> _constraints;
 
+        private int _stateIndex;
+
         public DfaMatcherBuilder(
             ParameterPolicyFactory parameterPolicyFactory,
             EndpointSelector selector,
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -306,31 +313,32 @@ public override Matcher Build()
...
-            return new DfaMatcher(_selector, states.ToArray(), maxSegmentCount);
+            return new DfaMatcher(_selector, states, maxSegmentCount);
         }
 
         private int AddNode(
             DfaNode node,
-            List<DfaState> states,
-            List<(JumpTableBuilder pathBuilder, PolicyJumpTableBuilder policyBuilder)> tableBuilders)
+            DfaState[] states,
+            ref int stateIndex,
```
JamesNK: Should probably be moved to a field on the builder

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -306,31 +315,31 @@ public override Matcher Build()
...
             DfaNode node,
-            List<DfaState> states,
-            List<(JumpTableBuilder pathBuilder, PolicyJumpTableBuilder policyBuilder)> tableBuilders)
+            DfaState[] states,
+            (JumpTableBuilder pathBuilder, PolicyJumpTableBuilder policyBuilder)[] tableBuilders)
         {
             node.Matches?.Sort(_comparer);
 
-            var stateIndex = states.Count;
+            int currentStateIndex = _stateIndex;
```
rynowak: var

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -377,20 +386,28 @@ public override Matcher Build()
...
                 // Break cycles
-                return ReferenceEquals(node, next) ? stateIndex : AddNode(next, states, tableBuilders);
+                if (ReferenceEquals(node, next))
+                {
+                    return _stateIndex;
+                }
+                else
+                {
+                    _stateIndex++;
+                    return AddNode(next, states, tableBuilders);
```
rynowak: AddNode should take the index as a parameter rather than tunneling through a field.

JamesNK: That is what I originally had but it becomes unwieldy because it has to be passed by ref.

See: https://github.com/aspnet/Routing/pull/767/commits/082dafae94a790453059fb033e86177faa4d89d4

Are you sure?

drieseng:
Not related to this PR, but:
* There's no need to keep **_policies** around as it's only used in the .ctor.
* Would it be possible to receive **policies** "pre-sorted" instead of having to sort them each time a **DfaMatcherBuilder** is instantiated?
* Would it be better to construct the **_nodeBuilders** and **_comparer** arrays (or use Lists instead) manually instead of iterating over the policies twice (using Linq)?
  This allows us to eliminate the `ToArray()` on the ordered policies.

If you want, I'll test whether this actually reduces allocations and/or improves performance later this week.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
rynowak: ah I see, yeah its fine as is

PR #768, opened 2018-09-03
Optimize builder and HTTP policy allocations

JamesNK:
Before:
```
    Dfa | 324.0 ms | 6.375 ms | 14.52 ms | 3.087 | 562.5000 | 500.0000 | 375.0000 | 130.02 MB |
```

After:
```
    Dfa | 279.5 ms | 5.582 ms | 7.452 ms | 3.578 | 687.5000 | 500.0000 | 312.5000 | 115.25 MB |
```

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
@@ -343,16 +343,16 @@ namespace Microsoft.AspNetCore.Routing.Matching
 
             if (node.Literals != null)
             {
+                var entries = new (string text, int destination)[node.Literals.Count];
+
+                var index = 0;
                 foreach (var kvp in node.Literals)
                 {
-                    if (kvp.Key == null)
-                    {
-                        continue;
-                    }
-
                     var transition = Transition(kvp.Value);
-                    pathBuilder.AddEntry(kvp.Key, transition);
+                    entries[index++] = (kvp.Key, transition);
                 }
+
+                pathBuilder.AddEntries(entries);
             }
 
             if (node.Parameters != null &&
@@ -388,10 +388,15 @@ namespace Microsoft.AspNetCore.Routing.Matching
                 var policyBuilder = new PolicyJumpTableBuilder(node.NodeBuilder);
                 tableBuilders[currentStateIndex] = (pathBuilder, policyBuilder);
 
+                var entries = new PolicyJumpTableEdge[node.PolicyEdges.Count];
+
+                var index = 0;
                 foreach (var kvp in node.PolicyEdges)
                 {
-                    policyBuilder.AddEntry(kvp.Key, Transition(kvp.Value));
+                    entries[index++] = new PolicyJumpTableEdge(kvp.Key, Transition(kvp.Value));
                 }
+
+                policyBuilder.AddEntries(entries);
             }
 
             return currentStateIndex;
--- a/src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
@@ -80,7 +80,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
             //
             // For now we're just building up the set of keys. We don't add any endpoints
             // to lists now because we don't want ordering problems.
-            var allHttpMethods = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var allHttpMethods = new List<string>();
             var edges = new Dictionary<EdgeKey, List<Endpoint>>();
             for (var i = 0; i < endpoints.Count; i++)
             {
@@ -120,10 +120,15 @@ namespace Microsoft.AspNetCore.Routing.Matching
                     // Also if it's not the *any* method key, then track it.
                     if (!string.Equals(AnyMethod, httpMethod, StringComparison.OrdinalIgnoreCase))
                     {
-                        allHttpMethods.Add(httpMethod);
+                        if (!ContainsHttpMethod(allHttpMethods, httpMethod))
+                        {
+                            allHttpMethods.Add(httpMethod);
+                        }
                     }
                 }
             }
+
+            allHttpMethods.Sort(StringComparer.OrdinalIgnoreCase);
 
             // Now in a second loop, add endpoints to these lists. We've enumerated all of
             // the states, so we want to see which states this endpoint matches.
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -261,6 +266,19 @@ private Endpoint CreateRejectionEndpoint(IEnumerable<string> httpMethods)
                 Http405EndpointDisplayName);
         }
 
+        public static bool ContainsHttpMethod(List<string> httpMethods, string httpMethod)
+        {
+            for (int i = 0; i < httpMethods.Count; i++)
```
drieseng: Wouldn't it be better to either use a foreach or introduce a local for the `httpMethods.Count`?
#askben

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -261,6 +266,19 @@ private Endpoint CreateRejectionEndpoint(IEnumerable<string> httpMethods)
                 Http405EndpointDisplayName);
         }
 
+        public static bool ContainsHttpMethod(List<string> httpMethods, string httpMethod)
```
drieseng: Does this method have to be public?

src/Microsoft.AspNetCore.Routing/Matching/PolicyJumpTableBuilder.cs
```diff
@@ -8,20 +8,19 @@ namespace Microsoft.AspNetCore.Routing.Matching
...
         {
             _nodeBuilder = nodeBuilder;
-            _entries = new List<PolicyJumpTableEdge>();
         }
 
         // The destination state for a non-match.
         public int ExitDestination { get; set; } = JumpTableBuilder.InvalidDestination;
 
-        public void AddEntry(object state, int destination)
+        public void AddEntries(PolicyJumpTableEdge[] entries)
```
drieseng: One might question if this class is still useful in this form. See also remarks on its usage in **DfaMatcherBuilder**.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -388,10 +388,15 @@ public override Matcher Build()
...
+                var entries = new PolicyJumpTableEdge[node.PolicyEdges.Count];
+
+                var index = 0;
                 foreach (var kvp in node.PolicyEdges)
                 {
-                    policyBuilder.AddEntry(kvp.Key, Transition(kvp.Value));
+                    entries[index++] = new PolicyJumpTableEdge(kvp.Key, Transition(kvp.Value));
                 }
+
+                policyBuilder.AddEntries(entries);
```
drieseng: Consider using the node builder directly (i.e. no longer use the **PolciyJumpTableBuilder**)?

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
JamesNK: It shouldn't be public

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -261,6 +266,19 @@ private Endpoint CreateRejectionEndpoint(IEnumerable<string> httpMethods)
...
+            for (int i = 0; i < httpMethods.Count; i++)
+            {
+                if (StringComparer.OrdinalIgnoreCase.Equals(httpMethods[i], httpMethod))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
```
rynowak: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.binarysearch?view=netframework-4.7.2

JamesNK: That would require the list to be sorted, and this is used while building the list, so it would either need to be resorted after each addition, or added to the right place in the list. I'm not sure it is worth the complexity for what will always be a small list.

JamesNK:
🆙 📅

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
rynowak: It just makes me a little sad to see somethign like this in in the BCL :(

src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
```diff
@@ -19,12 +21,18 @@ internal class JumpTableBuilder
...
-            _entries.Add((text, destination));
+            _pathEntries = entries;
         }
 
-        public JumpTable Build()
+        public void SetPolicyEntries(INodeBuilderPolicy nodeBuilder, PolicyJumpTableEdge[] entries)
+        {
+            _nodeBuilder = nodeBuilder;
+            _policyEntries = entries;
+        }
```
rynowak: Ok this has definitely got wierd. What's the value of this vs just making `BuildPath` a static method?

JamesNK: Static where? I put these fields on JumpTableBuilder because we need both the policy edges plus the node builder, and it felt cleaner to have here and have an array of JumpTableBuilders than have a tuple array of  JTB, node builder and policy edges.

rynowak: Ah I see.

I super don't get why these have become a single class :(

JamesNK: What about getting rid of JumpTableBuilder altogether? I suspect you mostly did it because you didn't know the exit point until all DfaNodes had been created. Now that the exit is known beforehand the JumpTable and PolicyJumpTable could be created as we go.

src/Microsoft.AspNetCore.Routing/Matching/HttpMethodMatcherPolicy.cs
rynowak: Btw if you want a really nice trick for this... https://github.com/aspnet/Mvc/blob/a67d9363e22be8ef63a1a62539991e1da3a6e30e/src/Microsoft.AspNetCore.Mvc.Core/Formatters/TextOutputFormatter.cs#L251

List.BinarySearch is best API ever. I noticed that you are sorting `allHttpMethods` eventually anyway.

JamesNK:
Removing jump table builders and their array knocked another 5 megs off allocations 🍰

PR #769, opened 2018-09-03
Add flag for including label with DfaNodes

JamesNK:
Before:
```
    Dfa | 279.5 ms | 5.582 ms | 7.452 ms | 3.578 | 687.5000 | 500.0000 | 312.5000 | 115.25 MB |
```

After:
```
    Dfa | 243.7 ms | 4.873 ms | 7.441 ms | 4.103 | 500.0000 | 375.0000 | 250.0000 |  88.25 MB |
```

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Internal/DfaGraphWriter.cs
+++ b/src/Microsoft.AspNetCore.Routing/Internal/DfaGraphWriter.cs
@@ -47,7 +47,7 @@ namespace Microsoft.AspNetCore.Routing.Internal
             // Assign each node a sequential index.
             var visited = new Dictionary<DfaNode, int>();
             
-            var tree = builder.BuildDfaTree();
+            var tree = builder.BuildDfaTree(includeLabel: true);
 
             writer.WriteLine("digraph DFA {");
             tree.Visit(WriteNode);
--- a/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
@@ -273,7 +273,13 @@ namespace Microsoft.AspNetCore.Routing.Matching
 
         public override Matcher Build()
         {
-            var root = BuildDfaTree();
+#if DEBUG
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Routing/Internal/DfaGraphWriter.cs
```diff
@@ -47,7 +47,7 @@ public void Write(EndpointDataSource dataSource, TextWriter writer)
             // Assign each node a sequential index.
             var visited = new Dictionary<DfaNode, int>();
             
-            var tree = builder.BuildDfaTree();
+            var tree = builder.BuildDfaTree(true);
```
rynowak: named param?

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -273,7 +273,13 @@ private RoutePatternPathSegment GetCurrentSegment(RouteEndpoint endpoint, int de
 
         public override Matcher Build()
         {
-            var root = BuildDfaTree();
+#if DEBUG
+            var includeLabel = true;
+#else
+            var includeLabel = false;
+#endif
```
rynowak: noice

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -219,7 +219,7 @@ public DfaNode BuildDfaTree()
                                 parent.Parameters = new DfaNode()
                                 {
                                     PathDepth = parent.PathDepth + 1,
-                                    Label = parent.Label + "{...}/",
+                                    Label = includeLabel ? parent.Label + "{...}/" : null,
```
rynowak: Another way to approach this would be to include a `Parent` link - then it would be possible to compute the pathdepth and label when needed.

JamesNK: I considered that but there are some places where three pieces of data are used to build the label.

PR #770, opened 2018-09-04
Revisions to LinkGenerator

rynowak:


benchmarks/Microsoft.AspNetCore.Routing.Performance/LinkGeneration/LinkGenerationGithubBenchmark.cs
```diff
@@ -57,9 +57,10 @@ public void TreeRouter()
         [Benchmark]
         public void EndpointRouting()
         {
-            var actualUrl = _linkGenerator.GetLink(
+            var actualUrl = _linkGenerator.GetPathByRouteValues(
                 _requestContext.HttpContext,
-                values: new RouteValueDictionary(_lookUpValues));
+                routeName: null,
```
JamesNK: Isn't getting a path by values alone fairly common? What do you think of adding GetPathByRouteValues overloads without name?

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
```diff
@@ -1,29 +1,90 @@
...
+        /// <param name="values">The route values. Used to expand parameters in the route template. Optional.</param>
+        /// <param name="scheme">The URI scheme, applied to the resulting URI.</param>
+        /// <param name="host">The URI host/authority, applied to the resulting URI.</param>
+        /// <param name="pathBase">An optional URI path base. Prepended to the path in the resulting URI.</param>
+        /// <param name="fragment">An optional URI fragment. Appended to the resulting URI.</param>
+        /// <param name="options">
+        /// An optional <see cref="LinkOptions"/>. Settings on provided object override the settings with matching
+        /// names from <c>RouteOptions</c>.
+        /// </param>
+        /// <returns>An abosolute URI, or <c>null</c>.</returns>
```
JamesNK: absolute

src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -35,200 +38,249 @@ internal class DefaultLinkGenerator : LinkGenerator
...
+            FragmentString fragment,
+            LinkOptions options)
+        {
+            for (var i = 0; i < endpoints.Count; i++)
             {
-                link = MakeLink(
-                    httpContext,
+                var endpoint = endpoints[i];
+                if (TryProcessTemplate(
+                    httpContext: null,
```
JamesNK: indent++

src/Microsoft.AspNetCore.Routing/IRouteValuesAddressMetadata.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.AspNetCore.Routing
 {
     public interface IRouteValuesAddressMetadata
     {
-        string Name { get; }
+        string RouteName { get; }
```
JamesNK: Will need a feature branch for MVC

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Http;
```
JamesNK: order

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
```diff
@@ -68,6 +80,12 @@ public bool Accept(string value, bool encodeSlashes)
                 return false;
             }
 
+            // NOTE: this needs to be above all 'EncodeValue' and _path.Append calls
+            if (LowercaseUrls)
```
JamesNK: 👍 

Is there a unit test for this? (I'm sure there is, I missed seeing it)

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
```diff
@@ -27,20 +29,30 @@ public class UriBuildingContext
         public UriBuildingContext(UrlEncoder urlEncoder)
         {
             _urlEncoder = urlEncoder;
-            _uri = new StringBuilder();
+            _path = new StringBuilder();
+            _query = new StringBuilder();
             _buffer = new List<BufferValue>();
-            Writer = new StringWriter(_uri);
+            PathWriter = new StringWriter(_path);
+            QueryWriter = new StringWriter(_query);
```
JamesNK: Are the path and query written out of order? Is it not possible to get away with one builder & writer and keep an index of when the path stops and the query begins?

src/Microsoft.AspNetCore.Routing/LinkGeneratorRouteValuesAddressExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+            var address = CreateAddress(httpContext: null, routeName, values);
+            return generator.GetUriByAddress<RouteValuesAddress>(address, address.ExplicitValues, scheme, host, pathBase, fragment, options);
+        }
+
+
+        private static RouteValuesAddress CreateAddress(HttpContext httpContext, string routeName, object values)
+        {
+            return new RouteValuesAddress()
+            {
+                AmbientValues = httpContext?.GetRouteData()?.Values,
```
JamesNK: I've been wondering about `GetRouteData` and whether it should be considered legacy. I needed exactly `RouteData` to pass to MVC in `MvcEndpointDataSource` so I thought GetRouteData was acceptable but for getting just values I think we should use IRouteValuesFeature.

src/Microsoft.AspNetCore.Routing/Template/TemplateBinder.cs
```diff
@@ -1,16 +1,16 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing.Internal;
+using Microsoft.AspNetCore.Routing.Patterns;
+using Microsoft.Extensions.ObjectPool;
```
JamesNK: order

Is VS doing this for you automatically? I changed a setting to tell it to place System first and the refactoring tools followed suit.

test/Microsoft.AspNetCore.Routing.Tests/LinkGeneratorTestBase.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Routing.Internal;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.ObjectPool;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public abstract class LinkGeneratorTestBase
```
JamesNK: 👍

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
JamesNK: That's true, but do need to ToString them separately though, and then combine the result.

rynowak: That's intended though (at least for now) so that we can use `UriHelper` instead of creating our own thing.

rynowak:
Updated

PR #787, opened 2018-09-07
Fix #553 - Add logging for new LinkGenerator

### LINKED ISSUE #553: Dispatcher link generation: Improve diagnostics
- log when no endpoint was found.
- log when we have an endpoint but parameters and constraints are not satisfied.
- Maybe throw an exception when link generation fails as an option?

rynowak:
Adds logging for all of the failure cases of link generation.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
+++ b/src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using System.Text.Encodings.Web;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Routing/DefaultLinkGenerator.cs
```diff
@@ -317,5 +338,119 @@ public static RouteValueDictionary GetAmbientValues(HttpContext httpContext)
...
+                public static readonly EventId TemplateFailedExpansion = new EventId(104, "TemplateFailedExpansion");
+
+                public static readonly EventId LinkGenerationSucceeded = new EventId(105, "LinkGenerationSucceeded");
+                public static readonly EventId LinkGenerationFailed = new EventId(106, "LinkGenerationFailed");
+            }
+
+            private static readonly Action<ILogger, IEnumerable<string>, object, Exception> _endpointsFound = LoggerMessage.Define<IEnumerable<string>, object>(
+                LogLevel.Debug,
+                EventIds.EndpointsFound,
+                "Found the endpoints '{Endpoints}' for address '{Address}'");
```
JamesNK: nit: `IEnumerable<string>` is serialized as a comma separated list I am guessing. It would be better if each name was quoted rather than the combined string.

Applicable to all collections in these log messages.

rynowak: Yeah I think you're right about that. Will revise it.

rynowak: I just removed the quoting from this. We don't generally quote things in logs anyway. I added formatting for the route values so that those would show up in a reasonable way.

PR #799, opened 2018-09-18
Endpoing routing 3.0 registration

JamesNK:
EndpointCon 2018 feedback


src/Microsoft.AspNetCore.Routing/EndpointDataSourceBuilder.cs
```diff
@@ -1,15 +1,18 @@
...
-    public abstract class EndpointDataSourceBuilder
+    public abstract class EndpointDataSourcesBuilder
     {
-        public abstract ICollection<EndpointBuilder> Endpoints { get; }
-
         public abstract IApplicationBuilder CreateApplicationBuilder();
+
+        public abstract IServiceProvider ServiceProvider { get; }
+
+        public abstract IList<EndpointDataSource> EndpointDataSources { get; }
```
JamesNK: What should change on `EndpointDataSourcesBuilder` to become `IEndpointRoutesBuilder`?

With calls like `Map`, `MapGet`, etc, the user is creating a single endpoint and there is no consideration for that endpoint ever changing.

Meanwhile a call like `MapMvc("{controller}/{action}")` will configure a datasource that has a change token.

My idea: Perhaps these methods are populated onto a single static datasource.

```cs
public interface IEndpointRoutesBuilder
{
    IApplicationBuilder CreateApplicationBuilder();
    IServiceProvider ServiceProvider { get; }
    IList<EndpointDataSource> EndpointDataSources { get; }
    IList<Endpoint> StaticEndpoints { get; }
}
```

src/Microsoft.AspNetCore.Routing/IApplyEndpointBuilder.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public interface IApplyEndpointBuilder
```
JamesNK: PLEASE 👏 CHANGE 👏 THIS 👏 NAME

samples/RoutingSample.Web/AuthorizationMiddleware/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,59 @@
...
+                if (metadata != null)
+                {
+                    if (!httpContext.Request.Headers.TryGetValue("x-role", out var role) ||
+                        !metadata.Roles.Contains(role.ToString()))
+                    {
+                        httpContext.Response.StatusCode = 401;
+                        httpContext.Response.ContentType = "text/plain";
+                        await httpContext.Response.WriteAsync($"Unauthorized access to '{endpoint.DisplayName}'.");
+                        return;
+                    }
```
rynowak: Good enough for a fake 👍

samples/RoutingSample.Web/HelloExtension/EndpointDataSourceBuilderExtensions.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.Builder
 {
     public static class EndpointDataSourceBuilderExtensions
     {
-        public static EndpointBuilder MapHello(this EndpointDataSourceBuilder builder, string template, string greeter)
+        public static IApplyEndpointBuilder MapHello(this EndpointDataSourcesBuilder builder, string template, string greeter)
```
rynowak: `EndpointDataSourcesBuilder` -> `IEndpointRouteBuilder`

src/Microsoft.AspNetCore.Routing/EndpointDataSourceBuilder.cs
davidfowl: What is CreateApplicationBuilder for?

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
```diff
@@ -90,11 +99,9 @@ public void Configure(IApplicationBuilder app)
...
 
                         return Task.CompletedTask;
-                    },
-                    "/graph",
-                    "DFA Graph",
-                    new HttpMethodMetadata(new[] { "GET", }));
-                builder.MapEndpoint(
+                    });
+                builder.MapGet(
+                    "/WithSingleAsteriskCatchAll/{*path}",
```
rynowak: What would you think about trimming down the set of things that are in this class. They aren't super meaningful, and I'm OK with just deleting the functional tests that cover them.

src/Microsoft.AspNetCore.Routing/Builder/EndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,168 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class EndpointDataSourceBuilderExtensions
```
rynowak: `EndpointRouteBuilderExtensions`?

src/Microsoft.AspNetCore.Routing/Builder/EndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,168 @@
...
+    {
+        #region Http
+        public static IApplyEndpointBuilder MapGet(
```
rynowak: What if we did the following:
```
EndpointBuilder -> EndpointModel
IApplyEndpointBuilder -> IEndpointModel
```

src/Microsoft.AspNetCore.Routing/Builder/EndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        public static IApplyEndpointBuilder MapDelete(
+            this EndpointDataSourcesBuilder builder,
+            string pattern,
+            string displayName,
+            RequestDelegate requestDelegate,
+            params object[] metadata)
+        {
+            return MapHttp(builder, pattern, displayName, requestDelegate, new[] { "DELETE" }, metadata);
+        }
```
rynowak: I think this is a good set to start with

src/Microsoft.AspNetCore.Routing/Builder/EndpointDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        private static IApplyEndpointBuilder MapHttp(
```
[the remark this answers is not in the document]
rynowak: Yeah, I'd suggest `MapVerbs`

src/Microsoft.AspNetCore.Routing/EndpointDataSourceBuilder.cs
rynowak: Perhaps I'm missing something. What would go wrong if `Map` used the same technique as `MapMvc...`? Create and store a single data source, and then add stuff to it repeatedly?

src/Microsoft.AspNetCore.Routing/Builder/EndpointDataSourceBuilderExtensions.cs
JamesNK: Hmm, I would expect EndpointModel to implement IEndpointModel, and the members to be the same.

src/Microsoft.AspNetCore.Routing/EndpointDataSourceBuilder.cs
JamesNK: @davidfowl So you can easily turn existing middleware into an endpoint

```cs
public static class EndpointDataSourceBuilderExtensions
{
    public static IApplyEndpointBuilder MapHello(this EndpointDataSourcesBuilder builder, string template, string greeter)
    {
        if (builder == null)
        {
            throw new ArgumentNullException(nameof(builder));
        }

        var pipeline = builder.CreateApplicationBuilder()
           .UseHello(greeter)
           .Build();

        return builder.Map(
            template,
            "Hello " + greeter,
            pipeline);
    }
}
```

Good? Bad? Better way of doing it?

JamesNK: @rynowak There is nothing wrong with doing that. To me it feels weird that this builder has a list of datasources on it, yet we aren't using it. We're just using ServiceProvider and building datasources that have been registered in DI.

Maybe MvcEndpointDataSource shouldn't be registered in DI but created and stored on this collection (i.e. get it from the list, if it is null then create the singleton instance).
Use the same approach for "static" endpoints.

samples/RoutingSample.Web/HelloExtension/EndpointDataSourceBuilderExtensions.cs
JamesNK: Yes

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
JamesNK: I'd like to keep them. These functional tests helped me spot a bug yesterday.

What about splitting this site into a site for functional tests and a sandpit for experiments?

src/Microsoft.AspNetCore.Routing/Builder/EndpointDataSourceBuilderExtensions.cs
JamesNK: Yes

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
JamesNK: Splitting the site can wait for another PR

src/Microsoft.AspNetCore.Routing/DefaultEndpointRouteBuilder.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal class DefaultEndpointRouteBuilder : IEndpointRouteBuilder
+    {
+        public DefaultEndpointRouteBuilder()
+        {
+            EndpointDataSources = new List<EndpointDataSource>();
+        }
+
+        public IApplicationBuilder ApplicationBuilder { get; set; }
+
+        public IApplicationBuilder CreateApplicationBuilder() => ApplicationBuilder.New();
```
JamesNK: @davidfowl re earlier question about this: So you can easily turn existing middleware into an endpoint

```cs
public static class EndpointDataSourceBuilderExtensions
{
    public static IApplyEndpointBuilder MapHello(this EndpointDataSourcesBuilder builder, string template, string greeter)
    {
        if (builder == null)
        {
            throw new ArgumentNullException(nameof(builder));
        }

        var pipeline = builder.CreateApplicationBuilder()
           .UseHello(greeter)
           .Build();

        return builder.Map(
            template,
            "Hello " + greeter,
            pipeline);
    }
}
```

Good? Bad? Better way of doing it?

src/Microsoft.AspNetCore.Routing/Builder/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,189 @@
...
+            routeEndpointModel.DisplayName = displayName;
+            if (metadata != null)
+            {
+                foreach (var item in metadata)
+                {
+                    routeEndpointModel.Metadata.Add(item);
+                }
+            }
+
+            var modelEndpointDataSource = builder.EndpointDataSources.OfType<ModelEndpointDataSource>().FirstOrDefault();
```
JamesNK: Storing the data source here instead of in DI. Can do the same for the MvcEndpointDataSource.

JamesNK:
🆙 📅

src/Microsoft.AspNetCore.Routing/DefaultEndpointRouteBuilder.cs
rynowak: > Good? ~~Bad?~~ Better way of doing it?

This is elegant and cool. Just like `Map` for middleware but it uses routing. Any middleware or middware pipeline can be 'endpointed'

samples/RoutingSample.Web/UseEndpointRoutingStartup.cs
rynowak: Yeah that sounds good to me. I'd really prefer that the sample evolve based on what we're working on, or what's important for users to see. That requires us to not try and keep stuff just because it was valuable at some point 👍 functional test are that.

src/Microsoft.AspNetCore.Routing/Builder/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,189 @@
...
+           RequestDelegate requestDelegate,
+           IList<string> httpMethods,
+           params object[] metadata)
+        {
+            if (httpMethods == null)
+            {
+                throw new ArgumentNullException(nameof(httpMethods));
+            }
+
+            List<object> resolvedMetadata = new List<object>();
```
rynowak: var

src/Microsoft.AspNetCore.Routing/Builder/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public static IEndpointConventionBuilder Map(
+            this IEndpointRouteBuilder builder,
+            RoutePattern pattern,
+            string displayName,
+            RequestDelegate requestDelegate,
+            params object[] metadata)
+        {
+            const int defaultOrder = 0;
```
rynowak: It seems like more stuff should be getting null checked.

src/Microsoft.AspNetCore.Routing/Builder/EndpointRouteBuilderExtensions.cs
rynowak: I think from a scenario point of view.
1. I should be able to resolve **something** from DI taht can find all of the endpoints
2. I should be able to add endpoints via startup
3. I should be able to add endpoints via DI (frameworks)

As long as these aren't violated, that sounds good.

rynowak: Another alternate I see would be for the composite data source to be mutable, and have this API wrap/expose it. Then all of these things are the same.

src/Microsoft.AspNetCore.Routing/IEndpointModelConvention.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public interface IEndpointConventionBuilder
```
rynowak: Yeah this is a much better name.

test/Microsoft.AspNetCore.Routing.Tests/LinkGeneratorTestBase.cs
```diff
@@ -77,16 +77,17 @@ private protected DefaultLinkGenerator CreateLinkGenerator(EndpointDataSource[]
...
             dataSources = dataSources ?? Array.Empty<EndpointDataSource>();
 
-            services.Configure<EndpointOptions>((o) =>
+            var options = Options.Create(routeOptions);
+            var serviceProvider = services.BuildServiceProvider();
+
+            if (dataSources != null || dataSources.Length > 0)
             {
-                for (var i = 0; i < dataSources.Length; i++)
+                var compositeEndpointDataSource = serviceProvider.GetRequiredService<CompositeEndpointDataSource>();
```
JamesNK: I think there is an example of where there is value in an options type. To add a data source to the `CompositeEndpointDataSource` we need to fetch it from a service provider and then modify it. The optons `services.Configure<EndpointOptions>((o) => { ... })` syntax feels better.

Options could just expose the `IList<EndpointDataSource>` found on `CompositeEndpointDataSource`.

rynowak:
I am not aware of a known issue causing that error. My guess is that something is causing a link to fail to generate (yeah real helpful I know).

JamesNK:
This is what happens when I try to run the benchmarks in this branch and in master:

```
// Build Exception:   Restoring packages for C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2
.2\b9d8f342-923c-475e-ba86-b91c24343556\BenchmarkDotNet.Autogenerated.csproj...
C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2.2\b9d8f342-923c-475e-ba86-b91c24343556\Bench
markDotNet.Autogenerated.csproj : error NU1102: Unable to find package Microsoft.NETCore.App with version (>= 3.0.0)
C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2.2\b9d8f342-923c-475e-ba86-b91c24343556\Bench
markDotNet.Autogenerated.csproj : error NU1102:   - Found 971 version(s) in https://dotnet.myget.org/F/dotnet-core/api/v3/index.json [ Nearest version
: 3.0.0-preview1-26923-01 ]
C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2.2\b9d8f342-923c-475e-ba86-b91c24343556\Bench
markDotNet.Autogenerated.csproj : error NU1102:   - Found 42 version(s) in nuget.org [ Nearest version: 2.2.0-preview2-26905-02 ]
C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2.2\b9d8f342-923c-475e-ba86-b91c24343556\Bench
markDotNet.Autogenerated.csproj : error NU1102:   - Found 40 version(s) in https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json [ Nearest versi
on: 2.1.0-preview2-26308-01 ]
C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2.2\b9d8f342-923c-475e-ba86-b91c24343556\Bench
markDotNet.Autogenerated.csproj : error NU1102:   - Found 0 version(s) in https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json
  Generating MSBuild file C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2.2\b9d8f342-923c-47
5e-ba86-b91c24343556\obj\BenchmarkDotNet.Autogenerated.csproj.nuget.g.props.
  Generating MSBuild file C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2.2\b9d8f342-923c-47
5e-ba86-b91c24343556\obj\BenchmarkDotNet.Autogenerated.csproj.nuget.g.targets.
  Restore failed in 3.71 sec for C:\Development\Source\Routing\benchmarks\Microsoft.AspNetCore.Routing.Performance\bin\Release\netcoreapp2.2\b9d8f342-
923c-475e-ba86-b91c24343556\BenchmarkDotNet.Autogenerated.csproj.
```

samples/RoutingSample.Web/AuthorizationMiddleware/ApplyEndpointExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Routing;
+using RoutingSample.Web.AuthorizationMiddleware;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class ApplyEndpointExtensions
```
rynowak: ` EndpointConventionBuilderExtensions` 👍

src/Microsoft.AspNetCore.Routing/Builder/EndpointRoutingApplicationBuilderExtensions.cs
```diff
@@ -18,20 +18,39 @@ public static IApplicationBuilder UseEndpointRouting(this IApplicationBuilder bu
...
+                var endpointRouteBuilder = builder.ApplicationServices.GetRequiredService<IEndpointRouteBuilder>();
+                if (endpointRouteBuilder is DefaultEndpointRouteBuilder defaultEndpointRouteBuilder)
+                {
+                    defaultEndpointRouteBuilder.ApplicationBuilder = builder;
+                }
+                configure(endpointRouteBuilder);
+
+                foreach (var dataSource in endpointRouteBuilder.DataSources)
+                {
+                    // REVIEW: Should there be a way to add multiple datasources at once to avoid potentially re-triggering change token multiple times?
```
rynowak: Well the way to do that would be to have an 'AddRange` method. I don't really see it as a problem because I don't really think anyone is listening to the change tokens yet. Everything gets initialized on the first request (or first link generation).

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -18,24 +20,44 @@ namespace Microsoft.AspNetCore.Routing
     [DebuggerDisplay("{DebuggerDisplayString,nq}")]
     public sealed class CompositeEndpointDataSource : EndpointDataSource
     {
-        private readonly EndpointDataSource[] _dataSources;
         private readonly object _lock;
+        private readonly ObservableCollection<EndpointDataSource> _dataSources;
```
rynowak: lol!

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -18,24 +20,44 @@ namespace Microsoft.AspNetCore.Routing
     [DebuggerDisplay("{DebuggerDisplayString,nq}")]
     public sealed class CompositeEndpointDataSource : EndpointDataSource
```
rynowak: Is it possible to make this internal at this point? or will that require a bunch more things to change?

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -18,24 +20,44 @@ namespace Microsoft.AspNetCore.Routing
...
+            {
+                // Only trigger changes if composite data source has already initialized endpoints
+                if (_endpoints != null)
+                {
+                    HandleChange();
+                }
+            }
+        }
+
+        public IList<EndpointDataSource> DataSources => _dataSources;
```
rynowak: Consider instead making the composite Implement `ICollection` - if this ends up ever going public, that just feels a little more first class.

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
JamesNK: If it was made internal then there would not be a way to register new data sources outside of UseEndpointData. And there would be no way to get all data sources, endpoints, or a change token. I think it is a useful type, both the link generator and the matcher are using it.

JamesNK:
I'm not a fan of being able to add and remove data sources at run-time to `CompositeDataSource`. Having the list of data sources be changable and then individual data sources also be changable feels like overlap and it is too dynamic.

Does it make sense to add and remove data sources after `ConfigureServices`? I wonder if data source registration should happen only in `ConfigureServices`, e.g. `AddMvc` would add the `MvcEndpointDataSource` and then `Configure` would be setting up configuration on the data source.

If we do that then `CompositeDataSource` goes back to taking an `Enumerable<EndpointDataSource>`.

src/Microsoft.AspNetCore.Routing/DependencyInjection/RoutingServiceCollectionExtensions.cs
```diff
@@ -49,21 +49,17 @@ public static IServiceCollection AddRouting(this IServiceCollection services)
...
-            // Collect all data sources from DI.
-            services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<EndpointOptions>, ConfigureEndpointOptions>());
-
             // Allow global access to the list of endpoints.
             services.TryAddSingleton<CompositeEndpointDataSource>(s =>
             {
-                var options = s.GetRequiredService<IOptions<EndpointOptions>>();
-                return new CompositeEndpointDataSource(options.Value.DataSources);
+                // Call internal ctor
+                return new CompositeEndpointDataSource();
```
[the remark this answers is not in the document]
rynowak: There will be two in the default configuration. 1 global one, and one created by `UseEndpointPointRouting(Action<...>)`. This means that all of the work MVC does, for instance, will happen twice.

rynowak:
> Does it make sense to add and remove data sources after ConfigureServices? I wonder if data source registration should happen only in ConfigureServices, e.g. AddMvc would add the MvcEndpointDataSource and then Configure would be setting up configuration on the data source.

If this is the case then we can't allow multiple instance of the middleware. Each usage of `UseEndpointRouting` would need to create its own data source right?

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
rynowak: well, the discussion on Friday was about how we will satisfy the following requirements:
1. A way to get all endpoints for various information things
2. A way to get all endpoints + change notifications for link generation (can be internal)
3. A way to add data sources (can be through a builder)

I still think this this is the best API for all of those things and it should just be public.

JamesNK:
🆙 📅 

Hmm, for some reason now it now builds, both locally and on the server 🤷‍♂️ 

I added `RouteOptions.EndpointDataSources` and changed the registered composite to be registered as `EndpointDataSource`. There is nothing special about `CompositeEndpointDataSource` anymore. It could be made internal, but it might be useful to keep in public as something for developers to combine data sources.

src/Microsoft.AspNetCore.Routing/Builder/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,210 @@
...
+            return MapVerbs(builder, pattern, displayName: null, requestDelegate, httpMethods, metadata);
+        }
+
+        public static IEndpointConventionBuilder MapVerbs(
+           this IEndpointRouteBuilder builder,
+           string pattern,
+           string displayName,
+           RequestDelegate requestDelegate,
+           IList<string> httpMethods,
+           params object[] metadata)
```
davidfowl: Why do we have metadata passed in here? Why not instead just make it so that you can add it via `IEndpointConventionBuilder`?

src/Microsoft.AspNetCore.Routing/Builder/EndpointRoutingApplicationBuilderExtensions.cs
```diff
@@ -18,20 +19,39 @@ public static IApplicationBuilder UseEndpointRouting(this IApplicationBuilder bu
...
+
             if (configure != null)
             {
-                var dataSourceBuilder = (DefaultEndpointDataSourceBuilder)builder.ApplicationServices.GetRequiredService<EndpointDataSourceBuilder>();
-                dataSourceBuilder.ApplicationBuilder = builder;
-                configure(dataSourceBuilder);
+                var endpointRouteBuilder = builder.ApplicationServices.GetRequiredService<IEndpointRouteBuilder>();
+                if (endpointRouteBuilder is DefaultEndpointRouteBuilder defaultEndpointRouteBuilder)
+                {
+                    defaultEndpointRouteBuilder.ApplicationBuilder = builder;
```
davidfowl: ?

src/Microsoft.AspNetCore.Routing/Builder/EndpointRouteBuilderExtensions.cs
JamesNK: Sure metadata could be in an extension method. Do we want it to be an extension method?

src/Microsoft.AspNetCore.Routing/Builder/EndpointRoutingApplicationBuilderExtensions.cs
JamesNK: Yeah it is a little wierd. ApplicationBuilder isn't settable on the interface.

rynowak: The cleaner way to do this would be to DI the application builder factory. A cut a corner building this originally because it was for a demo.

JamesNK:
🆙 📅 

Any thing let to change? Would be good to merge

JamesNK:
🆙 📅 

It occured to me that there is a hole around how MVC attribute and razor page routing light up. It was implicit with `UseMvc` that attributes and razor pages would just work.

With `IEndpointRouteBuilder` I think either it could be:
1. An explicit method, e.g. `builder.MapMvcRazorPages()`, or
2. There is something like UseMvc today, e.g. `builder.MapMvc(routes => /* config conventional routes here */ )`, and part of MapMvc would be to automatically light up attribute routing and razor pages.

src/Microsoft.AspNetCore.Routing/Builder/EndpointRoutingApplicationBuilderExtensions.cs
```diff
@@ -6,32 +6,44 @@
...
-                var dataSourceBuilder = (DefaultEndpointDataSourceBuilder)builder.ApplicationServices.GetRequiredService<EndpointDataSourceBuilder>();
-                dataSourceBuilder.ApplicationBuilder = builder;
-                configure(dataSourceBuilder);
+                defaultEndpointRouteBuilder.ApplicationBuilder = builder;
             }
+            configure(endpointRouteBuilder);
+
+            foreach (var dataSource in endpointRouteBuilder.DataSources)
+            {
+                routeOptions.Value.EndpointDataSources.Add(dataSource);
```
rynowak: I'm still a little hung up on mutating options inside `Configure` - did you and @davidfowl talk about this?

JamesNK:
I've added explicit methods for mapping attribute routing controllers and Razor pages. See https://github.com/aspnet/Mvc/pull/8470

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -18,24 +20,49 @@ namespace Microsoft.AspNetCore.Routing
...
+        }
 
-            CreateChangeToken();
-            _dataSources = dataSources.ToArray();
-            _lock = new object();
+        private void OnDataSourcesChanged(object sender, NotifyCollectionChangedEventArgs e)
+        {
+            lock (_lock)
+            {
+                // Only trigger changes if composite data source has already initialized endpoints
```
rynowak: MMM spicy. I like this.

PR #804, opened 2018-09-20
Make parameter transformer test verifiable

JamesNK:


test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorProcessTemplateTest.cs
```diff
@@ -306,7 +306,7 @@ public void TryProcessTemplate_GeneratesLowercaseUrl_Includes_BufferedValues_Set
         public void TryProcessTemplate_ParameterPolicy_Includes_BufferedValues()
         {
             // Arrange
-            var endpoint = EndpointFactory.CreateRouteEndpoint("Foo/{bar=BAR}/{id?}", policies: new { bar = new SlugifyParameterTransformer(), });
+            var endpoint = EndpointFactory.CreateRouteEndpoint("Foo/{bar=MyBar}/{id?}", policies: new { bar = new SlugifyParameterTransformer(), });
             var linkGenerator = CreateLinkGenerator(new RouteOptions() { LowercaseUrls = true }, endpoints: new[] { endpoint, });
```
rynowak: Could you also delete the lowercase URL thing? I meant to leave this out, since the parameter tranformer is supposed to be the think lowering the casing.

PR #811, opened 2018-09-25
Fix a bug in GetUriByRouteValues

rynowak:
Pride cometh before the fall...


src/Microsoft.AspNetCore.Routing/LinkGeneratorRouteValuesAddressExtensions.cs
```diff
@@ -135,7 +135,7 @@ public static class LinkGeneratorRouteValuesAddressExtensions
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            var address = CreateAddress(httpContext: null, routeName, values);
+            var address = CreateAddress(httpContext, routeName, values);
```
rynowak: 😲

Issue #797, opened 2018-09-17
Clean up tasks


JamesNK:
Misc small tasks that need doing:

- [x] xmldocs on IRouteValuesAddressMetadata and RouteValuesAddressMetadata
- [x] null checks on RouteValuesAddressMetadata ctor

PR #818, opened 2018-09-27
Check route values address metadata params not null

JamesNK:
https://github.com/aspnet/Routing/issues/797

We could add ctors for the different args:

```cs
RouteValuesAddressMetadata(string routeName)
RouteValuesAddressMetadata(IReadOnlyDictionary<string, object> requiredValues)
RouteValuesAddressMetadata(string routeName, IReadOnlyDictionary<string, object> requiredValues)
```

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/RouteValuesAddressMetadata.cs
+++ b/src/Microsoft.AspNetCore.Routing/RouteValuesAddressMetadata.cs
@@ -3,21 +3,45 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Linq;
 
 namespace Microsoft.AspNetCore.Routing
 {
+    /// <summary>
+    /// Represents metadata used during link generation to find
+    /// the associated endpoint using route values.
+    /// </summary>
     [DebuggerDisplay("{DebuggerToString(),nq}")]
     public sealed class RouteValuesAddressMetadata : IRouteValuesAddressMetadata
     {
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Routing/RouteValuesAddressMetadata.cs
```diff
@@ -3,22 +3,62 @@
...
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Linq;
 
 namespace Microsoft.AspNetCore.Routing
 {
+    /// <summary>
+    /// Represents metadata used during link generation to find
```
rynowak: ewww represents

test/Microsoft.AspNetCore.Routing.Tests/EndpointFactory.cs
```diff
@@ -22,7 +22,7 @@ internal static class EndpointFactory
             var d = new List<object>(metadata ?? Array.Empty<object>());
             if (requiredValues != null)
             {
-                d.Add(new RouteValuesAddressMetadata(null, new RouteValueDictionary(requiredValues)));
+                d.Add(new RouteValuesAddressMetadata("routeName!", new RouteValueDictionary(requiredValues)));
```
rynowak: Should we revert all of these changes? Route Name isn't something that's required, and it's not relevant to these tests. Additionally **this** part of the code applies a route name to a whole bunch of places where it wasn't asked for....

Generally conventionally routed endpoints have a route name, and attribute routed endpoints don't (but could).

PR #819, opened 2018-09-27
Fix #629 - Add logging to DfaMatcher

### LINKED ISSUE #629: Add logging to new features
- [x] Middleware
- [x] Link Generation Features
- [ ] DFA Matcher
- [ ] Others???

rynowak:
Adds logging for the most common things that can prevent an endpoint
from matching.

Note that we already have good logging in other parts of the system, the
stuff here completes the story by providing details at the debug level.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
@@ -6,17 +6,20 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing.Patterns;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Routing.Matching
 {
     internal sealed class DfaMatcher : Matcher
     {
+        private readonly ILogger _logger;
         private readonly EndpointSelector _selector;
         private readonly DfaState[] _states;
         private readonly int _maxSegmentCount;
         
-        public DfaMatcher(EndpointSelector selector, DfaState[] states, int maxSegmentCount)
+        public DfaMatcher(ILogger<DfaMatcher> logger, EndpointSelector selector, DfaState[] states, int maxSegmentCount)
         {
+            _logger = logger;
             _selector = selector;
             _states = states;
             _maxSegmentCount = maxSegmentCount;
@@ -34,6 +37,9 @@ namespace Microsoft.AspNetCore.Routing.Matching
                 throw new ArgumentNullException(nameof(context));
             }
 
+            // All of the logging we do here is at level debug, so we can get away with doing a single check.
+            var log = _logger.IsEnabled(LogLevel.Debug);
+
             // The sequence of actions we take is optimized to avoid doing expensive work
             // like creating substrings, creating route value dictionaries, and calling
             // into policies like versioning.
@@ -49,7 +55,17 @@ namespace Microsoft.AspNetCore.Routing.Matching
             var candidates = FindCandidateSet(httpContext, path, segments);
             if (candidates.Length == 0)
             {
+                if (log)
+                {
+                    Logger.CandidatesNotFound(_logger, path);
+                }
+
                 return Task.CompletedTask;
+            }
+
+            if (log)
+            {
+                Logger.CandidatesFound(_logger, path, candidates);
             }
 
             // At this point we have a candidate set, defined as a list of endpoints in
@@ -117,21 +133,34 @@ namespace Microsoft.AspNetCore.Routing.Matching
                 // Now that we have the route values, we need to process complex segments.
                 // Complex segments go through an old API that requires a fully-materialized
                 // route value dictionary.
+                var isMatch = true;
                 if ((flags & Candidate.CandidateFlags.HasComplexSegments) != 0)
                 {
-                    if (!ProcessComplexSegments(candidate.ComplexSegments, path, segments, values))
+                    if (!ProcessComplexSegments(candidate.Endpoint, candidate.ComplexSegments, path, segments, values))
                     {
                         candidateSet.SetValidity(i, false);
-                        continue;
+                        isMatch = false;
                     }
                 }
 
                 if ((flags & Candidate.CandidateFlags.HasConstraints) != 0)
                 {
-                    if (!ProcessConstraints(candidate.Constraints, httpContext, values))
+                    if (!ProcessConstraints(candidate.Endpoint, candidate.Constraints, httpContext, values))
                     {
                         candidateSet.SetValidity(i, false);
-                        continue;
+                        isMatch = false;
+                    }
+                }
+
+                if (log)
+                {
+                    if (isMatch)
+                    {
+                        Logger.CandidateValid(_logger, path, candidate.Endpoint);
+                    }
+                    else
+                    {
+                        Logger.CandidateNotValid(_logger, path, candidate.Endpoint);
                     }
                 }
             }
@@ -202,6 +231,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
         }
 
         private bool ProcessComplexSegments(
+            Endpoint endpoint,
             (RoutePatternPathSegment pathSegment, int segmentIndex)[] complexSegments,
             string path,
             ReadOnlySpan<PathSegment> segments,
@@ -209,10 +239,12 @@ namespace Microsoft.AspNetCore.Routing.Matching
         {
             for (var i = 0; i < complexSegments.Length; i++)
             {
-                var segment = segments[complexSegments[i].segmentIndex];
+                (var complexSegment, var segmentIndex) = complexSegments[i]; 
+                var segment = segments[segmentIndex];
                 var text = path.Substring(segment.Start, segment.Length);
-                if (!RoutePatternMatcher.MatchComplexSegment(complexSegments[i].pathSegment, text, values))
+                if (!RoutePatternMatcher.MatchComplexSegment(complexSegment, text, values))
                 {
+                    Logger.CandidateRejectedByComplexSegment(_logger, path, endpoint, complexSegment);
                     return false;
                 }
             }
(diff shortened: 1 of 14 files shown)


rynowak:
Sample:
```
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost:60045/withconstraints/3
dbug: Microsoft.AspNetCore.Routing.Matching.DfaMatcher[1000]
      1 candidate(s) found for the request path '/withconstraints/3'
dbug: Microsoft.AspNetCore.Routing.Matching.DfaMatcher[1002]
      Endpoint 'withconstraints' with route pattern '/withconstraints/{id:endsWith(_001)}' was rejected by constraint 'id' with value '3' for the request path '/withconstraints/3'
dbug: Microsoft.AspNetCore.Routing.Matching.DfaMatcher[1004]
      Endpoint 'withconstraints' with route pattern '/withconstraints/{id:endsWith(_001)}' is not valid for the request path '/withconstraints/3'
dbug: Microsoft.AspNetCore.Routing.EndpointRoutingMiddleware[2]
      Request did not match any endpoints.
dbug: Microsoft.AspNetCore.Server.Kestrel[9]
      Connection id "0HLH4MEKIOHDP" completed keep alive response.
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 13.8146ms 404
```

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
```diff
@@ -226,11 +258,100 @@ public sealed override Task MatchAsync(HttpContext httpContext, EndpointFeature
...
+            public static readonly EventId CandidateNotValid = new EventId(1004, "CandiateNotValid");
+            public static readonly EventId CandidateValid = new EventId(1005, "CandiateValid");
+        }
+
+        private static class Logger
+        {
+            private static readonly Action<ILogger, string, Exception> _candidatesNotFound = LoggerMessage.Define<string>(
+                LogLevel.Debug,
+                EventIds.CandidatesNotFound,
+                "No candidates found for the request path '{Path}'");
```
JamesNK: Log messages elsewhere end with a full stop. Repeat for all log messages.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
```diff
@@ -226,11 +258,100 @@ public sealed override Task MatchAsync(HttpContext httpContext, EndpointFeature
...
+            public static void CandidatesFound(ILogger logger, string path, Candidate[] candidates)
+            {
+                _candidatesFound(logger, candidates.Length, path, null);
+            }
+
+            public static void CandidateRejectedByComplexSegment(ILogger logger, string path, Endpoint endpoint, RoutePatternPathSegment segment)
+            {
+                // This should return a real pattern since we're processing complex segments.... but just in case.
+                var routePattern = GetRoutePattern(endpoint);
+                _candidateRejectedByComplexSegment(logger, endpoint.DisplayName, routePattern, segment.DebuggerToString(), path, null);
```
JamesNK: Usually a log level check happens inside the method when doing something allocaty. Add a comment that the DebuggerToString is ok because we've already checked the log level.

test/Microsoft.AspNetCore.Routing.Tests/Matching/DfaMatcherTest.cs
```diff
@@ -26,18 +28,23 @@ private RouteEndpoint CreateEndpoint(string template, int order, object defaults
...
 
-        private Matcher CreateDfaMatcher(EndpointDataSource dataSource, EndpointSelector endpointSelector = null)
+        private Matcher CreateDfaMatcher(EndpointDataSource dataSource, EndpointSelector endpointSelector = null, ILoggerFactory loggerFactory = null)
         {
             var serviceCollection = new ServiceCollection()
                 .AddLogging()
                 .AddOptions()
                 .AddRouting();
-
+
```
JamesNK: nit: white space

test/Microsoft.AspNetCore.Routing.Tests/Matching/DfaMatcherTest.cs
```diff
@@ -157,6 +164,173 @@ public async Task MatchAsync_MultipleMatches_EndpointSelectorCalled()
...
+                sink.Writes,
+                (log) => 
+                {
+                    Assert.Equal(DfaMatcher.EventIds.CandidatesFound, log.EventId);
+                    Assert.Equal("1 candidate(s) found for the request path '/One'", log.Message);
+                },
+                (log) =>
+                {
+                    Assert.Equal(DfaMatcher.EventIds.CandidateRejectedByConstraint, log.EventId);
+                    Assert.Equal("Endpoint '/{p:int}' with route pattern '/{p:int}' was rejected by constraint 'p' with value 'One' for the request path '/One'", log.Message);
```
JamesNK: p is the parameter name, not constraint name.

Did you mean to say "...constraint on parameter 'p'..."? FYI non-inline constraints won't have an inline name. Maybe include the type name of the constraint?

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
JamesNK: Actually this method isn't wrapped in an `if (log)` check. Should add it here.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
```diff
@@ -35,6 +38,9 @@ public sealed override Task MatchAsync(HttpContext httpContext, EndpointFeature
                 throw new ArgumentNullException(nameof(feature));
             }
 
+            // All of the logging we do here is at level debug, so we can get away with doing a single check.
+            var log = _logger.IsEnabled(LogLevel.Debug);
```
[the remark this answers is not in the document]
rynowak: Line 41??? I don't feel like we need to put "for perf" at the end of each sentence. Everything in here is done for perf 😆

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
rynowak: That's not true, I haven't heard such a guideline in the past no followed it elsewhere

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
rynowak: Yeah, this should be 👍

test/Microsoft.AspNetCore.Routing.Tests/Matching/DfaMatcherTest.cs
rynowak: I said constraint name for terseness - do you think there's a way to make it clearer without so many words?. non-inline constraints definitely have a name.

rynowak: I guess I dunno what you mean then. I don't know what an inline prefix is.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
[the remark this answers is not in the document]
JamesNK: ok!

test/Microsoft.AspNetCore.Routing.Tests/Matching/DfaMatcherTest.cs
JamesNK: These things - https://github.com/aspnet/Routing/blob/release/2.2/src/Microsoft.AspNetCore.Routing/RouteOptions.cs#L52-L76

rynowak: OK, I think I get what you mean now. I was thinking that logging the parameter name is enough. since you generally you wrote the route so you should know what's going on. Maybe we should log the `ToString()` of the constraint instance? 

By default that will be the type name, but if we had a case we cared about we could customize it.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
rynowak: I'm removing the periods from the other logging in this repo for consistency. The most visible and oldest parts of our stack (logging that happens for all projects at info level, like Hosting, MVC) don't use periods. 

If you think this is the wrong decision then that's a topic for the broader engineering team. We haven't had a defined practice in this area.

rynowak:
@james updated

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcher.cs
JamesNK: ok!

test/Microsoft.AspNetCore.Routing.Tests/Matching/DfaMatcherTest.cs
JamesNK: I still think this log message is questionable. p is the parameter name. Saying the value was rejected by constraint 'p' doesn't make sense to me.

Also what happens if a parameter has multiple constraints on it. The logging doesn't give you an indication of what constraint it was that failed.

rynowak: So I ended up pushing this with broken tests becuase I'm a silly billy. Now it looks like this:

> "Endpoint '/{p:int}' with route pattern '/{p:int}' was rejected by constraint 'p':'Microsoft.AspNetCore.Routing.Constraints.IntRouteConstraint' with value 'One' for the request path '/One'"

Does that address your concerns? If it doesn't please make a suggestion.

JamesNK: :shipit:

rynowak: Cool beans!

PR #833, opened 2018-10-03
Remove reliance on host header

rynowak:
This will require a reaction in MVC

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
@@ -59,10 +59,10 @@ namespace Microsoft.AspNetCore.Routing
         /// <param name="httpContext">The <see cref="HttpContext"/> associated with the current request.</param>
         /// <param name="values">The route values. Used to expand parameters in the route template. Optional.</param>
         /// <param name="scheme">
-        /// The URI scheme, applied to the resulting URI. Optional. If not provided, the value of <see cref="HttpRequest.Scheme"/> will be used.
+        /// The URI scheme, applied to the resulting URI.
         /// </param>
         /// <param name="host">
-        /// The URI host/authority, applied to the resulting URI. Optional. If not provided, the value <see cref="HttpRequest.Host"/> will be used.
+        /// The URI host/authority, applied to the resulting URI.
         /// </param>
         /// <param name="pathBase">
         /// An optional URI path base. Prepended to the path in the resulting URI. If not provided, the value of <see cref="HttpRequest.PathBase"/> will be used.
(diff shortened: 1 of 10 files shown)


test/Microsoft.AspNetCore.Routing.Tests/LinkGeneratorRouteValuesAddressExtensionsTest.cs
```diff
@@ -154,15 +154,17 @@ public void GetUriByRouteValues_WithHttpContext_WithPathBaseAndFragment()
             var linkGenerator = CreateLinkGenerator(endpoint1, endpoint2);
 
             var httpContext = CreateHttpContext();
-            httpContext.Request.Scheme = "http";
-            httpContext.Request.Host = new HostString("example.com");
+            httpContext.Request.Scheme = "httsp";
```
khellang: `httsp`? 🤔

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerationTemplate.cs
```diff
@@ -76,8 +76,8 @@ public abstract class LinkGenerationTemplate
         public abstract string GetUri(
             HttpContext httpContext,
             object values,
-            string scheme = default,
-            HostString? host = default,
+            string scheme,
+            HostString host,
```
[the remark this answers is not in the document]
rynowak: So the one thing that made me uncomfortable here was changing from `HostString?` to `HostString`. I think it's OK though because it's never valid to provide an empty host string. Therefore `default(HostString)` is OK to use as a sentinel value. But it's not a breaking change to make parameters optional in the future. 

For reference `PathString? basePath` is nullable because it is valid to have an 'empty' base path. We want to know the difference between 'empty' and unspecified.

src/Microsoft.AspNetCore.Routing.Abstractions/LinkGenerator.cs
```diff
@@ -97,9 +97,9 @@ public abstract class LinkGenerator
             HttpContext httpContext,
             TAddress address,
             RouteValueDictionary values,
-            RouteValueDictionary ambientValues = default,
-            string scheme = default,
-            HostString? host = default,
+            RouteValueDictionary ambientValues,
```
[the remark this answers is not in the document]
rynowak: Yeah, this made me sad. Do you have any brilliant ideas about it?

test/Microsoft.AspNetCore.Routing.Tests/LinkGeneratorRouteValuesAddressExtensionsTest.cs
rynowak: httsp is HTTP 3 - I'm not much public information is available yet 😆

rynowak: I'm going to leave this here for posterity. This value doesn't show up in the test output, it's specifically there to demonstrate that it doesn't get used.

JamesNK:
A wild opinion appears: https://github.com/aspnet/Mvc/pull/8550#discussion_r222563027

PR #834, opened 2018-10-04
Make endpoint selector policies per-candidate-set

rynowak:
This allows us to filter `IEndpointSelectorPolicy` instance based on
whether the apply to a given candidate set. This should allow us to
remove some HAXXX from MVC.

The idea here is the ESP becomes much more pay-for-play if you can
statically eliminate many of the cases where it would usually no op.


rynowak:
This will require a reaction PR in MVC

rynowak:
Note: I deleted a category of perf tests that have largely become redundant. I haven't run these in months, and they are a subset of some other tests that I use more often.

src/Microsoft.AspNetCore.Routing/Matching/INodeBuilderPolicy.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
 {
     public interface INodeBuilderPolicy
     {
-        bool AppliesToNode(IReadOnlyList<Endpoint> endpoints);
+        bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints);
```
JamesNK: 👍

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -375,8 +377,30 @@ public override Matcher Build()
             }
 
             var candidates = CreateCandidates(node.Matches);
+
+            // Perf: most of the time there aren't any endpoint selector policies, create
+            // this lazily.
+            List<IEndpointSelectorPolicy> endpointSelectorPolicies = null;
+            if (node.Matches?.Count > 0)
```
rynowak: Note there was a bug here that I fixed. This was passing in a 'null' list of matches to `AppliesToEndpoints`.

PR #835, opened 2018-10-04
Add blurb about Host header

rynowak:


rynowak:
@blowdart will correct me if I am wrong, but I don't think the same issues apple to scheme. The issue at hand is that the host-> URI connection can direct end users to a different site. Scheme can't do that.

We also have good ways to require you to use `https`. The concern was whether we had good ways to require an expected hostname.

blowdart:
Yea I wouldn't worry about scheme so much, if you care about HTTPS you're going to be using HSTS anyway

Issue #825, opened 2018-09-29
Domains with EP routing


JamesNK:
I came across this question while waiting for a flight - https://www.reddit.com/r/dotnet/comments/9jn69g/question_not_having_much_luck_on_stack_overflow/

What can we do to make multiple domains with asp.net core easy? Am I right that a policy could work? Should come up with a sample.

rynowak:
It depends what you want. This is a really open ended question. I think the start for this would be to define what the scenarios and requirements are.

JamesNK:
I can think of three scenarios:
1. One endpoint but the domain is a route value.
```cs
public IActionResult GetBlogPosts(string domain) { }
```
Probably not valuable because you can get the host off the HttpRequest in the controller anyway.

2. Different controllers (maybe grouped into an area) depending on the domain
```cs
[Route("/")]
[Domain("contoso.com")]
public class HomeController {
    public IActionResult GetBlogPosts() { }
}

[Route("/")]
[Domain("adventureworks.com")]
public class HomeController {
    public IActionResult GetBlogPosts() { }
}
```
3. Different razor pages (with the domain in the path) depending on the domain
```
Pages/adventureworks/AboutUs
Pages/contoso/AboutUs
```

rynowak:
> One endpoint but the domain is a route value.

What would you use that route value for? I'm not sure this is a thing, because *route values -> selecting an endpoint* OR *route values -> URL generation* - if you're not using it to change how either of these behave, then it's not really a routing feature. 

If the whole ask is just to be able to tokenize a hostname with routing, that's kinda weak. I've seen people ask for `{sub}.mydomain/com/....` where all that we do is populate a route value from that so they can model bind it. 

> Different controllers (maybe grouped into an area) depending on the domain

I've seen people ask for restricting access to parts of an app by domain or by port, or other various criteria. It seems like a good fit for policies.

PR #847, opened 2018-10-10
Add domain policy

JamesNK:
Fixes https://github.com/aspnet/Routing/issues/825

Note that this is based on https://github.com/aspnet/Routing/pull/799


samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
```diff
@@ -0,0 +1,184 @@
...
+            // are then then execute them in linear order.
+            var ordered = edges
+                .Select(e => (domain: (HostString)e.State, destination: e.Destination))
+                .OrderBy(e => GetScore(e.domain))
+                .ToArray();
+
+            return new DomainPolicyJumpTable(exitDestination, ordered);
+        }
+
+        private int GetScore(in HostString domain)
```
rynowak: This probably deserves a comment about why it's not used by the comparer. It makes sense to me. But I might forget.

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        public IReadOnlyList<PolicyNodeEdge> GetEdges(IReadOnlyList<Endpoint> endpoints)
+        {
+            if (endpoints == null)
+            {
+                throw new ArgumentNullException(nameof(endpoints));
+            }
+
+            // The algorithm here is designed to be preserve the order of the endpoints
+            // while also being relatively simple. Preserving order is important.
```
rynowak: This comment is designed to be totally original, while also being a copy paste from something I wrote. S P I C Y

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
```diff
@@ -0,0 +1,184 @@
...
+                    {
+                        kvp.Value.Add(endpoint);
+                    }
+                }
+                else
+                {
+                    // OK this endpoint matches specific domains
+                    foreach (var kvp in edges)
+                    {
+                        // The edgeKey maps to a possible request header value
```
rynowak: host?

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
```diff
@@ -0,0 +1,184 @@
...
+                        var edgeKey = kvp.Key;
+
+                        for (var j = 0; j < domains.Count; j++)
+                        {
+                            var domain = domains[j];
+
+                            if (edgeKey == domain)
```
rynowak: Is this logic correct when you have wildcards? It sounds like it's not, but you expect the ordering of the domains to help you. Consider a case like this:

Endpoint1:
domain: foo.example.com
version: v2

Endpoint2:
domain *.example.com
version: v1

This code will be build edges like this:

foo.example.com -> (Endpoint1)
*.example.com -> (Endpoint2)

So what if the request is for foo.example.com - version 1?

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
rynowak: Maybe this should be used by the comparer. See https://github.com/aspnet/Routing/pull/847/files#r224504940

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
```diff
@@ -0,0 +1,184 @@
...
+                _destinations = destinations;
+            }
+
+            public override int GetDestination(HttpContext httpContext)
+            {
+                var destinations = _destinations;
+                for (var i = 0; i < destinations.Length; i++)
+                {
+                    if ((string.IsNullOrEmpty(destinations[i].domain.Host) ||
+                        httpContext.Request.Host.Host == destinations[i].domain.Host) &&
```
rynowak: maybe I misunderstand what you're saying about wildcards? I kinda expected to see handling for something like `*.example.com`

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
rynowak: OR 

> maybe I misunderstand what you're saying about wildcards? I kinda expected to see handling for something like *.example.com

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
JamesNK: 😋

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
JamesNK: By wildcards I meant that it supports *:port, or host:*. I didn't consider wildcards inside the host. That would be a required feature. Will add!

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
JamesNK: I copied this from the ConsumesMatcherPolicy. Should it be using the score in the comparer as well?

samples/RoutingSample.Web/DomainPolicy/DomainMatcherPolicy.cs
JamesNK: host is a request header value

Technically correct: The best kind of correct

natemcmaster:
We're merging Routing today into aspnet/AspNetCore. If you can merge this in the next 30 minutes or so, we can bring this along. Otherwise, you'll have to open a new PR against https://github.com/aspnet/AspNetCore.

PR #858, opened 2018-10-14
Add Remove(string key, out object value) overload to RouteValueDictionary

drieseng:
Adds a `Remove(string key, out object value)` overload to **RouteValueDictionary**, and use it where applicable.

Note that we could also benefit from using the same overload on **Dictionary<TKey, TValue>** in other places, but we cannot since both **.NET Standard 2.0** and **.NET Core 2.2** are being target (unless we use conditional compilation).

### DIFF
--- a/src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
+++ b/src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
@@ -405,6 +405,46 @@ namespace Microsoft.AspNetCore.Routing
         }
 
         /// <summary>
+        /// Attempts to remove and return the value that has the specified key from the <see cref="RouteValueDictionary"/>.
+        /// </summary>
+        /// <param name="key">The key of the element to remove and return.</param>
+        /// <param name="value">When this method returns, contains the object removed from the <see cref="RouteValueDictionary"/>, or <c>null</c> if key does not exist.</param>
+        /// <returns>
+        /// <c>true</c> if the object was removed successfully; otherwise, <c>false</c>.
+        /// </returns>
+        public bool Remove(string key, out object value)
+        {
+            if (key == null)
+            {
+                ThrowArgumentNullExceptionForKey();
+            }
+
+            if (Count == 0)
+            {
+                value = default;
+                return false;
+            }
+
+            EnsureCapacity(Count);
+
+            var index = FindIndex(key);
+            if (index >= 0)
+            {
+                _count--;
+                var array = _arrayStorage;
+                value = array[index].Value;
+                Array.Copy(array, index + 1, array, index, _count - index);
+                array[_count] = default;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -404,6 +404,46 @@ public bool Remove(string key)
...
+                ThrowArgumentNullExceptionForKey();
+            }
+
+            if (Count == 0)
+            {
+                value = default;
+                return false;
+            }
+
+            EnsureCapacity(Count);
```
rynowak: Add a comment that this is called for the side-effect.

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
```diff
@@ -404,6 +404,46 @@ public bool Remove(string key)
...
+        /// <c>true</c> if the object was removed successfully; otherwise, <c>false</c>.
+        /// </returns>
+        public bool Remove(string key, out object value)
+        {
+            if (key == null)
+            {
+                ThrowArgumentNullExceptionForKey();
+            }
+
+            if (Count == 0)
```
rynowak: nit: there's lots of mixing of `Count` and `_count` here. Does it make sense to use a local and update at the end?

test/Microsoft.AspNetCore.Mvc.Routing.Abstractions.Tests/RouteValueDictionaryTests.cs
```diff
@@ -1393,6 +1393,161 @@ public void Remove_ListStorage_True_CaseInsensitive()
...
+            // Act
+            var result = dict.Remove("kEy", out var removedValue);
+
+            // Assert
+            Assert.True(result);
+            Assert.Same(value, removedValue);
+            Assert.Empty(dict);
+            Assert.IsType<KeyValuePair<string, object>[]>(dict._arrayStorage);
+        }
+
```
rynowak: For array storage I'd also like to see tests for the case where the value removed is the first/middle/last

src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs
drieseng: I updated the new Remove overload to consistently use the _count field.
Note that there's a mix throughout this class.

test/Microsoft.AspNetCore.Mvc.Routing.Abstractions.Tests/RouteValueDictionaryTests.cs
drieseng: Done

PR #869, opened 2018-10-17
Add test for link generation with high priority required parameter

JamesNK:
Sanity test that re-arranging route order in MVC will generate the correct results.

Looks good to me

### DIFF
--- a/test/Microsoft.AspNetCore.Routing.Tests/LinkGeneratorRouteValuesAddressExtensionsTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/LinkGeneratorRouteValuesAddressExtensionsTest.cs
@@ -18,64 +18,6 @@ namespace Microsoft.AspNetCore.Routing
     public class LinkGeneratorRouteValuesAddressExtensionsTest : LinkGeneratorTestBase
     {
         [Fact]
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Routing.Tests/LinkGeneratorRouteValuesAddressExtensionsTest.cs
```diff
@@ -17,6 +17,64 @@ namespace Microsoft.AspNetCore.Routing
     // Does not cover the RouteValuesAddressScheme in detail. see RouteValuesAddressSchemeTest
     public class LinkGeneratorRouteValuesAddressExtensionsTest : LinkGeneratorTestBase
     {
+        [Fact]
+        public void GetPathByRouteValues_UsePrioritizedEndpointOnlyWithMatchingRouteValue()
```
rynowak: This could really be called `UsesFirstTemplateThatSucceeds`.

rynowak: This is really a test for https://github.com/aspnet/Routing/blob/release/2.2/test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs - it's super general and isn't related to these overloads.

JamesNK:
🆙 📅

PR #870, opened 2018-10-17
Prefer StringComparison over StringComparer when not using comparers

JamesNK:
https://stackoverflow.com/a/8918112/11829


rynowak:
The code change seem fine, but I've read that post over a few times and I still don't get it. How does this make user's lives better? Does this more perf?

JamesNK:
There is no real benefit, it's just StringComparer is for cases when an API uses IComparer, e.g. the Dictionary comparer, and StringComparison is for regular string comparisons.

rynowak:
Ok - you get an approve because its clearer? I guess?

JamesNK:
You have forgotten the most valuable thing of all: 1 James point

PR #872, opened 2018-10-19
Allow parameter names to match required keys in templates

JamesNK:
Just works? Or was your certain some additional work happening in GetValues

### DIFF
--- a/test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
+++ b/test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
@@ -441,9 +441,9 @@ namespace Microsoft.AspNetCore.Routing
 
             // Act
             var uri = linkGenerator.GetPathByAddress(
-                httpContext, 
-                1, 
-                values: new RouteValueDictionary(new { action = "Index", }), 
+                httpContext,
+                1,
+                values: new RouteValueDictionary(new { action = "Index", }),
                 ambientValues: new RouteValueDictionary(new { controller = "Home", }));
 
             // Assert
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -656,6 +656,115 @@ public void GetPathByRouteValues_UsesFirstTemplateThatSucceeds()
...
+
+            var pathToNonDefaultActionWithId = linkGenerator.GetPathByRouteValues(
+                httpContext,
+                routeName: null,
+                values: new RouteValueDictionary(new { controller = "Home", action = "Login", id = "3" }));
+
+            var pathToInvalidAction = linkGenerator.GetPathByRouteValues(
+                httpContext,
+                routeName: null,
+                values: new RouteValueDictionary(new { controller = "Home", action = "Fake", id = "3" }));
```
rynowak: I'm really not a fan of this. Is this like 5 tests in a single method?

test/Microsoft.AspNetCore.Routing.Tests/DefaultLinkGeneratorTest.cs
```diff
@@ -656,6 +656,115 @@ public void GetPathByRouteValues_UsesFirstTemplateThatSucceeds()
             Assert.Equal("/?custom=Custom", pathWithCustom);
         }
 
+        [Fact]
+        public void GetPathByRouteValues_ParameterMatchesRequireValues_HasAmbientValues()
```
rynowak: I don't think this test belongs here - this is a functional tests

PR #873, opened 2018-10-19
Declare out variables inline, and use discards

drieseng:
Declare out variables inline, and use discards when we do not need the value.


drieseng:
Let me know if you prefer to use **out var**.

rynowak:
> Let me know if you prefer to use out var.

Our guidelines are to use var everywhere that it's possible/feasible.

src/Microsoft.AspNetCore.Routing/Constraints/RangeRouteConstraint.cs
```diff
@@ -58,12 +58,10 @@ public RangeRouteConstraint(long min, long max)
...
             }
 
-            object value;
-            if (values.TryGetValue(routeKey, out value) && value != null)
+            if (values.TryGetValue(routeKey, out object value) && value != null)
             {
-                long longValue;
                 var valueString = Convert.ToString(value, CultureInfo.InvariantCulture);
-                if (Int64.TryParse(valueString, NumberStyles.Integer, CultureInfo.InvariantCulture, out longValue))
+                if (Int64.TryParse(valueString, NumberStyles.Integer, CultureInfo.InvariantCulture, out long longValue))
```
JamesNK: nit: Int64.TryParse -> long.TryParse

drieseng: Done

drieseng:
Should be ok now.
Sorry for the commit/merge mess.

PR #883, opened 2018-10-21
Add basic test for RouteValueEqualityComparer.Equals(...).

drieseng:


test/Microsoft.AspNetCore.Routing.Tests/RouteValueEqualityComparerTest.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Xunit;
```
JamesNK: Add the copyright header used everywhere to the new file

drieseng: Done

PR #886, opened 2018-10-22
Eliminate extra allocation due to use of method group.

drieseng:
Eliminate an extra allocation due to the use of method group.

This can be reverted once this is improved in either [Roslyn](https://github.com/dotnet/roslyn/issues/5835) or [RyuJIT](https://github.com/dotnet/coreclr/issues/1784).

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/EndpointNameAddressScheme.cs
+++ b/src/Microsoft.AspNetCore.Routing/EndpointNameAddressScheme.cs
@@ -15,7 +15,13 @@ namespace Microsoft.AspNetCore.Routing
 
         public EndpointNameAddressScheme(CompositeEndpointDataSource dataSource)
         {
-            _cache = new DataSourceDependentCache<Dictionary<string, Endpoint[]>>(dataSource, Initialize);
+            // Do not use method group to avoid extra allocation
+            // Can be removed once this is improved in either:
+            // Roslyn: https://github.com/dotnet/roslyn/issues/5835
+            // -or-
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Routing/EndpointNameAddressScheme.cs
```diff
@@ -15,7 +15,13 @@ internal class EndpointNameAddressScheme : IEndpointAddressScheme<string>
...
         public EndpointNameAddressScheme(CompositeEndpointDataSource dataSource)
         {
-            _cache = new DataSourceDependentCache<Dictionary<string, Endpoint[]>>(dataSource, Initialize);
+            // Do not use method group to avoid extra allocation
+            // Can be removed once this is improved in either:
+            // Roslyn: https://github.com/dotnet/roslyn/issues/5835
+            // -or-
+            // JIT: https://github.com/dotnet/coreclr/issues/1784
+            _cache = new DataSourceDependentCache<Dictionary<string, Endpoint[]>>(dataSource,
+                                                                                  (endpoints) => Initialize(endpoints));
```
JamesNK: Is this an extra allocation each time the cache is used, or an extra allocation when EndpointNameAddressScheme is created? If it is only one allocation when EndpointNameAddressScheme is created then it isn't worth changing - this class is a singleton.

JamesNK: The improvement is so minor, and it is one the jit/clr will eventually make anyway, that it isn't worth it. Thanks for the suggestion tho!

PR #897, opened 2018-10-31
Fix Routing's misuse of PathString

rynowak:
PathString is meant to be used with *unencoded* text. So in our case, we
should only be responsible for encoding slashes we don't want to
roundtrip.

This is working fine in 2.2 but it's inefficient so I'm targeting this
at 3.0

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
+++ b/src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
@@ -14,8 +14,14 @@ namespace Microsoft.AspNetCore.Routing.Internal
     public class UriBuildingContext
     {
         // Holds the 'accepted' parts of the path.
+        //
+        // Note that the path is stored as **unencoded** (except slashes in some cases) for parity with PathString.
         private readonly StringBuilder _path;
-        private StringBuilder _query;
+
+        // Holds the query string
+        //
+        // Note that the query is stored as **encoded** for parity with QueryString.
+        private readonly StringBuilder _query;
 
         // Holds the 'optional' parts of the path. We need a secondary buffer to handle cases where an optional
         // segment is in the middle of the uri. We don't know if we need to write it out - if it's
@@ -32,7 +38,6 @@ namespace Microsoft.AspNetCore.Routing.Internal
             _path = new StringBuilder();
             _query = new StringBuilder();
             _buffer = new List<BufferValue>();
-            PathWriter = new StringWriter(_path);
             QueryWriter = new StringWriter(_query);
             _lastValueOffset = -1;
 
@@ -49,8 +54,6 @@ namespace Microsoft.AspNetCore.Routing.Internal
         public SegmentState BufferState { get; private set; }
 
         public SegmentState UriState { get; private set; }
-
-        public TextWriter PathWriter { get; }
 
         public TextWriter QueryWriter { get; }
 
@@ -97,7 +100,7 @@ namespace Microsoft.AspNetCore.Routing.Internal
 
                 if (buffer[i].RequiresEncoding)
                 {
-                    EncodeValue(bufferValue);
+                    EncodePathValue(bufferValue);
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


rynowak:
I also expect a nasty reaction PR from MVC 😢 - I think this pattern with TestEncoder was used in lots of places in MVC.

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateBinderTests.cs
```diff
@@ -594,7 +594,7 @@ public void GetUrlVerifyEncoding()
                 new RouteValueDictionary(new { controller = "Home" }),
                 new RouteValueDictionary(new { controller = "home", action = "Index", id = (string)null }),
                 values,
-                "/%23;%3F%3A@%26%3D%2B$,.mvc/showcategory/123?so%3Frt=de%3Fsc&maxPrice=100");
+                "/%23;%3F:@&=+$,%2F.mvc/showcategory/123?so%3Frt=de%3Fsc&maxPrice=100");
```
rynowak: @Tratcher - please confirm that this is right 👍 - I think PathString is less aggressive than UrlEncoder on purpose.

Tratcher: 👍

test/Microsoft.AspNetCore.Routing.Tests/Internal/UriBuildingContextTest.cs
```diff
@@ -1,24 +1,26 @@
...
 
 namespace Microsoft.AspNetCore.Routing.Internal
 {
+    // Note that these tests look a little wierd. It's because each test treats the input
+    // as a single route segment that *might* contain slashes. While we're producing paths, these
+    // are unit tests for the encoding functionality.
     public class UriBuildingContextTest
     {
         [Fact]
         public void EncodeValue_EncodesEntireValue_WhenEncodeSlashes_IsFalse()
```
Tratcher: These tests names are matching their contents.

src/Microsoft.AspNetCore.Routing/Internal/UriBuildingContext.cs
```diff
@@ -279,42 +282,43 @@ public QueryString ToQueryString()
...
                 int end;
-                int length = start + characterCount;
-                while ((end = value.IndexOf('/', start, characterCount)) >= 0)
+                int length = start + count;
+                while ((end = value.IndexOf('/', start, count)) >= 0)
                 {
-                    _urlEncoder.Encode(PathWriter, value, start, end - start);
-                    _path.Append("/");
+                    _path.Append(value, start, end - start);
+                    _path.Append("%2F");
```
JamesNK: nit: Add comment why %2F for anyone who didn't read the comment on _path

test/Microsoft.AspNetCore.Routing.Tests/Internal/UriBuildingContextTest.cs
```diff
@@ -28,66 +30,63 @@ public void EncodeValue_EncodesEntireValue_WhenEncodeSlashes_IsFalse()
...
-        [InlineData("Author", true, false, "/UrlEncode[[Author]]/")]
-        [InlineData("Author", true, true, "/UrlEncode[[Author]]/")]
+        [InlineData("/Author", false, false, "/Author")]
+        [InlineData("/Author", false, true, "/Author")]
+        [InlineData("/Author", true, false, "/Author/")]
+        [InlineData("/Author", true, true, "/Author/")]
+        [InlineData("/Author/", false, false, "/Author/")]
+        [InlineData("/Author/", false, true, "/Author%2F")]
+        [InlineData("/Author/", true, false, "/Author/")]
+        [InlineData("/Author/", true, true, "/Author%2F/")]
```
JamesNK: Are these expected paths right?

rynowak: Yeah this is kinda a janky test - it's taking a single string/segment, then encoding it, then getting the result as a full path. 

It's doing a cross product of all possible cases instead of factoring the code into independent testable units. Theory tests are always a bad decision.

rynowak:
@natemcmaster - this will require cross-repo testing. Since this isn't urgent I'm going to pick this up after the Razor and MVC repos get merged.

natemcmaster:
Thanks for the heads up.

natemcmaster:
Closing as this repo is merging to aspnet/AspNetCore today.

PR #901, opened 2018-11-03
Improve performance and reduce allocations in RouteParameterParser.

drieseng:
Performance is improved by eliminating use of Linq's `Enumerable.ToArray()` and avoiding the creation of a new array. The latter is courtesy of CoreFX's ArrayBuilder, which also results in a drop in allocations.

Result:

|              Method | Version   |      Mean |    Error  |   StdDev  |         Op/s |  Gen 0 | Allocated |
|-------------------- |:---------:|----------:|----------:|----------:|-------------:|-------:|----------:|
| MultipleConstraints | master    | 318.46 ns | 4.631 ns  | 4.332 ns  |  3,140,071.8 | 0.0067 |     608 B |
| MultipleConstraints | PR        | 248.42 ns | 2.5049 ns | 2.2205 ns |  4,025,506.6 | 0.0057 |     512 B |
|       NoConstraints | master    |  87.08 ns | 2.085 ns  | 1.950 ns  | 11,483,551.4 | 0.0014 |     128 B |
|       NoConstraints | PR        |  44.77 ns | 0.8919 ns | 0.9543 ns | 22,336,882.6 | 0.0010 |      88 B |
|    SingleConstraint | master    | 150.51 ns | 2.514 ns  | 2.352 ns  |  6,644,202.1 | 0.0036 |     320 B |
|    SingleConstraint | PR        |  98.80 ns | 1.0250 ns | 0.9587 ns | 10,121,503.9 | 0.0030 |     280 B |


drieseng:
As noted in the PR description, I've started using **ArrayBuilder\<T>** from CoreFX.
If this is considered ok, I'll use it in more classes (eg. **RoutePatternParser**) in future PRs.

**Note:**
See https://github.com/dotnet/corefx/issues/33169 for a discussion on making **ArrayBuilder\<T>** public.

benchmarks/Microsoft.AspNetCore.Routing.Performance/RouteParameterParserBenchmark.cs
```diff
@@ -0,0 +1,26 @@
+﻿using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    public class RouteParameterParserBenchmark
```
drieseng: Let me know if you want me to remove the benchmark.

JamesNK:
This doesn't hurt but parsing routes isn't a bottleneck of ASP.NET apps. It is likely only done once at startup. URL matching and link generation are what happens every request.

benchmarks/Microsoft.AspNetCore.Routing.Performance/RouteParameterParserBenchmark.cs
JamesNK: Yes remove them

PR #902, opened 2018-11-04
Improve performance and reduce allocations in TemplateBinder ctor

drieseng:
Improves performance and reduce allocations in TemplateBinder ctors:
* Avoids looping multiple times over parameter policies, and no longer uses Linq.
* Only filters out parameter names from defaults if there are defaults.
* When assigning slots, obtains parameter count only once.

Result:

|                                                Method |  Version |     Mean |    Error |   StdDev |        Op/s |  Gen 0 | Allocated |
|------------------------------------------------------ |:--------:|---------:|---------:|---------:|------------:|-------:|----------:|
|       ParametersAndDefaultsAndRequiredKeysAndPolicies | master   | 768.0 ns | 5.927 ns | 5.544 ns | 1,302,060.6 | 0.0105 |     968 B |
|       ParametersAndDefaultsAndRequiredKeysAndPolicies | PR - List\<T> |589.1 ns | 5.514 ns | 5.158 ns | 1,697,611.7 | 0.0095 |     840 B |
|       ParametersAndDefaultsAndRequiredKeysAndPolicies | PR - ArrayBuilder\<T> |565.9 ns | 1.284 ns | 1.072 ns | 1,767,084.3 | 0.0076 |     760 B |
|     ParametersAndNoDefaultsAndRequiredKeysAndPolicies | master   | 637.8 ns | 6.036 ns | 5.351 ns | 1,567,837.6 | 0.0095 |     824 B |
|     ParametersAndNoDefaultsAndRequiredKeysAndPolicies | PR - List\<T> |371.5 ns | 3.651 ns | 3.415 ns | 2,691,922.5 | 0.0076 |     696 B |
|     ParametersAndNoDefaultsAndRequiredKeysAndPolicies | PR - ArrayBuilder\<T> | 349.1 ns | 2.952 ns | 2.616 ns | 2,864,878.1 | 0.0072 |     616 B |
|     NoParametersAndDefaultsAndRequiredKeysAndPolicies | master   | 690.6 ns | 3.484 ns | 3.089 ns | 1,448,016.1 | 0.0105 |     984 B |
|     NoParametersAndDefaultsAndRequiredKeysAndPolicies | PR - List\<T> | 483.5 ns | 9.588 ns | 8.968 ns | 2,068,154.3 | 0.0095 |     856 B |
|     NoParametersAndDefaultsAndRequiredKeysAndPolicies | PR - ArrayBuilder\<T> | 458.9 ns | 3.747 ns | 3.505 ns | 2,179,081.6 | 0.0086 |     776 B |
| NoParametersAndNoDefaultsAndRequiredKeysAndNoPolicies | master   | 386.1 ns | 2.847 ns | 2.523 ns | 2,590,204.2 | 0.0048 |     424 B |
| NoParametersAndNoDefaultsAndRequiredKeysAndNoPolicies | PR - List\<T> | 316.5 ns | 3.476 ns | 3.252 ns | 3,159,387.1 | 0.0052 |     504 B |
| NoParametersAndNoDefaultsAndRequiredKeysAndNoPolicies | PR - ArrayBuilder\<T> | 300.6 ns | 2.025 ns | 1.795 ns | 3,326,396.0 | 0.0043 |     424 B |

Benchmark is available [here](https://gist.github.com/drieseng/23a663a5d7521b48b431cf41793016da).

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Internal/ArrayBuilder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Internal/ArrayBuilder.cs
@@ -1,0 +1,169 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+// LICENSING NOTE: This file is from the dotnet corefx repository.
+// 
+// See https://github.com/dotnet/corefx/blob/143df51926f2ad397fef9c9ca7ede88e2721e801/src/Common/src/System/Collections/Generic/ArrayBuilder.cs 
+
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Routing.Internal
+{
+    /// <summary>
+    /// Helper type for avoiding allocations while building arrays.
+    /// </summary>
+    /// <typeparam name="T">The element type.</typeparam>
+    internal struct ArrayBuilder<T>
+    {
+        private const int DefaultCapacity = 4;
+        private const int MaxCoreClrArrayLength = 0x7fefffff; // For byte arrays the limit is slightly larger
+
+        private T[] _array; // Starts out null, initialized on first Add.
+        private int _count; // Number of items into _array we're using.
+
+        /// <summary>
+        /// Initializes the <see cref="ArrayBuilder{T}"/> with a specified capacity.
+        /// </summary>
+        /// <param name="capacity">The capacity of the array to allocate.</param>
+        public ArrayBuilder(int capacity) : this()
+        {
+            Debug.Assert(capacity >= 0);
+            if (capacity > 0)
+            {
+                _array = new T[capacity];
+            }
+        }
+
+        /// <summary>
+        /// Gets the number of items this instance can store without re-allocating,
+        /// or 0 if the backing array is <c>null</c>.
+        /// </summary>
+        public int Capacity => _array?.Length ?? 0;
+
+        /// <summary>Gets the current underlying array.</summary>
+        public T[] Buffer => _array;
+
+        /// <summary>
+        /// Gets the number of items in the array currently in use.
+        /// </summary>
+        public int Count => _count;
+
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Routing/Internal/ArrayBuilder.cs
```diff
@@ -0,0 +1,169 @@
...
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Routing.Internal
+{
+    /// <summary>
+    /// Helper type for avoiding allocations while building arrays.
+    /// </summary>
+    /// <typeparam name="T">The element type.</typeparam>
+    internal struct ArrayBuilder<T>
```
JamesNK: This builder saves an allocation from the list. Does it do anything else?

drieseng: Due to the fact that it is only used for constructing an array, it does not need to perform any version checks and in general it perform less argument checks (eg. doesn't throw if negative capacity is specified). In other words, it uses less memory and and provides better throughput.


If the **TemplateBinder** ctor is not performance critical then we could just use **List\<T>**, but that offers no benefit.

I've updated the PR description to show results for both **List\<T>** and **ArrayBuilder\<T>**.

JamesNK:
> Do you only ever intend to use the first parameter transformer?

Yes, for simplicity.

> Also, why not pass the constraints and transformer(s) separately to the TemplateBinder ctor? Right now, we do the work of extracting them from the list of parameter policies twice.

It is a public constructor. If additional usages of policies in the future then the constructor would need to change.

JamesNK (approved):

natemcmaster:
@rynowak @JamesNK We're merging Routing today into aspnet/AspNetCore. If you can merge this in the next 30 minutes or so, we can bring this along. Otherwise, you'll have to open a new PR against https://github.com/aspnet/AspNetCore.

Issue #906, opened 2018-11-07
RouteValuesAddressScheme issue


JamesNK:
`RouteValuesAddressScheme` will match endpoints that don't have `RouteValuesAddressMetadata` metadata. Seems like a bug to me.

From investigating https://github.com/aspnet/Mvc/issues/8690

rynowak:
Oh that's bad 😢

Eilon:
@rynowak bad enough to take in 2.2?

Eilon:
Oh sorry I missed the email where you said it was OK to go into 3.0.

PR #908, opened 2018-11-07
Fix RouteValueAddressScheme to require metadata to match

JamesNK:
Fixes https://github.com/aspnet/Routing/issues/906

Note that this issue also exists in release/2.2 but because all endpoints created by `MvcEndpointDataSource` have `IRouteValuesAddressMetadata`. This would be an issue if anyone creates their own data source. Should it be fixed on release/2.2 as well?


JamesNK:
Replaced by https://github.com/aspnet/Routing/pull/910

PR #910, opened 2018-11-07
Fix RouteValueAddressScheme to require metadata to match

### LINKED ISSUE #906: RouteValuesAddressScheme issue

JamesNK:
Fixes https://github.com/aspnet/Routing/issues/906

Note that this issue also exists in release/2.2 but because all endpoints created by `MvcEndpointDataSource` have `IRouteValuesAddressMetadata`. This would be an issue if anyone creates their own data source. Should it be fixed on release/2.2 as well?

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/RouteValuesAddressScheme.cs
+++ b/src/Microsoft.AspNetCore.Routing/RouteValuesAddressScheme.cs
@@ -125,6 +125,11 @@ namespace Microsoft.AspNetCore.Routing
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


rynowak (approved):

rynowak (approved):

JamesNK:
The rare double Ryan approval! 🤗

PR #911, opened 2018-11-13
Validate the maximum number of route segments

JamesNK:
Fixes https://github.com/aspnet/AspNetCore/issues/3001

I also updated the unit tests so the RoutePattern overloads are tested.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Template/RoutePrecedence.cs
+++ b/src/Microsoft.AspNetCore.Routing/Template/RoutePrecedence.cs
@@ -20,6 +20,8 @@ namespace Microsoft.AspNetCore.Routing.Template
         //       /api/template/{id:int} == 1.12
         public static decimal ComputeInbound(RouteTemplate template)
         {
+            ValidateSegementLength(template.Segments.Count);
+
             // Each precedence digit corresponds to one decimal place. For example, 3 segments with precedences 2, 1,
             // and 4 results in a combined precedence of 2.14 (decimal).
             var precedence = 0m;
@@ -40,6 +42,8 @@ namespace Microsoft.AspNetCore.Routing.Template
         // See description on ComputeInbound(RouteTemplate)
         internal static decimal ComputeInbound(RoutePattern routePattern)
         {
+            ValidateSegementLength(routePattern.PathSegments.Count);
+
             var precedence = 0m;
 
             for (var i = 0; i < routePattern.PathSegments.Count; i++)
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Routing/Template/RoutePrecedence.cs
```diff
@@ -99,6 +107,15 @@ internal static decimal ComputeOutbound(RoutePattern routePattern)
             return precedence;
         }
 
+        private static void ValidateSegementLength(int length)
+        {
+            if (length > 28)
+            {
+                // An OverflowException will be thrown by Math.Pow when greater than 28
+                throw new InvalidOperationException("Route contains too many segments and is unable to be parsed.");
```
rynowak: We should also include the max number allowed in the error message

test/Microsoft.AspNetCore.Routing.Tests/Template/RoutePatternPrecedenceTests.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Routing.Template
+{
+    public class RoutePatternPrecedenceTests : RoutePrecedenceTestsBase
```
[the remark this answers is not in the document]
JamesNK: That's all the test class does.

There are two copies of computing route precedence: one for the old template and one for pattern. This change just applies all the old template tests to pattern without copy & paste.

rynowak (approved):

test/Microsoft.AspNetCore.Routing.Tests/Template/RoutePrecedenceTestsBase.cs
```diff
@@ -2,13 +2,11 @@
...
 
 using System;
-using Microsoft.Extensions.Options;
-using Moq;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Routing.Template
 {
-    public class RoutePrecedenceTests
+    public abstract class RoutePrecedenceTestsBase
```
dallancarr: Tests in abstract class?

JamesNK: They are run in classes that inherit from it

PR #912, opened 2018-11-15
Add support for required values to RoutePattern

rynowak:
This change adds first class support for what MVC does to substitute route values into templates. Rather than munging strings, I chose to introduce required values as a fundamental concept.

The idea is that a *conventional-route-style* template could we written once, and then *expanded* for every action and controller that exists. Making this first class in routing moves the complexity into the right place, and will reduce the number of endpoints produced for a conventionally routed application significantly.

Still left to do here is to add support to `DfaMatcher` to process the required values. Then we should probably obsolete `IRouteValuesAddressMetadata` since it's now first class.

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Patterns/DefaultRoutePatternTransformer.cs
+++ b/src/Microsoft.AspNetCore.Routing/Patterns/DefaultRoutePatternTransformer.cs
@@ -1,0 +1,227 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    internal class DefaultRoutePatternTransformer : RoutePatternTransformer
+    {
+        private readonly ParameterPolicyFactory _policyFactory;
+
+        public DefaultRoutePatternTransformer(ParameterPolicyFactory policyFactory)
+        {
+            if (policyFactory == null)
+            {
+                throw new ArgumentNullException(nameof(policyFactory));
+            }
+
+            _policyFactory = policyFactory;
+        }
+
+        public override RoutePattern SubstituteRequiredValues(RoutePattern original, object requiredValues)
+        {
+            if (original == null)
+            {
+                throw new ArgumentNullException(nameof(original));
+            }
+
+            return SubstituteRequiredValuesCore(original, new RouteValueDictionary(requiredValues));
+        }
+
+        private RoutePattern SubstituteRequiredValuesCore(RoutePattern original, RouteValueDictionary requiredValues)
+        {
+            // Process each required value in sequence. Bail if we find any rejection criteria. The goal
+            // of rejection is to avoid creating RoutePattern instances that can't *ever* match.
+            //
+            // If we succeed, then we need to create a new RoutePattern with the provided required values.
+            //
+            // Substitution can merge with existing RequiredValues already on the RoutePattern as long
+            // as all of the success criteria are still met at the end.
+            foreach (var kvp in requiredValues)
+            {
+                // There are three possible cases here:
+                // 1. Required value is null-ish
+                // 2. Required value corresponds to a parameter
+                // 3. Required value corresponds to a matching default value
+                //
+                // If none of these are true then we can reject this substitution.
+                RoutePatternParameterPart parameter;
+                if (RouteValueEqualityComparer.Default.Equals(kvp.Value, string.Empty))
+                {
+                    // 1. Required value is null-ish - check to make sure that this route doesn't have a
+                    // parameter or filter-like default.
+
+                    if (original.GetParameter(kvp.Key) != null)
+                    {
+                        // Fail: we can't 'require' that a parameter be null. In theory this would be possible
+                        // for an optional parameter, but that's not really in line with the usage of this feature
+                        // so we don't handle it.
+                        //
+                        // Ex: {controller=Home}/{action=Index}/{id?} - with required values: { controller = "" }
+                        return null;
+                    }
+                    else if (original.Defaults.TryGetValue(kvp.Key, out var defaultValue) &&
+                        !RouteValueEqualityComparer.Default.Equals(kvp.Value, defaultValue))
+                    {
+                        // Fail: this route has a non-parameter default that doesn't match.
+                        //
+                        // Ex: Admin/{controller=Home}/{action=Index}/{id?} defaults: { area = "Admin" } - with required values: { area = "" }
+                        return null;
+                    }
+
+                    // Success: (for this parameter at least)
+                    //
+                    // Ex: {controller=Home}/{action=Index}/{id?} - with required values: { area = "", ... }
+                    continue;
+                }
+                else if ((parameter = original.GetParameter(kvp.Key)) != null)
+                {
+                    // 2. Required value corresponds to a parameter - check to make sure that this value matches
+                    // any IRouteConstraint implementations.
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.Routing/Patterns/DefaultRoutePatternTransformer.cs
```diff
@@ -0,0 +1,227 @@
...
+                updatedSegments ?? original.PathSegments);
+        }
+
+        private bool MatchesConstraints(RoutePattern pattern, RoutePatternParameterPart parameter, string key, RouteValueDictionary requiredValues)
+        {
+            if (pattern.ParameterPolicies.TryGetValue(key, out var policies))
+            {
+                for (var i = 0; i < policies.Count; i++)
+                {
+                    var policy = _policyFactory.Create(parameter, policies[i]);
```
JamesNK: I think route constraints today were created twice - once to denormalize routes and again by the matcher. I think there is an opportunity to cache the created route constraints and reuse them now that both sides of the code are in routing.

Low priority tho

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -243,13 +273,14 @@ public static RoutePattern Pattern(string rawText, params RoutePatternPathSegmen
                 throw new ArgumentNullException(nameof(segments));
             }
 
-            return PatternCore(rawText, new RouteValueDictionary(defaults), new RouteValueDictionary(parameterPolicies), segments);
+            return PatternCore(rawText, new RouteValueDictionary(defaults), new RouteValueDictionary(parameterPolicies), requiredValues: null, segments);
         }
-
+
```
JamesNK: nit: spacey

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -311,12 +342,57 @@ public static RoutePattern Pattern(string rawText, params RoutePatternPathSegmen
...
+            }
+
+
             return new RoutePattern(
                 rawText,
                 updatedDefaults ?? EmptyDefaultsDictionary,
                 updatedParameterPolicies != null
                     ? updatedParameterPolicies.ToDictionary(kvp => kvp.Key, kvp => (IReadOnlyList<RoutePatternParameterPolicyReference>)kvp.Value.ToArray())
                     : EmptyPoliciesDictionary,
+                requiredValues ?? EmptyDefaultsDictionary,
```
JamesNK: Rename EmptyDefaultsDictionary to something more generic

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
```diff
@@ -311,12 +342,57 @@ public static RoutePattern Pattern(string rawText, params RoutePatternPathSegmen
...
+                    {
+                        throw new InvalidOperationException(
+                            $"No corresponding parameter or default value could be found for the required value " +
+                            $"'{kvp.Key}={kvp.Value}'. A non-null required value must correspond to a route parameter or the " +
+                            $"route pattern must have a matching default value.");
+                    }
+                }
+            }
+
+
```
JamesNK: nit: liney

test/Microsoft.AspNetCore.Routing.Tests/Patterns/DefaultRoutePatternTransformerTest.cs
```diff
@@ -0,0 +1,339 @@
...
+
+using Microsoft.AspNetCore.Routing.Constraints;
+using Microsoft.Extensions.DependencyInjection;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing.Patterns
+{
+    public class DefaultRoutePatternTransformerTest
```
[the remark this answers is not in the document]
JamesNK: I added a lot of functional tests for routing to MVC, but I worry that there will be some bugs I fixed in MvcEndpointDataSource that only have unit tests, and we could lose coverage by deleting them. Should investigate its history to be safe

test/Microsoft.AspNetCore.Routing.Tests/RouteValuesAddressSchemeTest.cs
```diff
@@ -325,6 +369,7 @@ private CustomRouteValuesBasedAddressScheme CreateAddressScheme(params EndpointD
             string template,
             object defaults = null,
             object requiredValues = null,
```
JamesNK: Rename to metadataRequiredValues to be clearer about its difference with routePatternRequiredValues

src/Microsoft.AspNetCore.Routing/Patterns/DefaultRoutePatternTransformer.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            if (policyFactory == null)
+            {
+                throw new ArgumentNullException(nameof(policyFactory));
+            }
+
+            _policyFactory = policyFactory;
+        }
+
+        public override RoutePattern SubstituteRequiredValues(RoutePattern original, object requiredValues)
```
JamesNK: I'm a little confused about this returning a single route pattern. One of the reasons the logic in MVC was so complicated is a route pattern denormalized to multiple endpoints because of defaults.

{controller=Home}/{action=Index}/{id?} became
* /
* /Home
* /Home/Index/{id}

You're returning one pattern but removing the parameters. How will the necessary permutations be calculated?

rynowak: There is no need for permutations - the DFA matcher will do the right thing (once it's implemented). This is a new fundamental feature of `RoutePattern`

rynowak: > You're returning one pattern but removing the parameters

I am not removing the parameters. I am removing a default value when it conflicts with a required value.

test/Microsoft.AspNetCore.Routing.Tests/Patterns/DefaultRoutePatternTransformerTest.cs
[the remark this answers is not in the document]
rynowak: The problem with that this that this fundamentally works on a different currency. I think using MVC's tests to see some URL matching integration tests would be really valuable.

src/Microsoft.AspNetCore.Routing/Patterns/DefaultRoutePatternTransformer.cs
rynowak: I super don't want to do this for layering reasons.

src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternFactory.cs
rynowak: sure

rynowak:
@JamesNK updated - if you're happy with the state of things, feel free to add the DFA matcher part.

JamesNK:
Looks good to me.

JamesNK (approved):

JamesNK:
> Then we should probably obsolete IRouteValuesAddressMetadata since it's now first class.

IRouteValuesAddressMetadata has route name. It would either be new metadata or on RoutePattern as well.

rynowak:
> IRouteValuesAddressMetadata has route name. It would either be new metadata or on RoutePattern as well

yeah exactly 😁

rynowak:
Did you plan for me to merge this into master or are you going to build on top of it?

PR #913, opened 2018-11-17
Upgrade projects and packages to target netcoreapp3.0

natemcmaster:
Changes:

* Upgrade dependencies and build tools
* Change TFM on Microsoft.AspNetCore.Routing packages to netcoreapp3.0
* Remove .NET Framework tests
* Remove the IL_EMIT conditional compilation because this assembly only targets .NET Core now.

Part of https://github.com/aspnet/AspNetCore/issues/3754

### DIFF
--- a/src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
+++ b/src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
@@ -84,11 +84,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
                 fallback = new DictionaryJumpTable(defaultDestination, exitDestination, pathEntries);
             }
 
-#if IL_EMIT
             return new ILEmitTrieJumpTable(defaultDestination, exitDestination, pathEntries, vectorize: null, fallback);
-#else
-            return fallback;
-#endif
         }
     }
 }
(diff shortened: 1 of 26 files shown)


src/Microsoft.AspNetCore.Routing/Matching/JumpTableBuilder.cs
```diff
@@ -84,11 +84,7 @@ public static JumpTable Build(int defaultDestination, int exitDestination, (stri
                 fallback = new DictionaryJumpTable(defaultDestination, exitDestination, pathEntries);
             }
 
-#if IL_EMIT
             return new ILEmitTrieJumpTable(defaultDestination, exitDestination, pathEntries, vectorize: null, fallback);
-#else
-            return fallback;
```
JamesNK: Off topic:

@rynowak What do you think about testing whether IL emit is supported here? 3.0 will add `RuntimeFeature.IsDynamicCodeSupported`

davidfowl: We should be checking RuntimeFeature.IsDynamicCodeCompiled no?

JamesNK: Sure, since we're doing this for perf that one would be better

PR #914, opened 2018-11-17
Denormalize required values to nodes

JamesNK:
- [x] Matching
- [x] Link generation
- [x] Fix MvcEndpointDataSource tests


src/Microsoft.AspNetCore.Routing/Patterns/DefaultRoutePatternTransformer.cs
```diff
@@ -53,16 +53,7 @@ private RoutePattern SubstituteRequiredValuesCore(RoutePattern original, RouteVa
                     // 1. Required value is null-ish - check to make sure that this route doesn't have a
                     // parameter or filter-like default.
 
-                    if (original.GetParameter(kvp.Key) != null)
```
JamesNK: This test caused issues with action routes that contained {page} and page was a required value because of Razor pages.

src/Microsoft.AspNetCore.Routing/Template/RoutePrecedence.cs
```diff
@@ -233,6 +236,12 @@ private static int ComputeInboundPrecedenceDigit(RoutePatternPathSegment pathSeg
             }
             else if (part is RoutePatternParameterPart parameterPart)
             {
+                // Parameter with a required value is matched as a literal
+                if (RequiredValueHelpers.TryGetRequiredValue(routePattern, parameterPart, out _))
```
JamesNK: I've been focused on getting matching working. Have not tested yet whether link generation should do the same

src/Microsoft.AspNetCore.Routing/CompositeEndpointDataSource.cs
```diff
@@ -170,10 +170,11 @@ private string DebuggerDisplayString
                         var routeValuesAddressMetadata = routeEndpoint.Metadata.GetMetadata<IRouteValuesAddressMetadata>();
                         sb.Append(", Route Name: ");
                         sb.Append(routeValuesAddressMetadata?.RouteName);
-                        if (routeValuesAddressMetadata?.RequiredValues != null)
+                        var routeValues = routeEndpoint.ResolveRouteValues();
```
rynowak: uhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

src/Microsoft.AspNetCore.Routing/Internal/RequiredValueHelpers.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Routing.Internal
+{
+    internal static class RequiredValueHelpers
```
rynowak: no helpers.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -254,6 +264,23 @@ public DfaNode BuildDfaTree(bool includeLabel = false)
...
+                next = new DfaNode()
+                {
+                    PathDepth = parent.PathDepth + 1,
+                    Label = includeLabel ? parent.Label + literal + "/" : null,
+                };
+                parent.AddLiteral(literal, next);
+            }
+
+            nextParents.Add(next);
+        }
```
rynowak, line 282: This was just moved as-is right?

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -145,72 +147,80 @@ public DfaNode BuildDfaTree(bool includeLabel = false)
...
-                                next = new DfaNode()
-                                {
-                                    PathDepth = parent.PathDepth + 1,
-                                    Label = includeLabel ? parent.Label + literal + "/" : null,
-                                };
-                                parent.AddLiteral(literal, next);
-                            }
-
-                            nextParents.Add(next);
+                            AddLiteralNode(includeLabel, nextParents, parent, literalPart.Content);
```
rynowak, line 149: If you're going to start moving this code around, I'd expect the same treatment for the catch all part (L156-186) - basically either make this low level and awesome, or make it readable and self-documenting

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -145,72 +147,80 @@ public DfaNode BuildDfaTree(bool includeLabel = false)
...
-                                nextParents.Add(parent.Parameters);
-                            }
-
-                            // We also create a 'catchall' here. We don't do further traversals
-                            // on the catchall node because only catchalls can end up here. The
-                            // catchall node allows us to capture an unlimited amount of segments
-                            // and also to match a zero-length segment, which a parameter node
-                            // doesn't allow.
-                            if (parent.CatchAll == null)
+                            if (parameterPart.IsCatchAll)
```
rynowak: Why move just this part into a separate if? this seems really non-obvious

src/Microsoft.AspNetCore.Routing/Patterns/DefaultRoutePatternTransformer.cs
rynowak: I'm not sure I get this. Does that work at all in MVC today? I don't understand why we would change a routing feature for this.

src/Microsoft.AspNetCore.Routing/RouteEndpoint.cs
```diff
@@ -59,5 +59,14 @@ public sealed class RouteEndpoint : Endpoint
         /// Gets the <see cref="RoutePattern"/> associated with the endpoint.
         /// </summary>
         public RoutePattern RoutePattern { get; }
+
+        internal IReadOnlyDictionary<string, object> ResolveRouteValues()
```
rynowak: If we're planning on removing `IRouteValuesAddressMetadata` can we just leave this ugly for now? This shouldn't be used anywhere besides the RVAS

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
JamesNK: Ya

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
JamesNK: I extracted this piece to a method because it is called from multiple places: for either a literal segment or a parameter with required value segment.

src/Microsoft.AspNetCore.Routing/RouteEndpoint.cs
JamesNK: Is IRouteValuesAddressMetadata being removed or made obsolete?
And if it is obsolete, is it being ignored all together, or used as a fallback source of required values?

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
JamesNK: Because there are now multiple actions to take for parameter parts:
* Is it a catch all
* Does it have a required value
* Everything else

The alternative to nesting is replicating the check that the part is a parameter in the parent if else and that seemed less intuitive to me.

src/Microsoft.AspNetCore.Routing/Internal/RequiredValueHelpers.cs
JamesNK: Is an internal method on RoutePattern an acceptable place?

src/Microsoft.AspNetCore.Routing/Patterns/DefaultRoutePatternTransformer.cs
JamesNK: ~~You're right, this change is a bug. REVERTED~~

It is correct for convention routes but not attribute routes. I added some logic to attribute route substitution to remove empty required values when there is a matching parameter.

JamesNK:
Matching is passing!
Link generation is almost passing! Something in link generation is breaking with areas.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
rynowak: That seems great to me.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -145,72 +147,80 @@ public DfaNode BuildDfaTree(bool includeLabel = false)
...
+                                        var parameterPolicy = _parameterPolicyFactory.Create(parameterPart, reference);
+                                        if (parameterPolicy is IOutboundParameterTransformer parameterTransformer)
+                                        {
+                                            requiredValue = parameterTransformer.TransformOutbound(requiredValue);
+                                            break;
+                                        }
+                                    }
+                                }
+
+                                AddLiteralNode(includeLabel, nextParents, parent, requiredValue.ToString());
```
rynowak: In a case like `Foo/{action=Index}` where `rv: { action = "Index" }`  how does `Foo` match? Is that handled by the handling for default values? This is worth a comment at least.

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -500,11 +527,22 @@ internal Candidate CreateCandidate(Endpoint endpoint, int score)
...
-                        _slots.Add(hasDefaultValue ? new KeyValuePair<string, object>(parameterPart.Name, parameterPart.Default) : default);
+                        if (RequiredValueHelpers.TryGetRequiredValue(routeEndpoint.RoutePattern, parameterPart, out var requiredValue))
+                        {
+                            _slots.Add(new KeyValuePair<string, object>(parameterPart.Name, requiredValue));
+                        }
+                        else
+                        {
+                            var hasDefaultValue = parameterPart.Default != null || parameterPart.IsCatchAll;
+                            _slots.Add(hasDefaultValue ? new KeyValuePair<string, object>(parameterPart.Name, parameterPart.Default) : default);
+                        }
```
rynowak, line 540: I think this needs some comments explaining what cases as possible. I had to read this a few times to convince myself that it's right 👍

src/Microsoft.AspNetCore.Routing/Template/RoutePrecedence.cs
rynowak: I think for outbound we should treat it like a parameter. Inbound seems right.

test/Microsoft.AspNetCore.Routing.Tests/EndpointFactory.cs
```diff
@@ -21,17 +24,40 @@ internal static class EndpointFactory
...
+
+            var routePattern = RoutePatternFactory.Parse(template, defaults, policies);
+
             if (requiredValues != null)
             {
-                d.Add(new RouteValuesAddressMetadata(new RouteValueDictionary(requiredValues)));
+                var policyFactory = CreateParameterPolicyFactory();
+                var defaultRoutePatternTransformer = new DefaultRoutePatternTransformer(policyFactory);
+
+                routePattern = defaultRoutePatternTransformer.SubstituteRequiredValues(routePattern, requiredValues);
```
rynowak: Do we need to do this here? The pattern can be constructed with required values, but it work incorrectly if it's invalid. It's **really** important that route patterns can be constructed without services. That's a more important invariant than features like running constraints for required values.

test/Microsoft.AspNetCore.Routing.Tests/Matching/DfaMatcherBuilderTest.cs
```diff
@@ -727,6 +729,253 @@ public void BuildDfaTree_WithPolicies_AndBranches_BothPoliciesSkipped()
...
+            next = Assert.Single(home.Literals);
+            Assert.Equal("view-all", next.Key);
+
+            var index = next.Value;
+            Assert.Same(endpoint, Assert.Single(index.Matches));
+            Assert.Null(index.Literals);
+        }
+
+        [Fact]
+        public void BuildDfaTree_RequiredValues_AndDefaults_AndParameterTransformer()
```
rynowak, line 938: This seems to me like all of the important tests are covered here 👍

test/Microsoft.AspNetCore.Routing.Tests/Matching/DfaMatcherTest.cs
```diff
@@ -106,6 +106,45 @@ public async Task MatchAsync_InvalidRouteConstraint_NoEndpointMatched()
             Assert.Null(context.Endpoint);
         }
 
+        [Fact]
+        public async Task MatchAsync_ParameterTransformer_EndpointMatched()
```
rynowak, line 293: Needs more integration tests that actually do URL matching and link generation with required values

rynowak (commented):
As promised, I gave this another look. It really seems like its on the right track

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
JamesNK: > Is that handled by the handling for default values?

Yes, HasAdditionalRequiredSegments will add the endpoint as a match for the current node. I don't think this needs a comment to call that out here. There is no difference between how that logic interacts with required values and anything else.

I'll add a comment about the what is happening to the required value before it is used as a literal.

test/Microsoft.AspNetCore.Routing.Tests/EndpointFactory.cs
JamesNK: Moved this logic to the few integration-style tests that use it to replace defaults.

JamesNK:
🆙 📅

natemcmaster:
We're merging Routing today into aspnet/AspNetCore. If you can merge this in the next 30 minutes or so, we can bring this along. Otherwise, you'll have to open a new PR against https://github.com/aspnet/AspNetCore.

src/Microsoft.AspNetCore.Routing/IRouteNameMetadata.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+
```
rynowak: extra line. According to people like you this is very important 
```suggestion

```

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -4,7 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing.Internal;
```
rynowak: What's `.Internal` here?

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
```diff
@@ -720,5 +763,15 @@ private static (INodeBuilderPolicy[] nodeBuilderPolicies, IEndpointComparerPolic
 
             return (nodeBuilderPolicies.ToArray(), endpointComparerPolicies.ToArray(), endpointSelectorPolicies.ToArray());
         }
+
+        private static bool TryGetRequiredValue(RoutePattern routePattern, RoutePatternParameterPart parameterPart, out object value)
```
rynowak, line 765: much better

src/Microsoft.AspNetCore.Routing/Template/RoutePrecedence.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Linq;
+using Microsoft.AspNetCore.Routing.Internal;
```
rynowak: Can we remove this?

test/Microsoft.AspNetCore.Routing.Tests/Template/TemplateBinderTests.cs
```diff
@@ -1317,6 +1321,96 @@ public void BindValues_ParameterTransformer()
...
+            // Act
+            var result = binder.GetValues(ambientValues, explicitValues);
+            var boundTemplate = binder.BindValues(result.AcceptedValues);
+
+            // Assert
+            Assert.Equal(expected, boundTemplate);
+        }
+
+        [Fact]
+        public void BindValues_sdfsdf()
```
rynowak: Hmmm? Better name maybe?

rynowak (approved):
Awesome! 👍

src/Microsoft.AspNetCore.Routing/IRouteNameMetadata.cs
JamesNK: Thank you, this is important 🙏

src/Microsoft.AspNetCore.Routing/Template/RoutePrecedence.cs
JamesNK: We can

src/Microsoft.AspNetCore.Routing/Matching/DfaMatcherBuilder.cs
JamesNK: Nothing. Removed

natemcmaster:
FYI @ryanbrandenburg this will likely be the last change to master before I make this repo readonly.

rynowak:
This requires a reaction in MVC tho

JamesNK:
Will be rebased. Close this 💀

natemcmaster:
Just chatted with James. Yes, this will need to be rebased onto aspnet/AspNetCore. I'm going to close this PR, but I'll make sure to leave this branch around so you can copy the code over.

